PR #11, opened 2015-02-04
Added redirects for Passenger

Ginja:
When serving this application out through Passenger, certain actions
would lead to a HTTP 500 error. After inspecting the Rack::Request.env
attribute, it appears that request_path is missing and/or not set when
using Passenger. And since this value is used to redirect the user in
certain cases, Rack was throwing an error because it was nil. The fix
was to statically set the redirect variable for certain actions.

### DIFF
--- a/lib/onetime/app/web.rb
+++ b/lib/onetime/app/web.rb
@@ -65,7 +65,7 @@ module Onetime
     end
 
     def receive_feedback
-      publically do
+      publically('/feedback') do
         logic = OT::Logic::ReceiveFeedback.new sess, cust, req.params
         logic.raise_concerns
         logic.process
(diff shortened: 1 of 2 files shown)


delano:
Thanks the report and the fix!

There is still a case where that error can happen (see `def carefully` in `helpers.rb`). I'm on mobile currently but I'll pull in this request and add a fix for that as well.

Ginja:
Hi @delano,

I saw that, but I couldn't trigger the same behaviour I was seeing as when I was trying to submit a blank secret for example. If I try to hit a fake page I do get properly redirected to the "Page not found" page, so I'm not sure if `def carefully` needs to be changed.

Ginja:
This probably can be closed; looks like you've fixed it in your recent releases :+1: 

I haven't tested it, but will try to update sometime this month/early May.

PR #20, opened 2015-06-15
Change to work natively with current app config

GrayCoder:
When I  was trying to spin this up in a docker container with the default files here, this line commented in the redis config made the application error out. If CHANGEME was provided by default as the redis database password, the app should work almost perfectly out of a cloned version of the repo.

### DIFF
--- a/etc/redis.conf
+++ b/etc/redis.conf
@@ -7,7 +7,7 @@ dir /var/lib/onetime/
 dbfilename default.rdb
 appendfilename default.aof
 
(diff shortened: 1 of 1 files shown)


delano:
What was the error?

pgray:
It failed to startup completely without a password set.

If the solution is to just script the configuration of the password in the dockerfile, I can do that too.

Just thought it might be nice for it to work straight from a cloned repo.

Issue #30, opened 2016-05-14
Debian: libreadline5 was replaced witha new library


rugk:
```console
$ sudo apt-get install ntp libyaml-dev libevent-dev zlib1g zlib1g-dev openssl libssl-dev libxml2 libreadline5-dev
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package libreadline5-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  libreadline-gplv2-dev

E: Package 'libreadline5-dev' has no installation candidate
```

So I think the instructions in the Readme should be updated...

PR #32, opened 2016-05-14
Correct instructions in Readme

rugk:
1. Fixes https://github.com/onetimesecret/onetimesecret/issues/30
2. Adds one missing `mkdir`
3. Fixes  `cp` command to copy all files (otherwise the `locale` dir would be missing)
4. Adds note to remind user to secure the config directory (it contains passwords and other secrets!)
5. Update the language config suggestion


delano:
Thanks for this.

PR #40, opened 2016-09-29
Enable support for new TLDs

florianb:
Like my business email-address: @mindkeeper.solutions

Registering on onetimesecret.com wasn't possible because of my "invalid" email-address.


delano:
I pushed a similar fix this weekend so I'll close this one.

PR #42, opened 2016-10-14
Added Dockerfile

ajvb:
Added default Dockerfile for onetimesecret.

There is one piece of this PR that you will not like and I don't like either, but I left it in so that we could talk about how to solve this.

Why I am referring to is this block of code in the prod section of `config.ru`:

``` ruby
['img', 'css', 'js'].each do |s|
  map("/"+s+"/") { run Rack::File.new("#{PUBLIC_DIR}/"+s) }
end
```

By default, it does not seem that `public/web/` is route-able within the production Rack config.

What am I doing wrong here? I feel like I am missing something obvious.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,27 @@
+FROM ruby:2.3.1
+
+#
+# Make sure to edit etc/config in this repo prior to building this container.
+#
+
+RUN \
+  apt-get update -qq && \
+  apt-get install -y \
+  build-essential \
+  ntp \
+  libevent-dev \
+  zlib1g \
+  zlib1g-dev \
+  openssl \
+  libreadline-gplv2-dev
+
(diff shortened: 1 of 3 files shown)


ajvb:
Also, I do know that the right way to do this is to use http://www.rubydoc.info/github/rack/rack/Rack/Static - I was more just trying to get it too work and ask the question as to what I am missing around the default Production setup.

delano:
Hey!

The entire `public/web` directory is mapped to be served as static files as long as the environment is set to `:dev`:
https://github.com/onetimesecret/onetimesecret/blob/master/config.ru#L36

In production static files should be served by a proxy like nginx or httpd and I wouldn't want to serve them through thin b/c that would encourage folks to avoid HTTPS (which is the other reason for running a proxy).

PR #44, opened 2016-10-20
Bump eventmachine version to work with Ruby 2.2.*

mcasper:
The current version of eventmachine is 1.0.3, which has known build problems with Ruby 2.2.0-dev and greater: https://github.com/eventmachine/eventmachine/issues/495

Bumping the version to at least 1.0.4 fixes the problem, but I went ahead and bumped up to the latest stable. Let me know if you'd like to just take the patch version bump, or go about this in a different way.


mcasper:
Looks like this was addressed in #42, closing

PR #63, opened 2017-04-28
Retrieve Recent Metadata method should be GET

chelnak:
Had to look at the implemented route to work out that the method for this endpoint should be GET rather than POST


delano:
Nice catch, thank you!

PR #66, opened 2017-07-03
Update README.md with secure download of Ruby

kozmic:
Change Ruby installation from insecure FTP to secure HTTPS.

Couldn't find an official download for Redis over HTTPS :(


delano:
Thanks I updated the Redis URI too and updated it to v3 (haven't tested with redis v4 yet).

kozmic:
Perfect @delano :)

PR #82, opened 2018-05-21
adding secrets file upload

Oribracha:
adding a simple option for uploading a secret text file into the secrets "textarea" and downloading it as a secret file on the receiver side.

### DIFF
--- a/templates/web/partial/create_secret_form.mustache
+++ b/templates/web/partial/create_secret_form.mustache
@@ -3,12 +3,30 @@
 <form id="createSecret" method="post" autocomplete="off" action="/" class="form-horizontal">
   {{{add_shrimp}}}
   <fieldset>
+
     <div class="text-input">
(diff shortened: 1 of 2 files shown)


564456:
fa36b647

564456:
https://onetimesecret.com/secret/acgudm7ur632qfni5

isaija:
9rsO!nMPehD7

isaija:
роза и соня

isaija:
fa36b64

delano:
I'll be closing this one up unfortunately. Not implementing file support was and is an intentional design choice. Depending on the filetype, it's very easy to accidentally share sensitive information (PDFs, spreadsheets, images of course) without realizing it. We could add support for stripping that stuff out for common types, but we can't guarantee that a file is totally safe to share. 

Our recommendation is still to upload the file using a service you trust, then create a secret with the shareable link. Why do that? Because you can share the secret without revealing where you store files and the exact path to get to one.

Also, what was going on with this thread??

Issue #81, opened 2018-05-11
Installation instructions not complete


IS-KM:
Good day everybody,

I tried to install the OTS with the manual and here it comes to several errors. I've already progressed a few steps but I can not fix the following error. 

> ots@ots:~/onetimesecret-0.10$

 bundle install --frozen --deployment --without=dev
Fetching gem metadata from https://rubygems.org/..........
Using addressable 2.2.6
Using highline 1.7.8
Using annoy 0.5.6
Using attic 0.5.3
Using bcrypt-ruby 3.0.0
Using bundler 1.16.1
Using crack 0.1.8
Using daemons 1.2.4
Using drydock 0.6.9
Using encryptor 1.1.3
Using eventmachine 1.2.0.1
Using gibbler 0.8.9
Using multi_json 1.12.1
Using redis 2.2.2
Using storable 0.8.9
Using uri-redis 0.4.2
Using familia 0.7.1
Using httparty 0.7.7
Using mime-types 1.25.1
Using polyglot 0.3.5
Using treetop 1.4.15
Using mail 2.5.4
Using mustache 0.99.3
Using rack 1.4.5
Using otto 0.4.1
Fetching yajl-ruby 1.1.0
Installing yajl-ruby 1.1.0 with native extensions
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    current directory: /home/ots/onetimesecret-0.10/vendor/bundle/ruby/2.4.0/gems/yajl-ruby-1.1.0/ext/yajl
/usr/share/rvm/rubies/ruby-2.4.1/bin/ruby -r ./siteconf20180511-3631-1km21la.rb extconf.rb
creating Makefile

current directory: /home/ots/onetimesecret-0.10/vendor/bundle/ruby/2.4.0/gems/yajl-ruby-1.1.0/ext/yajl
make "DESTDIR=" clean

current directory: /home/ots/onetimesecret-0.10/vendor/bundle/ruby/2.4.0/gems/yajl-ruby-1.1.0/ext/yajl
make "DESTDIR="
compiling yajl.c
compiling yajl_alloc.c
compiling yajl_buf.c
compiling yajl_encode.c
compiling yajl_ext.c
yajl_ext.c: In function ‘yajl_encode_part’:
yajl_ext.c:117:21: warning: variable ‘status’ set but not used [-Wunused-but-set-variable]
     yajl_gen_status status;
                     ^~~~~~
[...]

PR #93, opened 2018-08-03
Update yajl-ruby version in Gemfile/Gemfile.lock

Jimadine:
This fixes an an install problem caused by referencing an older version of yajl-ruby in Gemfile, as reported at https://github.com/onetimesecret/onetimesecret/issues/81


TheWalkers:
It would be nice to get this merged, as onetimesecret cannot be built against a modern ruby without it.

@onetimesecret

PR #102, opened 2018-12-19
add reference to bash API/CLI wrapper

eengstrom:
I added to the API docs a reference to the bash API and CLI wrapper around OTS that I wrote and use regularly.  I hope someone else finds this helpful.


delano:
Thank you so much. This will be in the next release. 

Sorry for the protracted delay in merging this. This thing we call life got in the way.

PR #110, opened 2019-05-15
Add Dockerfile

lpl-hosting:
Building this Dockerfile gives you an image with all you need to use the bin/run.sh script :)

You can run this image with the source code included on your favorite docker orchestrator.

Or you can run it localy with and moount a volume on /home/ots/onetimesecret

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,32 @@
+FROM debian:stretch
+
+RUN apt-get update
+RUN apt-get install -y build-essential ruby ruby-dev redis-server procps
+RUN apt-get install -y libyaml-dev libevent-dev zlib1g zlib1g-dev openssl \
+					   libssl-dev libxml2 libreadline-gplv2-dev
+RUN gem install bundler
+
+# Copy source, you need to setup etc/redis.conf & etc/config before building image
+COPY . /home/ots/onetimesecret
+
+# Add user ots, set perms and copy conf files
+RUN useradd ots -s /bin/bash -d /home/ots -m
+RUN mkdir /etc/onetime && chown ots /etc/onetime
+RUN mkdir /var/log/onetime /var/run/onetime /var/lib/onetime
+RUN chown ots /var/log/onetime /var/run/onetime /var/lib/onetime /home/ots -R
+RUN cp -R /home/ots/onetimesecret/etc/* /etc/onetime/
+
+# Install app with ots user
+USER ots
+WORKDIR /home/ots/onetimesecret
+RUN bundle install --frozen --deployment --without=dev
+RUN bin/ots init
+
+# clean packages 
+USER root
+RUN apt-get purge -y build-essential && apt-get autoremove -y && apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# Run app with ots user
+USER ots
+WORKDIR /home/ots/onetimesecret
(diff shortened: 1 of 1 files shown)


oallain:
Cc @delano

windoz:
This looks good. There's been a few requests for docker support so this will make some people quite happy. 

Do you want to add a mention to the README to direct folks to the docker support?

lpl-hosting:
I don't know any folks at docker support ^_^
You can validate the PR like this if you like to.

swissbuechi:
@lpl-hosting Thank you for your work.
Could you please add some more information on how to start the container?
I created a docker-compose.yml file which can be viewed here: https://github.com/swissbuechi/onetimesecret/tree/master
When I try to start my Container, I get the followng error: 
`/home/ots/onetimesecret/vendor/bundle/ruby/2.3.0/gems/redis-2.2.2/lib/redis/client.rb:47:in call': ERR Client sent AUTH, but no password is set (RuntimeError)`
grafik

How can I set the password?
I cant get to the Bash-Shell of the Container because it exits after the error.
Thank you very much for your support.

swissbuechi:
> 
> 
> @lpl-hosting Thank you for your work.
> Could you please add some more information on how to start the container?
> I created a docker-compose.yml file which can be viewed here: https://github.com/swissbuechi/onetimesecret/tree/master
> When I try to start my Container, I get the followng error:
> `/home/ots/onetimesecret/vendor/bundle/ruby/2.3.0/gems/redis-2.2.2/lib/redis/client.rb:47:in call': ERR Client sent AUTH, but no password is set (RuntimeError)`
> grafik
> 
> How can I set the password?
> I cant get to the Bash-Shell of the Container because it exits after the error.
> Thank you very much for your support.

I fixed it by myself.
Here is the Repo which I used as referenc: https://github.com/dismantl/onetimesecret-docker

swissbuechi:
Actualy the reference: https://github.com/dismantl/onetimesecret-docker is a very bad source.
Since the author removed the custom redis.conf so you can not save your unopend secrets presistent.
Every time you reboot or pull your docker container, you loose all your data.
Please take a look at the docker-compose and dockerfile from my Repo. I managed to use a Volume for presistent data!
https://github.com/swissbuechi/onetimesecret

mildsunrise:
This Dockerfile would not be accepted by as an official image by Docker because it starts more than one service (Redis and the main application). "Containers should do one thing and do it well", the Redis server should be started separately according to them.

This is for a good reason. This image, as it is, has a race condition: if the main app starts before Redis has initialized, it will attempt to query data from it, which will result in a `LOADING Redis is loading the dataset in memory`. The whole container will die, and the cycle starts again. TL;DR: the container will start constantly crashing once the DB grows past a random size.

To fix the above, the command should be i.e. `bundle exec thin -R config.ru start`. I'd also set `ENV RACK_ENV prod`. There should be an environment variable to point the application to its redis server...

Also, this image doesn't mark the data directory as `VOLUME`. All data will be lost on container removal (unless user explicitely mounts a volume there) which is bad.

Some other things:
 - it should `EXPOSE` its http port
 - it should be based on the [official ruby image](https://hub.docker.com/_/ruby)
 - if `COPY .` is used, a `.dockerignore` file should be added too
 - `site.ssl` and `site.host` must be specified by the user via env vars

PR #145, opened 2020-10-17
Package with Docker

mildsunrise:
This PR adds `.dockerignore` and `Dockerfile` to allow it to be deployed with Docker. The Dockerfile contains instructions & examples of use.

This Dockerfile contains many improvements with respect to #110, see https://github.com/onetimesecret/onetimesecret/pull/110#issuecomment-710808082.

To allow the application to run correctly under docker, some extra changes have been made:
 - Allow the application to be configured by environment variables too (this is how Docker usually works)
 - `config.ru`: Set stdout to be always unbuffered (so logs appear immediately). This is useful for both Docker and other environments.
 - `config.ru`: Set `PUBLIC_DIR` in production environment too, otherwise `public` files are not served.

If you don't like those changes, tell me and I'll find a workaround.

### DIFF
--- a/lib/onetime.rb
+++ b/lib/onetime.rb
@@ -4,6 +4,7 @@ require 'bundler/setup'
 
 require 'onetime/core_ext'
 
+require 'erb'
 require 'syslog'
 
 require 'encryptor'
@@ -135,7 +136,7 @@ module Onetime
     attr_reader :env, :base, :bootstrap
     def load path=self.path
       raise ArgumentError, "Bad path (#{path})" unless File.readable?(path)
-      YAML.load_file path
+      YAML.load(ERB.new(File.read(path)).result)
     rescue => ex
       SYSLOG.err ex.message
       msg = path =~ /locale/ ?
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,10 @@
+.git
+
+.bundle
+log
+tmp
+.rudy
+
+.ruby-version
+
+.DS_Store
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,46 @@
+# To use this image, you need a Redis database with persistence enabled.
+# You can start one with Docker using i.e.:
+#
+# $ docker run -p 6379:6379 -d redis
+#
+# Then start this image, specifying the URL of the redis database:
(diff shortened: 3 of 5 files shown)


A-styler:
ruby:2.6 is the latest work version of ruby without errors

quentin-lpt (approved):
all good

quentin-lpt:
Hello @mildsunrise 
Could I add a docker-compose.yml directly in your Pull Request ? 
I guess I would need some dev access to the repo ? (never done that before)

mildsunrise:
I think this PR already has many changes as-is... to avoid complicating it further, I'd suggest you start a separate PR based on my branch, and add docker-compose.yml there :)

quentin-lpt:
Will do, thanks Alba :)

whysthatso:
is there anything missing to move this pr? i'd like to use it :)

lib/onetime.rb
```diff
@@ -135,7 +136,7 @@ module Config
     attr_reader :env, :base, :bootstrap
     def load path=self.path
       raise ArgumentError, "Bad path (#{path})" unless File.readable?(path)
-      YAML.load_file path
+      YAML.load(ERB.new(File.read(path)).result)
```
delano, line 139: I see. This allows the config file to be interpolated. Could you add a comment with a usage example?

delano (approved):
Thanks so much for this. A bunch of folks will be happy to see this merged. 

I had the one question re: loading the config file, but I'm going to go ahead and merge right away. I'm going to think about this change some more and I may save for a follow-on release so we can get the Docker config out the door. 

Nice work!

lib/onetime.rb
delano: I uncollapsed the config changes and see the obvious usecase 😹. All good

mildsunrise:
Thank you! I see the config file question is resolved. Btw, I just noticed a typo in the help comments:

> you should protect your Redis instance with authentication or Redis networks

Redis networks -> Docker networks

isAAAc:
bump :) 
is the docker version usable ?

7olstoy:
@isAAAc Sure, why not?

Issue #163, opened 2021-05-03
Git Clone Failing


elcottm:
When i run

git clone git@github.com:onetimesecret/onetimesecret.git

I get the following

Cloning into 'onetimesecret'...
The authenticity of host 'github.com (192.30.255.112)' can't be established.
RSA key fingerprint is SHA256:xxxxxx.
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
Warning: Permanently added 'github.com,192.30.255.112' (RSA) to the list of known hosts.
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

btpmike:
use git clone https://github.com/onetimesecret/onetimesecret.git

isAAAc:
https://github.com/onetimesecret/onetimesecret/pull/171

delano:
Ah, I get it now. If you don't have a github account or don't have an SSH key configured, you'll get that permission denied error. 

My old thinking was that it's reasonable to assume the person has a github account. That unnecessarily adds a negative bias towards people who don't.

For posterity, I'll also add a note about how to check

### With a github account
```bash
$ ssh -T git@github.com
Hi delano! You've successfully authenticated, but GitHub does not provide shell access.
```

### Without a github account
```bash
Warning: Permanently added the RSA host key for IP address '0.0.0.0/0' to the list of known hosts.
git@github.com: Permission denied (publickey).
```

Thanks folks.

PR #171, opened 2021-06-30
Update README.md

isAAAc:
git clone via https instead ssh


delano:
Thanks for helping make onetimesecret better @isAAAc. 

Resolves #163

PR #174, opened 2021-07-15
fix syslog err msg and upgrade jquery 3.6.0

dwdraju:
Upgrading jquery because of cross-site scripting (XSS) vulnerabilities on version less than 3.5.


delano:
Thank you so much! I'll merge this now and it will be included in the next release.

PR #184, opened 2021-11-22
Add ruby bundler and dev to required dependencies

bennettforkner:
When I tried to run this process on ubuntu 20.04, it did not recognize the "bundle" command and suggested that I install this package. It also failed on updating the bundle until I installed ruby-dev. I did so and it worked.


delano (approved):
Good call. Thank you and sorry about the delay

PR #202, opened 2022-06-09
Add docker compose config

s01ipsist:
https://docs.docker.com/compose/

Compose is a tool for defining and running multi-container Docker applications.

`docker compose up` is a very simple way to get a local running instance.


delano:
Hey, thanks for adding this. I've been slow as hell adding docker + docker-compose configs so this is great and will help a lot of folks in the meantime!

PR #203, opened 2022-06-09
Fix Arabic compilation error

s01ipsist:
Trying to run this as is with Ruby 2.3 gives this error:

```
root@79a926ee95f2:/usr/src/app# bundle exec thin -e dev -R config.ru -p 7143 start
onetime[7]: (<unknown>): did not find expected key while parsing a block mapping at line 108 column 5
Error loading locale: /usr/src/app/etc/locale/ar ((<unknown>): did not find expected key while parsing a block mapping at line 108 column 5)
onetime[7]: Error loading locale: /usr/src/app/etc/locale/ar ((<unknown>): did not find expected key while parsing a block mapping at line 108 column 5)
```

I don't know a character of Arabic(!) but I can see the extra quote characters, so I've replaced with single quotes, which fixes the runtime error.


onetimesecret:
Hey, thanks for this 🙏

PR #222, opened 2022-11-04
Added autofocus to secret textarea

nickhumberstone:
I use onetimesecret almost daily for work, often using ctrl+t to open a new tab, type o and then press enter for it to autofill and take me to onetimesecret. Adding autofocus means once the page loads, if you type it will already be inputting the content into the secret, saving time clicking :)


delano (approved):
Good call. Sorry for the delay 🙏

PR #232, opened 2023-03-03
Fixed issues running in Docker

ibuys:
Fixed syntax issue in Gemfile, Added bundler install to `Dockerfile`, Updated bcrypt library, old library was causing `Error: BCrypt::Errors::InvalidHash: invalid hash`, also commented out bug swallowing error in `helpers.rb`.


delano (approved):
🦾 Great catch with the logging fix.

PR #233, opened 2023-03-20
Create hu

halmosigabor:
I created the hungarian translation of the site.


delano:
😎 Thank you for making Onetime Secret better. this translation will be included in the next release.

PR #256, opened 2023-05-11
Add operating system packages upgrade

drustan:
This adds the OS packages upgrade. Onetime Secret seems to be compatible with Ruby 2.6 at best, and the ruby:2.6-slim upstream image is currently based on an outdated Debian image.


delano:
Thank you!

delano:
I've merged this into the Ruby 2.6 branch. The `develop` branch supports Ruby 3+ now and the docker experience has had an overhaul as well. I would have responded sooner if it weren't for some pesky health issues gumming up the works.

Issue #278, opened 2024-02-01
cant build the docker image


terminaldweller:
running `docker compose build app` on `v0.11.0` results in:
```
 => ERROR [app builder 3/5] RUN gem update --system                                                                                                                                      16.8s
 ------
  > [app builder 3/5] RUN gem update --system:
  > 16.70 ERROR:  Error installing rubygems-update:
  > 16.70   There are no versions of rubygems-update (= 3.5.5) compatible with your Ruby & RubyGems
  > 16.70   rubygems-update requires Ruby version >= 3.0.0. The current ruby version is 2.6.10.210.
  > 16.71 ERROR:  While executing gem ... (NoMethodError)
  > 16.71     undefined method `version' for nil:NilClass
  > 16.72 Updating rubygems-update
  > ------
  > failed to solve: process "/bin/sh -c gem update --system" did not complete successfully: exit code: 1""'))]]
```

milad-amirianfar-azihub:
me too

chanc201:
It looks like the Ruby version is too low:
"rubygems-update requires Ruby version >= 3.0.0. The current ruby version is 2.6.10.210."
Try changing FROM ruby:2.6-slim AS builder to FROM ruby:3.2.0-slim AS builder in Dockerfile

milad-amirianfar-azihub:
I already did that and works for me: FROM ruby:3.3.0-slim AS builder

rcarrascoz:
Hi, if i put 3.3.0.slim, when I start up it says this:
docker logs -f onetime-app
[Fri Mar 15 13:14:51 UTC 2024] INFO: Running entrypoint.sh...
Bundler 2.5.6 is running, but your lockfile was generated with 2.4.12. Installing Bundler 2.4.12 and restarting using that version.
Fetching gem metadata from https://rubygems.org/.
Fetching bundler 2.4.12
Installing bundler 2.4.12
Fetching gem metadata from https://rubygems.org/...........
Using addressable 2.2.6
Fetching highline 2.1.0
Using attic 0.5.3
Using bcrypt 3.1.17
Using bundler 2.4.12
Using byebug 11.1.0
Using ruby-dap 0.1.3
Using daemons 1.4.1
Using drydock 0.6.9
Using encryptor 3.0.0
Using eventmachine 1.2.7
Using multi_json 1.15.0
Using redis 2.2.2
Using storable 0.8.9
Using uri-redis 0.4.2
Using multi_xml 0.6.0
[...]

Issue #275, opened 2023-12-29
Bin/ots init RubyToken Error


raufbhatti24:
when i am trying to install Onetime secret on my ubuntu machine so after running this command (bin/ots init) terminal showing below error. Please guide me how can i fix this error
ots@ip-172-31-33-85:~/onetimesecret$ bin/ots init
/var/lib/gems/3.0.0/gems/storable-0.8.9/lib/proc_source.rb:41:in `<top (required)>': uninitialized constant RubyToken (NameError)
Did you mean?  RubyLex
        from /var/lib/gems/3.0.0/gems/storable-0.8.9/lib/storable.rb:35:in `require'
        from /var/lib/gems/3.0.0/gems/storable-0.8.9/lib/storable.rb:35:in `<class:Storable>'
        from /var/lib/gems/3.0.0/gems/storable-0.8.9/lib/storable.rb:33:in `<top (required)>'
        from /var/lib/gems/3.0.0/gems/sysinfo-0.8.1/lib/sysinfo.rb:2:in `require'
        from /var/lib/gems/3.0.0/gems/sysinfo-0.8.1/lib/sysinfo.rb:2:in `<top (required)>'
        from /home/ots/onetimesecret/lib/onetime.rb:15:in `require'
        from /home/ots/onetimesecret/lib/onetime.rb:15:in `<top (required)>'
        from <internal:/usr/local/lib/site_ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from <internal:/usr/local/lib/site_ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from bin/ots:9:in `<main>'
![Capture](https://github.com/onetimesecret/onetimesecret/assets/49442003/a6316adb-78d7-484e-a209-33d93323f782)

ParkerCampbell5:
Check your Ruby version is compatible

itsgps:
Are there Any Views to upgrade the Base code and Docker code to support Ruby 3+ ? not doing so really hampers installs meaning we have an Old implementation which is essentially full of vulnerabilities...but can't install a New version...

delano:
Our Ruby 3+ support is stable and available via the [latest release candidate](https://github.com/onetimesecret/onetimesecret/releases/tag/v0.13.0-RC6) and on an ongoing basis now the `develop` branch. There are official docker images available (onetimesecret/onetimesecret:latest) as well.

Issue #266, opened 2023-07-26
uninitialized constant RubyToken


alexch94:
Hi,
I just installed on a debian 11.5 server. 
However when I try to run the command "bin/ots init" I have this following error message:
Traceback (most recent call last):
        10: from bin/ots:9:in `<main>'
         9: from /usr/lib/ruby/vendor_ruby/rubygems/core_ext/kernel_require.rb:85:in `require'
         8: from /usr/lib/ruby/vendor_ruby/rubygems/core_ext/kernel_require.rb:85:in `require'
         7: from /home/ots/onetimesecret/lib/onetime.rb:15:in `<top (required)>'
         6: from /home/ots/onetimesecret/lib/onetime.rb:15:in `require'
         5: from /var/lib/gems/2.7.0/gems/sysinfo-0.8.1/lib/sysinfo.rb:2:in `<top (required)>'
         4: from /var/lib/gems/2.7.0/gems/sysinfo-0.8.1/lib/sysinfo.rb:2:in `require'
         3: from /var/lib/gems/2.7.0/gems/storable-0.8.9/lib/storable.rb:33:in `<top (required)>'
         2: from /var/lib/gems/2.7.0/gems/storable-0.8.9/lib/storable.rb:35:in `<class:Storable>'
         1: from /var/lib/gems/2.7.0/gems/storable-0.8.9/lib/storable.rb:35:in `require'
/var/lib/gems/2.7.0/gems/storable-0.8.9/lib/proc_source.rb:41:in `<top (required)>': uninitialized constant RubyToken (NameError)
Did you mean?  RubyLex

How can I resolve this error?

THANKS

raufbhatti24:
Same issue.

delano:
This issue has been resolved as of [v0.13.0](https://github.com/onetimesecret/onetimesecret/releases/tag/latest). Also available via the official docker images now. 

https://hub.docker.com/r/onetimesecret/onetimesecret/tags

Issue #199, opened 2022-05-09
bin/ots init fails


MikeO3:
This is the same as #165 however later versions.
Ubuntu 22.04
bundle install completes

```
ots@onetimesecret:~/onetimesecret$ bin/ots init
/var/lib/gems/3.0.0/gems/storable-0.8.9/lib/proc_source.rb:41:in `<top (required)>': uninitialized constant RubyToken (NameError)
Did you mean?  RubyLex
        from /var/lib/gems/3.0.0/gems/storable-0.8.9/lib/storable.rb:35:in `require'
        from /var/lib/gems/3.0.0/gems/storable-0.8.9/lib/storable.rb:35:in `<class:Storable>'
        from /var/lib/gems/3.0.0/gems/storable-0.8.9/lib/storable.rb:33:in `<top (required)>'
        from /var/lib/gems/3.0.0/gems/sysinfo-0.7.3/lib/sysinfo.rb:2:in `require'
        from /var/lib/gems/3.0.0/gems/sysinfo-0.7.3/lib/sysinfo.rb:2:in `<top (required)>'
        from /home/ots/onetimesecret/lib/onetime.rb:13:in `require'
        from /home/ots/onetimesecret/lib/onetime.rb:13:in `<top (required)>'
        from <internal:/usr/lib/ruby/vendor_ruby/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from <internal:/usr/lib/ruby/vendor_ruby/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from bin/ots:9:in `<main>'
ots@onetimesecret:~/onetimesecret$

```

vinay-chintalapati:
I have the same issue as well

River2906:
Yep - same here

delano:
This issue has been resolved as of [v0.13.0](https://github.com/onetimesecret/onetimesecret/releases/tag/latest). Also available via the official docker images now. 

https://hub.docker.com/r/onetimesecret/onetimesecret/tags

PR #279, opened 2024-02-14
bumped ruby base image to 3.3 bookworm

terminaldweller:
bumped the ruby version in the Dockerfile to `3.3`.
solves #278.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,7 +53,7 @@
 ARG CODE_ROOT=/app
 ARG ONETIME_HOME=/opt/onetime
 
-FROM ruby:2.6-slim AS builder
+FROM ruby:3.3-slim-bookworm AS builder
 
 # Limit to packages needed for the system itself
 # NOTE: We only need the build tools installed if we need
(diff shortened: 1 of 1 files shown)


eric-glb:
Indeed it solves #278 (the `docker build` finishes), but #137, #199, #266 and #275 still prevent application to run (or may I have missed something?).

vipulwairagade:
@eric-glb is right. I bumped the Ruby version in Dockerfile but container throws error. Do we have solution for this, a quick fix maybe ? @onetimesecret 

image
image

Error from container 

image

delano:
Thanks folks. I've merged this in to the ruby 2.6 branch. The `develop` branch for Ruby 3+ has been updated as well as part of PR #312 (Update Docker and Docker Compose config).

delano:
> Indeed it solves #278 (the `docker build` finishes), but #137, #199, #266 and #275 still prevent application to run (or may I have missed something?).

@eric-glb Thanks so much for taking the time to list these and generally keeping up with this issue.

eric-glb:
Glad to help, but that's nothing compared to your work. Thanks for sharing this tool!

Issue #305, opened 2024-04-06
Install Renovatebot to help manage dependencies.


venkatamutyala:
I recently installed this in my work organization, and it's been huge in tracking updates and ensuring we are pinned to digest for all our dependencies. Ex. you have github actions that use @v1 instead of @v1:sha@XXXXXXX

Using @v1 can change or even worse be vulnerable to some supply chain issues while pinning to the @v1:sha... helps ensure no supply chain issues. It's worth noting if the action you are using isn't pinned you are still vulnerable to issues upstream from the action but it's at least once less thing to worry about.

It's simple to install: https://github.com/apps/renovate

Once you install it, all your repos will get a PR asking you to merge "renovate.json" feel free to accept it/merge and then it'll scan your repos for any dependency updates.

I have a slightly improved version of renovate.json that I'm using at work (OSS) and can share it with you as a follow-up PR.

delano:
Ah, pinning actions is smart. It looks like renovate has some really in depth functionality. For managing dependencies specifically, are there other benefits relative to dependabot?

venkatamutyala:
I just finished a roll out of dependabot at work at it seems to do everything while dependabot appears to just do CVE/security updates primarily. It doesn't appear to be uncommon to have both running.

As for what I like about renovatebot:
- It seems to group updates together. Ex. it'll batch updates like pinning dependencies into one PR.
- It works for everything, minor, patch, major. And covers: docker image updates, github actions, ruby/bundler, nodejs/package.json, etc. You can also create custom regex parsers that will monitor for a release and then update your file via PR via regex.

delano:
Awesome, I'm sold. Yeah, create the PR at your leisure and we'll get it merged in 🦾

venkatamutyala:
@delano can you confirm if you installed this? https://github.com/apps/renovate

PR #327, opened 2024-05-08
feat: adding renovate.json for renovatebot

venkatamutyala:
Merge this PR and then install: https://github.com/apps/renovate and once a day this will run and generate PR's. But if you want to disregard updates entirely [just remove line 15](https://github.com/venkatamutyala/onetimesecret/blob/develop/renovate.json#L15) and the default behavior should not re-open. 

I keep it as `always` so I repeatedly get reminded to eventually upgrade to the newer/latest version.

To uninstall just remove it using one of these links:
If you are using an org: https://github.com/organizations/onetimesecret/settings/installations
If you are using a personal account: https://github.com/settings/installations (obviously you would want to be logged into the onetimesecret user)

Here is an example of what it's like in my account: https://github.com/venkatamutyala/onetimesecret/issues/1
That issue will be created by renovatebot and it'll persist indefinitely. As you can see there are a lot of updates. It also lets you see all the updatable versions for many packages so rather then going to the very latest you can check the box for the PR/version you want to have created. 

Closes: https://github.com/onetimesecret/onetimesecret/issues/305

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,21 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:recommended",
+    "docker:pinDigests",
+    "helpers:pinGitHubActionDigests",
+    ":pinDevDependencies"
+  ],
+  "separateMinorPatch": true,
+  "separateMultipleMajor": true,
+  "separateMultipleMinor": true,
+  "labels": [
+    "auto-update"
+  ],
+  "recreateWhen": "always",
+  "forkProcessing": "enabled",
+  "timezone": "America/Los_Angeles",
+  "schedule": [
+    "before 6am"
+  ]
+}


venkatamutyala:
@delano Build is failing. Looks like something changed in the gems upstream? This PR doesn't need a working build so it might make sense to merge a fix separately from my PR? However, happy to take a look if yo

```
Fetching gem metadata from https://rubygems.org/.........
Your bundle is locked to uri-redis (1.1.1) from rubygems repository
https://rubygems.org/ or installed locally, but that version can no longer be
found in that source. That means the author of uri-redis (1.1.1) has removed it.
You'll need to update your bundle to a version other than uri-redis (1.1.1) that
hasn't been removed in order to install.
```

delano:
@venkatamutyala Yeah the upstream checks are passing again now. I'll merge develop in so they'll re-run here and then merge. 

Looking forward to kicking and then rolling the tires.

PR #348, opened 2024-05-23
added helm chart for k8s cluster

blacktig3r:
There is a need of helm charts for onetimesecret deployment on kubernetes cluster.
Updates:

1. added the chart's tgz file
2. added index.yaml file

we will have to enable pages in settings>pages> so that this chart's repo will be public to install on kubernetes cluster.

If this merge to "main" branch then I'll update the README.md file with installation steps also.


venkatamutyala:
It looks like you added a binary file rather then the raw templates. Was this intentional?

venkatamutyala:
@blacktig3r I'll try and help where i can. Do you have a repo with the code in it already? If so, could you share a link?

@delano I'd recommend creating a separate github repository. As for the name `helm`, `helm-charts` would probably be appropriate.

venkatamutyala:
@blacktig3r here we go: https://github.com/onetimesecret/helm-charts

Can you make a PR against that repo and then tag me?

delano:
@blacktig3 Nice, this will be helpful for a lot of folks!

Thanks for the assist @venkatamutyala.

blacktig3r:
Great..Let me create a PR against that repo.

Issue #372, opened 2024-06-09
Look into redis client upgrade from 4.8 to 5.0


delano:
See issue details and discussion on #370 Unexpected error while processing request: undefined method `custid' for nil. 

The error is related to this line in `RedisHash#method_missing`:

https://github.com/onetimesecret/onetimesecret/blob/ddec845b800dd2f92220522dbb57cef5262c70dc/lib/onetime/models.rb#L141

- [x] Include a regression test

delano:
See #426, #427.

Issue #238, opened 2023-03-31
Redis ssl support


sebastien-coavoux:
Hi there,

I am trying to run a self hosted instance of onetimesecret with a redis backend with TLS enabled (scheme `rediss://`). Unfortunately it looks like it is not supported for now

After a bit of digging I can see that, at least the following is needed for the app to start : 
- Patch [uri-redis](https://github.com/delano/uri-redis) to support `redis://`
- Upgrade [redis-rb](https://github.com/redis/redis-rb) in GemFile to support `redis://`
- Patch [familia](https://github.com/delano/familia) to remove deprecated field in the newer version of redis-rb (`thread_safe` and `logging`)

The good news is that @delano is the author / main contributor of uri-redis and familia so agreeing on supporting that should be fine. However as the familia lib hasn't been updated for a while you might not wish to keep it as a project dep as a whole and avoid maintaining the lib. 

Let me know your thought on that, is supporting redis TLS something you would consider ?

sebastien-coavoux:
I had a PR ready on all those 3 repo back at that time, unfortunately I never heard about @delano . The tree has changed a lot since, not sure if it is still relevant (maybe it is supported now ?)

If you have a few minutes to update me I could spend time to send you something that goes on that direction.

delano:
Hey @sebastien-coavoux I'm finally catching up on the backlog of tickets/PRs. Thanks so much for following up again. 

The big change to this codebase (and dependencies) was a hard fork for Ruby 3 support. The current `develop` branch is a new history so any previous commits won't merge into it (easily). The previous one has moved to [rel/0.12](https://github.com/onetimesecret/onetimesecret/tree/rel/0.12) so if it's easier for you, you can create a PR targeting that branch. I don't mind "porting" the changes over to `develop` at all, but ideally you can create a branch from `develop` and re-commit the changes there. That way yo get the proper attribution as well. 

The other libraries -- uri-redis and familia -- have been updated to support Ruby 3 but are otherwise not changed much.

sebastien-coavoux:
As far as I remember I had issue with old ruby and redis version. I can see ruby version was updated but redis seems still to be locked at 4.8.1 while 5.2 is the latest (ssl support is in since [4.70](https://github.com/redis/redis-rb/blob/v4.8.1/CHANGELOG.md#471) ). If there is no plan to move to 5.X I'll work with that

delano:
[...]

PR #427, opened 2024-06-16
Fix serialization of Boolean an Nil for redis > 5.X

sebastien-coavoux:
Probably need to rebase on #426 first , or maybe #426 can pick up the diff 
Related to #238

Edit : I rebased to test locally to mind the diff, i'll rebase on develop once #426 is merged

Last edit : This also fix a small bug in the session that was taking the user agent as custid

Related to #372

### DIFF
--- a/lib/onetime/app/helpers.rb
+++ b/lib/onetime/app/helpers.rb
@@ -132,7 +132,7 @@ class Onetime::App
       if req.cookie?(:sess) && OT::Session.exists?(req.cookie(:sess))
         @sess = OT::Session.load req.cookie(:sess)
       else
-        @sess = OT::Session.create req.client_ipaddress, req.user_agent
+        @sess = OT::Session.create req.client_ipaddress, "anon", req.user_agent
       end
       if sess
         sess.update_fields  # calls update_time!
--- a/lib/onetime/models/metadata.rb
+++ b/lib/onetime/models/metadata.rb
@@ -77,13 +77,13 @@ module Onetime
       # Make sure we don't go from :shared to :viewed
       return unless state?(:new) || state?(:viewed)
       @state = :received
-      update_fields :state => :received, :received => Time.now.utc.to_i, :secret_key => nil
+      update_fields :state => :received, :received => Time.now.utc.to_i, :secret_key => ""
     end
     def burned!
       # Make sure we don't go from :shared to :viewed
       return unless state?(:new) || state?(:viewed)
       @state = :burned
-      update_fields :state => :burned, :burned => Time.now.utc.to_i, :secret_key => nil
+      update_fields :state => :burned, :burned => Time.now.utc.to_i, :secret_key => ""
     end
     def state? guess
       state.to_s == guess.to_s
--- a/lib/onetime.rb
+++ b/lib/onetime.rb
@@ -100,7 +100,7 @@ module Onetime
           OT.ld format('Connecting to %s (%s)', uri, ping_result)
         }
       rescue Redis::CannotConnectError => e
-        OT.le "Cannot connect to redis #{Familia.id} (#{e.class})"
+        OT.le "Cannot connect to redis #{Familia.uri} (#{e.class})"
         exit 1
       rescue StandardError => e
         OT.le "Unexpected error `#{e}` (#{e.class})"
--- a/lib/onetime/logic.rb
+++ b/lib/onetime/logic.rb
@@ -51,7 +51,7 @@ module Onetime
         @cust = OT::Customer.create custid
         cust.update_passphrase password
         sess.update_fields :custid => cust.custid #, :authenticated => 'true'
-        cust.update_fields :planid => @plan.planid, :verified => false
+        cust.update_fields :planid => @plan.planid, :verified => "false"
         metadata, secret = Onetime::Secret.spawn_pair cust.custid, [sess.external_identifier]
         msg = "Thanks for verifying your account. "
         msg << %Q{We got you a secret fortune cookie!\n\n"%s"} % OT::Utils.random_fortune
--- a/lib/onetime/models.rb
+++ b/lib/onetime/models.rb
@@ -46,7 +46,7 @@ module Onetime::Models
   module Passphrase
     attr_accessor :passphrase_temp
     def update_passphrase v
-      self.passphrase_encryption = 1
+      self.passphrase_encryption = "1"
       @passphrase_temp = v
       self.passphrase = BCrypt::Password.create(v, :cost => 12).to_s
     end
--- a/lib/onetime/models.rb
+++ b/lib/onetime/models.rb
@@ -149,7 +149,7 @@ module Onetime::Models
       refresh_cache unless !instance_value.nil? || self.cache.has_key?(field)
       ret = case last_char
       when '='
-        self[field] = self.cache[field] = args.first
+        self[field] = self.cache[field] = args.first.to_s
       when '!'
         self.delete(field) and self.cache.delete(field) # Hash#delete returns the value
       when '?'
--- a/try/20_metadata_try.rb
+++ b/try/20_metadata_try.rb
@@ -12,9 +12,9 @@ OT.boot!
 #=> ['metadata:ivfn09cpriklqii1zagw6fc96suh8bp:object', 7, nil, {}]
 
 ## Keys don't change with values
-@metadata.secret_key = :hihi
+@metadata.secret_key = "hihi"
 [@metadata.rediskey, @metadata.secret_key, @metadata.all]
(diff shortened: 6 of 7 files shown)


sebastien-coavoux:
@delano the familia update with the exception raised prevent any boolean field to be stored in redis hashmap. 
This breaks the app from the start (authenticated field for a customer is a boolean) and most of the app itselft later. 

Not sure how you want to handle the serialization of boolean but as of now Familia is not able to write into redis. 
Either we convert any boolean / nil field before writing hasmap or not use boolean in any object that end up in redis 

I pushed another fix but not sure if that covers all

delano:
> Not sure how you want to handle the serialization of boolean but as of now Familia is not able to write into redis.

Do you mean Familia is not saving any fields correctly or that it's not saving booleans correctly?

> (authenticated field for a customer is a boolean)

I'm going to look at this now and fix. Please let me know of other boolean/nil field issues you find. 

> most of the app itselft later.

Hopefully not most 😅. Raising the exception is the right thing for Familia to do as a library. So it's okay that Onetime Secret needs to be updated -- it's actually improving the codebase as we remove the ambiguity.

sebastien-coavoux:
> Do you mean Familia is not saving any fields correctly or that it's not saving booleans correctly?
> 

Yes meant the Boolean and nil
> I'm going to look at this now and fix. Please let me know of other boolean/nil field issues you find.
> 
> Hopefully not most 😅. Raising the exception is the right thing for Familia to do as a library. So it's okay that Onetime Secret needs to be updated -- it's actually improving the codebase as we remove the ambiguity.

I tagged you in the commit I did to be more accurate so that we are on the same page. So far the "basic" flow was working (login as "root" -> create a secret -> read it)

I havent played with expiration or password parameter for instance so that might contain boolean / nil

Dockerfile
```diff
@@ -118,7 +118,8 @@ RUN set -eux && \
 # Create the directories that we need in the following image
 RUN echo "Creating directories"
 RUN mkdir -p "$CODE_ROOT"
-RUN mkdir -p "$ONETIME_HOME/{log,tmp}"
+RUN mkdir -p "$ONETIME_HOME/log"
+RUN mkdir -p "$ONETIME_HOME/tmp"
```
delano: Let's keep the PR focused on the main purpose and remove this and the docker compose file changes. 

```suggestion
RUN mkdir -p "$ONETIME_HOME/{log,tmp}"
```

config.ru
```diff
@@ -20,6 +20,8 @@ require 'onetime/app/web'
 require 'onetime/app/api'
 require 'onetime/app/colonel'
 
+# Uncomment for route debugging
+#Otto.debug = true
```
delano: Extra doc lines become cruft quickly (even when they're helpful). It's best to keep them out of the codebase from the start. 

```suggestion
```

docker-compose.yml
```diff
@@ -36,7 +36,10 @@ services:
       - redis
     env_file:
       - .env
-    entrypoint: ["./bin/entrypoint.sh"]
+    command:
+      - bash
+      - -c
+      - "cd /src && rm -rf /app/ && cp -r /src /app && cd /app ; exec ./bin/entrypoint.sh"
```
delano: Couple things here:
- This docker compose config is meant as a starting point. Overriding the entrypoint script with a command is exactly how to customize the startup behaviour. 
- We'll be moving the docker compose config to its own repo soon. That will open the door for more example configurations, rather than just having this single one. It would be great to get your input there when the time comes. 


```suggestion
    entrypoint: ["./bin/entrypoint.sh"]
```

docker-compose.yml
```diff
@@ -63,8 +66,12 @@ services:
     # the container, which is where the webapp lives. This is a
     # convenient way to develop in place without having to rebuild
     # the image each time you make a change.
+    # Because of otto behavior, that check ownership of files,
+    # we must copy the file to have the right user at runtime
+    # The other option would be to run the app as the same uid as the user on host
+    # Which is host dependent.
```
delano: I think this could be related to an open issue with #326 images and icons not showing up properly. I haven't been able to replicate it myself. It would helpful to follow-up on this feedback through that issue. 

```suggestion
```

docker-compose.yml
```diff
@@ -63,8 +66,12 @@ services:
...
     volumes:
-      - .:/app
+      - .:/src
```
delano:
```suggestion
      - .:/app
```

lib/onetime/app/helpers.rb
```diff
@@ -132,7 +132,7 @@ def check_session!
       if req.cookie?(:sess) && OT::Session.exists?(req.cookie(:sess))
         @sess = OT::Session.load req.cookie(:sess)
       else
-        @sess = OT::Session.create req.client_ipaddress, req.user_agent
+        @sess = OT::Session.create req.client_ipaddress, "anon", req.user_agent
```
delano, line 135: Top quality bug fix 👌

lib/onetime/models/metadata.rb
```diff
@@ -77,13 +77,13 @@ def received!
       # Make sure we don't go from :shared to :viewed
       return unless state?(:new) || state?(:viewed)
       @state = :received
-      update_fields :state => :received, :received => Time.now.utc.to_i, :secret_key => nil
+      update_fields :state => :received, :received => Time.now.utc.to_i, :secret_key => "null"
```
delano: Let's go with empty strings. Like you mention in your comment, "null" then becomes a *special* string. It's unlikely that we'd ever need to store a literal value of "null" but it's still a potential pitfall.

```suggestion
      update_fields :state => :received, :received => Time.now.utc.to_i, :secret_key => ""
```

lib/onetime/models/metadata.rb
```diff
@@ -77,13 +77,13 @@ def received!
...
     end
     def burned!
       # Make sure we don't go from :shared to :viewed
       return unless state?(:new) || state?(:viewed)
       @state = :burned
-      update_fields :state => :burned, :burned => Time.now.utc.to_i, :secret_key => nil
+      update_fields :state => :burned, :burned => Time.now.utc.to_i, :secret_key => "null"
```
delano:
```suggestion
      update_fields :state => :burned, :burned => Time.now.utc.to_i, :secret_key => ""
```

delano (changes requested):
Thanks for getting these changes into a PR so quick. Really nice find with that OT::Session.create bug. 

With a few changes, we'll be near ready to merge.

sebastien-coavoux:
Looking at the failed test, I am wondering if i should copy paste the if v.class I did in familia back to [models.rb](https://github.com/onetimesecret/onetimesecret/pull/427/files#diff-50ea411bf71ca7fa7e75007ce5c4e30dd42906e3c60f8287db9a980f24a5f4d7R152)

```
  ## Can store a one-way, encrypted passphrase               
16   s = Onetime::Secret.new :shared
17   s.update_passphrase 'poop'
18   [s.passphrase_encryption, s.passphrase?('poop')]
19   #=> [1, true]
     !=  ["1", true]
```

delano:
Sorry for the delay. 

Nice, both of these are simple fixes without further repercussions. 

> Looking at the failed test, I am wondering if i should copy paste the if v.class I did in familia back to [models.rb](https://github.com/onetimesecret/onetimesecret/pull/427/files#diff-50ea411bf71ca7fa7e75007ce5c4e30dd42906e3c60f8287db9a980f24a5f4d7R152)
> 
> ```
>   ## Can store a one-way, encrypted passphrase               
> 16   s = Onetime::Secret.new :shared
> 17   s.update_passphrase 'poop'
> 18   [s.passphrase_encryption, s.passphrase?('poop')]
> 19   #=> [1, true]
>      !=  ["1", true]
> ```

Yeah similar fix here. In `Onetime::Models::Passphrase#update_passphrase`, set the number as a string (`self.passphrase_encryption = "1"`). 

> for this one, not sure. Should the key is a Ruby be symbol and not a String ?
> 
> ```
>  ## Keys don't change with values                           
> 15   @metadata.secret_key = :hihi
> 16   [@metadata.rediskey, @metadata.secret_key, @metadata.all]
> 17   #=> ['metadata:ivfn09cpriklqii1zagw6fc96suh8bp:object', :hihi, {"secret_key"=>"hihi"}]
>      !=  ["metadata:ivfn09cpriklqii1zagw6fc96suh8bp:object", "hihi", {"secret_key"=>"hihi"}]
> ```

This one is a just a poorly written test. No where else in the codebase sets `secret_key` to a symbol. It's all strings. It's a rare case where updating the failing test expectation is the right thing to do 👍

delano (approved):
Good work 👌

sebastien-coavoux:
Test fixed accordingly

delano (approved):
Success 😺

PR #1037, opened 2025-01-31
fix: .env.example and support/smtp_test.rb uses `FROM_EMAIL`, but the actual app appears to use 'FROM'

venkatamutyala:
This was originally brought up to me here: https://github.com/onetimesecret/helm-chart/issues/27#issuecomment-2626250674

@delano could you back port this into a patch for v0.18.X this way i can update the help chart with v0.18.X as well? Alternatively, we can just force everyone to use v0.19.X but given it was just released I'm unclear on how stable it is so will differ to you on this.

### DIFF
--- a/etc/config.example.yaml
+++ b/etc/config.example.yaml
@@ -113,7 +113,7 @@
   # Production Settings (for reference)
   # ----------------------------------
   :mode: <%= ENV['EMAILER_MODE'] || 'smtp' %>
-  :from: <%= ENV['FROM'] || 'CHANGEME@example.com' %>
+  :from: <%= ENV['FROM_EMAIL'] || ENV['FROM'] || 'CHANGEME@example.com' %>
   :fromname: <%= ENV['FROMNAME'] || 'Support' %>
   :host: <%= ENV['SMTP_HOST'] || 'smtp.provider.com' %>
   :port: <%= ENV['SMTP_PORT'] || 587 %>
--- a/tests/unit/ruby/config.test.yaml
+++ b/tests/unit/ruby/config.test.yaml
@@ -73,7 +73,7 @@
   - 'CHANGEME@EXAMPLE.com'
 :emailer:
   :mode: <%= ENV['EMAILER_MODE'] || 'smtp' %>
-  :from: <%= ENV['FROM'] || 'tests@example.com' %>
+  :from: <%= ENV['FROM_EMAIL'] || ENV['FROM_EMAIL'] || 'tests@example.com' %>
   :fromname: <%= ENV['FROMNAME'] || 'Jan' %>
   :host: <%= ENV['SMTP_HOST'] || 'localhost' %>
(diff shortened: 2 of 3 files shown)


etc/config.example.yaml
```diff
@@ -113,7 +113,7 @@
   # Production Settings (for reference)
   # ----------------------------------
   :mode: <%= ENV['EMAILER_MODE'] || 'smtp' %>
-  :from: <%= ENV['FROM'] || 'CHANGEME@example.com' %>
+  :from: <%= ENV['FROM_EMAIL'] || 'CHANGEME@example.com' %>
```
delano:
```suggestion
  :from: <%= ENV['FROM_EMAIL'] || ENV['FROM'] || 'CHANGEME@example.com' %>
```

tests/unit/ruby/try/16_config_emailer_try.rb
```diff
@@ -43,7 +43,7 @@
 
 ## Emailer values can be set via environment variables
 ENV['EMAILER_MODE'] = 'sendmail'
-ENV['FROM'] = 'test@example.com'
+ENV['FROM_EMAIL'] = 'test@example.com'
```
delano: Unless this was causing issues, let's leave the tests as is. 

```suggestion
ENV['FROM'] = 'test@example.com'
```

tests/unit/ruby/try/16_config_emailer_try.rb
```diff
@@ -70,7 +70,7 @@
 
 ## Clearing environment variables restores default values
 ENV.delete('EMAILER_MODE')
-ENV.delete('FROM')
+ENV.delete('FROM_EMAIL')
```
delano:
```suggestion
ENV.delete('FROM')
```

tests/unit/ruby/config.test.yaml
```diff
@@ -73,7 +73,7 @@
   - 'CHANGEME@EXAMPLE.com'
 :emailer:
   :mode: <%= ENV['EMAILER_MODE'] || 'smtp' %>
-  :from: <%= ENV['FROM'] || 'tests@example.com' %>
+  :from: <%= ENV['FROM_EMAIL'] || 'tests@example.com' %>
```
delano:
```suggestion
  :from: <%= ENV['FROM_EMAIL'] || ENV['FROM_EMAIL'] || 'tests@example.com' %>
```

delano (approved):
Thanks for catching this discrepancy. I added a suggestion to use both env vars in the config so that we don't cause a similar issues with environments that are using `FROM`. 

If you don't mind committing the updates and rebasing to a single commit then I can cherry pick it into a v0.18.4 release.


And yeah, good call re: waiting a while for v0.19 since it's so hot off the presses (and it has _a lot_ of code movement).

venkatamutyala:
@delano should be all set.

PR #1771, opened 2025-10-04
Fix password generation config key handling for v0.22

delano:
This fixes a configuration key mismatch that was preventing password generation settings from being properly applied in v0.22.

The root cause was that the configuration uses symbol keys (`:password_generation`, `:character_sets`) but the code was trying to merge them with string keys from the payload. This meant the config values were being ignored and only payload values (if provided) would work.

## What changed

Standardized on string keys throughout the password generation flow:

- **`generate_secret.rb`**: Convert both config and payload keys to strings before merging, ensuring consistent key access regardless of source
- **`utils.rb`**: Updated `strand()` to use string keys internally and handle both symbol/string input for backward compatibility  
- **`config.example.yaml`**: Changed symbols default from `false` to `true` (matching environment variable behavior)

The string key approach maintains compatibility with v0.22's symbol-based config while preparing for v1.0's string-based config system.

## Testing

The fix ensures password generation works correctly with configuration settings and properly respects both config defaults and payload overrides.

### DIFF
--- a/apps/api/v2/logic/secrets/generate_secret.rb
+++ b/apps/api/v2/logic/secrets/generate_secret.rb
@@ -12,12 +12,21 @@ module V2::Logic
         # Get password generation configuration
         password_config = OT.conf.dig(:site, :secret_options, :password_generation) || {}
 
-        # Extract parameters from payload with fallbacks to configuration
-        length = payload[:length]&.to_i || password_config[:default_length] || 12
+        # Convert config keys to strings and merge with payload
+        #
+        # This is compatable with both v0.22 (mostly symbols) and v1.0
+        # configuration (all strings.
+        config_with_string_keys = password_config.transform_keys(&:to_s)
+        payload_with_string_keys = payload.transform_keys(&:to_s)
+        merged_options = config_with_string_keys.merge(payload_with_string_keys)
 
-        # Build character set options from payload or configuration
-        char_sets = payload[:character_sets] || password_config[:character_sets] || {}
+        # Extract parameters from merged options
+        length = merged_options['length']&.to_i || merged_options['default_length'] || 12
 
+        # Build character set options from merged configuration
+        char_sets = merged_options['character_sets'] || {}
+
+        OT.li "[GenerateSecret] Using the character sets: #{char_sets.inspect}"
         # Use the configurable password generation method
         @secret_value = Onetime::Utils.strand(length, char_sets)
       end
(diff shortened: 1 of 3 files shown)


delano:
## PR Review Feedback Analysis

I've analyzed the automated review suggestions and implemented the valid fix:

### ✅ Implemented: Suggestion #2 - Fix password complexity guarantee

**Valid bug found!** The original code had a subtle issue:

```ruby
# Old code - BROKEN
password_chars << UPPERCASE.sample if opts['uppercase']
# ... sample from other sets ...
if opts['exclude_ambiguous']
  password_chars.delete_if { |char| AMBIGUOUS_CHARS.include?(char) }
end
```

**Problem**: If the sampled character happened to be ambiguous (e.g., 'O' from UPPERCASE, '0' from NUMBERS), it would be removed, breaking the complexity guarantee.

**Fix applied** (c00cb09): Sample directly from non-ambiguous subsets when `exclude_ambiguous` is enabled:

```ruby
if opts['exclude_ambiguous']
  password_chars << (UPPERCASE - AMBIGUOUS_CHARS).sample if opts['uppercase']
  # ... etc for other sets
else
  password_chars << UPPERCASE.sample if opts['uppercase']
  # ... etc for other sets
end
```

**Verification**: Tested with 100 iterations confirming all character sets are represented and no ambiguous characters appear.

---

### ❌ Rejected: Suggestion #1 - Validate length parameter

**Analysis**: This is a false positive. The code already handles edge cases correctly:

- `merged_options['length']&.to_i` safely returns `nil` or integer (even empty string becomes `0`)
- The fallback chain `|| merged_options['default_length'] || 12` ensures we always get a value
- `strand()` already validates `len.positive?` and raises `ArgumentError` if needed

The suggested fix adds unnecessary complexity without providing additional safety.

---

### ❌ Rejected: Suggestion #3 - Validate options type early

**Analysis**: The current defensive behavior is more appropriate for this utility function:

- **Current**: Silently falls back to `{}` (safe defaults) for unexpected input
- **Suggested**: Raise error for unexpected input

**Reasoning**:
- This is a utility method with backward compatibility concerns (boolean vs hash)
- Sensible defaults prevent crashes from unexpected input
- The method will still produce a valid, secure password
- Defensive programming is preferred over fail-fast for utility functions with safe fallback behavior

The current implementation is more robust.

---

**Summary**: 1 valid fix implemented, 2 suggestions rejected as inappropriate for this context.

PR #1975, opened 2025-11-18
Complete translations for ten locales

delano:


delano:
Branched off of main instead of develop.

PR #2032, opened 2025-11-23
Review admin features

delano:
USER MANAGEMENT:
- List all users with filtering (GET /colonel/users)
- View user details and statistics (GET /colonel/users/:id)
- Update user plans (POST /colonel/users/:id/plan)

SYSTEM MONITORING:
- Database metrics from Redis INFO (GET /colonel/system/database)
- Redis metrics and stats (GET /colonel/system/redis)

IP BANNING:
- List banned IPs (GET /colonel/banned-ips)
- Ban IP addresses (POST /colonel/banned-ips)
- Unban IP addresses (DELETE /colonel/banned-ips/:ip)
- Middleware to block banned IPs

USAGE EXPORT:
- Export usage data by date range (GET /colonel/usage/export)
- Daily statistics for secrets and users

All endpoints require colonel role and use sessionauth.

### DIFF
--- a/apps/web/auth/config/hooks/audit_logging.rb
+++ b/apps/web/auth/config/hooks/audit_logging.rb
@@ -99,8 +99,7 @@ module Auth::Config::Hooks
       # ========================================================================
 
       auth.audit_log_message_for :login do
-        mfa_enabled          = auth.respond_to?(:otp_exists?) && auth.otp_exists?
-        mfa_required = mfa_enabled
+        mfa_required = auth.respond_to?(:otp_exists?) && auth.otp_exists?
         if mfa_required
           'Login successful - MFA required'
         else
--- a/src/views/colonel/ColonelDomains.vue
+++ b/src/views/colonel/ColonelDomains.vue
@@ -24,6 +24,7 @@
       case 'resolving':
         return 'bg-blue-100 text-blue-800 dark:bg-blue-900 dark:text-blue-200';
       default:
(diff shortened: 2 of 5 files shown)


delano:
## Performance Optimization - Future Work

**Regarding the secret count calculation in `list_users.rb`:**

The review correctly identified that the current user listing implementation has O(n*m) complexity for secret counting (line 50-54). We acknowledge this is inefficient for large user bases.

**Current Approach:**
```ruby
user_secrets_count = Onetime::Secret.new.dbclient.keys('secret*:object').select do |key|
  objid = key.split(':')[1]
  secret = Onetime::Secret.load(objid)
  secret&.owner_id == cust.objid
end.count
```

**Deferred Optimization:**
This will be addressed in future work once we better understand colonel usage patterns and can implement an appropriate solution (e.g., maintaining per-user counters, caching strategies, or denormalized counts).

For the initial release, colonel admin operations are expected to be infrequent administrative tasks rather than high-traffic endpoints, so this performance trade-off is acceptable.

**Tracking:** Will create a follow-up issue for performance optimization after initial deployment.

---

**Other review feedback addressed:**
- ✅ Fixed UI state conflicts in ColonelUsers/ColonelSecrets (empty state now uses `v-else-if`)
- ✅ Fixed redundant variable assignment in audit_logging.rb
- ✅ Added warning logging for unknown domain states
- ✅ Cleaned up dead code in validate_admin_interface.rb

PR #2059, opened 2025-11-26
Improve test isolation and skip Stripe during test boot 

delano:
- **Add unified email system with ERB templates and multi-provider support**
- **Improve test isolation and skip Stripe during test boot**


delano:
This branch was already merged into the #2053

PR #2060, opened 2025-11-26
dreamy banzai

delano:
- **Add unified email system with ERB templates and multi-provider support**
- **Improve test isolation and skip Stripe during test boot**
- **Fix test infrastructure for Familia v2 compatibility**
- **Remove natural_time calls from Colonel API**
- **Fix to prevent default CTA in EmptyState from showing**
- **Refactor incoming secrets error handling architecture**
- **Hide page header when incoming secrets feature is disabled**
- **Progress on vue tests**


delano:
Already merged into #2054

PR #2506, opened 2026-02-13
Harden v0.24 migration: fix index bugs, add validators, improve test coverage

delano:
## Summary

Comprehensive hardening of the Familia v1→v2 migration pipeline. This fixes three critical data consistency bugs, rewrites the enrichment layer for safety and performance, and adds 11+ validators that detect inconsistencies across all major model types (Customer, Organization, CustomDomain, Receipt, Secret).

- **Fix CustomDomain dual-index consistency** — `create!`, `destroy!`, and `claim_orphaned_domain` now properly maintain both the Familia unique_index and manual sorted-set/hash indexes. Previously, destroying a domain left orphaned index entries that prevented re-creation.
- **Fix migration field mapping** — Receipt `secret_shortkey` → `secret_shortid` rename was missing from DIRECT_COPY_FIELDS, causing null display in the frontend. Added safe_dump fallback deriving from `secret_identifier[0..7]`.
- **Fix customdomain index ID mismatch** — `create_indexes.rb` was looking for `custom_domain:values` instead of `customdomain:values`, and copying hex IDs instead of UUID v7s into the V2 instance index.
- **Rewrite enrichment as direct Redis RESTORE** — Replaced in-memory deserialization with two-phase streaming (mapping build → RESTORE with 30-day TTL). No Ruby memory overhead for large datasets.
- **Fix org and customdomain validators** — Were expecting Redis DUMP blobs for v2-only keys; rewrote to read from JSONL index files matching the actual migration output format.

## Validators added

Phase-specific validators run after each migration step and halt the pipeline on failure:

| Phase | Validator | Checks |
|-------|-----------|--------|
| 01-customer | instance_index | ZSET membership ↔ object existence |
| 02-organization | instance_index | Bidirectional org-customer-domain references (412 lines) |
| 03-customdomain | instance_index | Dual-index + participation + org membership (826 lines) |
| 04-receipt | instance_index | Receipt-secret-domain consistency |
| 04-receipt | domain_participation | Three-way receipt-domain-ZSET parity |
| 05-secret | instance_index | Secret instance checks |

Standalone post-migration validators for auditing:
- `detect_phantom_receipts.rb` — Six-category inconsistency detection
- `validate_display_domain_index.rb` — Stale/mismatch/missing domain entries
- `validate_org_domain_membership.rb` — Phantom/orphan/stale domain-org relationships
- `validate_orphan_domain_zset_leaks.rb` — Nil org_id entries in org sorted sets
- `validate_instances_owners_parity.rb` — Symmetric difference between instances and owners

## Model changes

**CustomDomain**: Enhanced `create!` with atomic HSETNX uniqueness, `destroy!` with dual-index cleanup, `claim_orphaned_domain` with WATCH/MULTI and proper owners-hash update. Added `update_display_domain()` with rollback.

**with_migration_fields**: Changed `_original_record` (jsonkey/JSON blob) → `_original_object` (hashkey/direct RESTORE). Supports model-specific original data types (`_original_brand`, `_original_logo`, `_original_icon`).

**CLI sync-auth-accounts**: Batched transactions (1000/batch) with resume capability via existing external_id skip-set.

## Test coverage

12 new test files (~1,200 lines):
- CustomDomain lifecycle: create/destroy/re-create, rename, orphan claim, dual-index parity
- Receipt: phantom entries, domain participation drift, safe_dump fallback
- Migration fields: hashkey API, field access patterns
- Frontend: SecretReceiptTableItem Vue component spec

## Test plan

- [ ] Run migration pipeline end-to-end on staging data with validators enabled
- [ ] Verify CustomDomain CRUD operations (create, rename, destroy, re-create)
- [ ] Confirm orphan domain claim updates all three index systems
- [ ] Validate receipt display shows secret_shortid (not "null")
- [ ] Check enrichment RESTORE produces correct `_original_object` keys with TTL
- [ ] Run standalone validators against production data snapshot


delano:
## Automated Review Responses

Thanks for the automated reviews. After verifying each item against the actual code, all 12 suggestions are either already addressed or not applicable. Responses grouped by category:

### Already Addressed in Code

**Gemini — Log rollback failure** (`custom_domain.rb:220`): Already implemented at line 219:
```ruby
OT.le "[CustomDomain.update_display_domain] Rollback save failed: #{rollback_ex.message}"
```

**Copilot — create! rollback missing instances/owners cleanup** (`custom_domain.rb:~624`): The rollback block at lines 636–639 already includes all four cleanup operations:
```ruby
display_domains.remove(normalized_domain)
obj.remove_from_class_display_domain_index
instances.remove(obj.to_s)
owners.remove(obj.to_s)
```

**Qodo — Stale domain parts after rename** (`custom_domain.rb:222`): Lines 195–200 explicitly re-parse derived fields via `PublicSuffix.parse(new_domain)` and update `@base_domain`, `@subdomain`, `@trd`, `@tld`, `@sld`.

**Copilot — update_display_domain ordering** (`custom_domain.rb:222`): Both `save` (line 203) and `display_domains.put` (line 204) are inside the same `begin/rescue` block. If `put` fails after `save` succeeds, the rescue catches it, restores all field values, re-adds the old display_domains entry, and re-saves to restore the auto-index.

### Not Applicable

**Qodo — `require 'set'`** (`sync_auth_accounts_command.rb:126`): Ruby 3.4 includes `Set` in core — no require needed. Confirmed: `ruby -e "puts Set.new([1,2,3]).inspect"` works without require on Ruby 3.4.8.

**Gemini — N+1 Redis in process_batch** (`sync_auth_accounts_command.rb:204`): This is a one-shot CLI migration command, not a hot path. The per-record `Customer.load` also allows skip logic (anonymous, system, invalid email) before any DB work. Batch-loading all customers would load records we immediately discard.

**Qodo — process_batch lacks StandardError rescue** (`sync_auth_accounts_command.rb:305`): The `rescue Sequel::Error => ex` at line 298 handles the database transaction failure case, then re-raises. Non-Sequel errors (Redis, runtime) propagate up naturally and abort the command — which is the correct behavior for a batch migration tool. Catching and continuing on arbitrary errors would mask real problems.

**Copilot — Race condition in claim_orphaned_domain** (`custom_domain.rb:694`): The org nil check is at lines 680–683, immediately after the load and before MULTI:
```ruby
org = Onetime::Organization.load(org_id)
unless org
  dbclient.unwatch
  raise Onetime::Problem, "Organization #{org_id} not found"
end
```
The `if org` guard on line 690 is belt-and-suspenders. The code flow the reviewer described doesn't occur.

**Gemini — Method signature size** (`sync_auth_accounts_command.rb:196`): Style preference. This is a private method in a single-operation CLI command — the explicit parameter list makes the data flow traceable without hunting for instance variable assignments.

**Copilot — Duplicate detection within batch** (`sync_auth_accounts_command.rb:292`): Within-batch duplicates can't occur: `batch_ids` comes from `Customer.values.members.each_slice(BATCH_SIZE)` which is a Redis SET (unique by definition). The `batch_processed_extids` array tracks what to merge into `existing_extids` after commit for cross-batch resume.

**Copilot — Metadata dir config** (`enrich_with_original_record.rb:84`): `dir: 'metadata'` is correct — the v1 dump lives at `data/upgrades/v0.24.0/metadata/metadata_dump.jsonl`. The `04-receipt/` directory is for the transform and index scripts, not the dump source.

**Copilot — Vue `??` vs `||`** (`SecretReceiptTableItem.vue:38`): Intentional. `??` is correct here — an empty string `secret_shortid` is a valid (if unusual) value that should display as-is. Only null/undefined (missing from migration) should show the em-dash fallback.

PR #2511, opened 2026-02-15
Fix incoming secret response schema nullability

delano:
## Summary

The `incomingSecretResponseSchema` Zod validator rejected valid 200 responses from the v2 guest/secret/conceal endpoint, surfacing as "Invalid data received. Expected string but got null" (and later "...got array") in the UI.

Two issues:
- Several metadata/secret fields used `z.string()` or `z.string().optional()` but the backend returns `null` — Zod's `.optional()` accepts `undefined`, not `null`
- `details.recipient` was typed as a string but the backend returns an array of email addresses

## Changes

- `metadataRecordSchema`: `state`, `secret_shortkey`, `memo`, `recipients` → `.nullish()`
- `secretRecordSchema`: `state` → `.nullish()`
- `details.recipient` → `z.array(z.string()).nullish()`
- Added missing `kind` and `recipient_safe` fields to details object
- 41 Vitest tests covering all four exported schemas, focused on nullish handling and the array-typed recipient fix

## Test plan

- [ ] `pnpm test` passes (41 new tests in `tests/unit/vue/schemas/incoming.spec.ts`)
- [ ] Submit an incoming secret on dev — no validation error, navigates to success view

### DIFF
--- a/src/schemas/api/incoming.ts
+++ b/src/schemas/api/incoming.ts
@@ -47,11 +47,11 @@ const metadataRecordSchema = z.object({
   identifier: z.string(),
   key: z.string(),
   custid: z.string(),
-  state: z.string(),
-  secret_shortkey: z.string(),
+  state: z.string().nullish(),
+  secret_shortkey: z.string().nullish(),
   shortkey: z.string(),
-  memo: z.string().optional(),
-  recipients: z.string().optional(),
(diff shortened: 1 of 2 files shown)


delano:
Wrong base branch — this fix targets rel/0.23, not main.

Issue #2500, opened 2026-02-11
Incoming not working on v0.23.3


hogduske:
### What happened?

I am getting the follwing error in the footer when trying to user /incoming to send a secret to a predetermined recipient: 
`Invalid data received. Expected string but got null.`

Log output showing:
```
10.255.2.2 - - [11/Feb/2026:18:00:01 +0000] "POST /api/v2/validate-shrimp HTTP/1.1" 200 79 0.0125
10.255.2.2 - - [11/Feb/2026:18:00:03 +0000] "POST /api/v2/guest/secret/conceal HTTP/1.1" 200 1043 0.0350
```

### Steps to reproduce

Version v0.23.3
Out of the Box config with docker compose.
Set up email sending correctly.

### Additional context

Email is working correctly when sending normal secrets.
API and guest routes are enabled in the config.

Incoming section of the config:
```
  incoming:
    enabled: true
    memo_max_length: 300
    default_passphrase: test
    default_ttl: 604800
    recipients:
      - email: 'support@example.com'
        name: 'Support'
      - email: 'john.doe@example.com'
        name: 'John Doe'

```

API section of the config:
```
    api:
      enabled: true
      guest_routes:
        enabled: true
        conceal: true
        generate: false
        show:  false
        reveal: false
        receipt: false
        burn: false
```

delano:
Thanks for quality report. This is fixed in [v0.23.4](https://github.com/onetimesecret/onetimesecret/releases/tag/v0.23.4). The images are building now.

hogduske:
@delano Thanks for you swift fix of this issue.
There is now another issue with /incoming 
Secret creation works but there is no email being sent out.
The log shows creation of the secret but the mailer is not called upon.

Log:
```
10.255.2.2 - - [15/Feb/2026:22:01:41 +0000] "GET /incoming HTTP/1.1" 200 10729 0.0800
10.255.2.2 - - [15/Feb/2026:22:01:41 +0000] "GET /api/v2/incoming/config HTTP/1.1" 200 192 0.0202
10.255.2.2 - - [15/Feb/2026:22:01:47 +0000] "POST /api/v2/guest/secret/conceal HTTP/1.1" 200 1079 0.0649
10.255.2.2 - - [15/Feb/2026:22:01:47 +0000] "GET /api/v2/guest/receipt/6a8jrngy73qxnzmzn7iu9st3gb6fo3w HTTP/1.1" 200 1512 0.0161
```

[...]

PR #2536, opened 2026-02-18
Fix currency conflict regex to match Stripe's current error format

delano:
Stripe changed the error message format for currency conflicts, which broke our detection regex. The old format included both currencies in the message; the new format only mentions the account's currency, so we need to supply the requested currency separately.

## What changed

**Currency conflict detection** (`currency_migration_service.rb`): Updated `CURRENCY_CONFLICT_PATTERN` to handle both the old Stripe format (`"had a subscription or payment in EUR...pay in USD"`) and the new format (`"You cannot combine currencies...with currency USD"`). When the requested currency isn't present in the error message, it now accepts a `requested_currency_hint` parameter.

**Billing controller** (`billing.rb`): Passes `requested_currency_hint: plan.currency` to `parse_currency_conflict()` so the new Stripe format is handled correctly. Also tightened error logging to use structured `error_class`/`error` fields and made the user-facing error message generic rather than leaking raw exception details.

**Schema cleanup** (`incoming.ts`): Reverted `.nullish()` on `state`, `details.memo`, and `details.recipient` — these fields must be strings when present, not null. This narrows the API contract back to what the backend actually sends.

**Test cleanup** (`incomingSchema.spec.ts`): Removed the 217-line test file that was specifically testing null-handling behavior for issue #2500, which is now resolved.

## Test plan

- [ ] Verify Stripe currency conflict errors are correctly detected with the new message format
- [ ] Confirm fallback to `requested_currency_hint` when regex can't extract currency from message
- [ ] Confirm user-facing error message is generic and doesn't leak Stripe details

### DIFF
--- a/apps/web/billing/controllers/billing.rb
+++ b/apps/web/billing/controllers/billing.rb
@@ -343,6 +343,7 @@ module Billing
               interval: plan.interval,
               amount: plan.amount,
               currency: plan.currency,
+              region: plan.region,
               features: plan.features.to_a,
               limits: plan.limits_hash.transform_values { |v| v == Float::INFINITY ? -1 : v },
               entitlements: plan.entitlements.to_a,
--- a/apps/web/billing/controllers/billing.rb
+++ b/apps/web/billing/controllers/billing.rb
@@ -193,7 +193,10 @@ module Billing
         json_error(ex.message, status: 403)
       rescue Stripe::InvalidRequestError => ex
         if Billing::CurrencyMigrationService.currency_conflict?(ex)
-          currencies = Billing::CurrencyMigrationService.parse_currency_conflict(ex)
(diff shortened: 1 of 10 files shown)


delano:
Re: `return nil unless existing && requested` guard in `parse_currency_conflict`

The only call site (`billing.rb:196-199`) always passes `requested_currency_hint: plan.currency` — the hint is the plan the user was trying to purchase, which is available at the rescue site. So `requested_currency` cannot be `nil` in production with the current code.

The partial-result behavior (returning `{existing_currency: ..., requested_currency: nil}`) is intentional: a caller that only needs to know what currency a customer is locked to — e.g., for display or logging — would get `nil` back if we added the guard, which is worse than a partial hash. The `@param` docstring and the regex comment both document this.

The new tests explicitly document `requested_currency: nil` as correct behavior for the no-hint case, not a bug.

No change needed here.

Issue #2803, opened 2026-03-29
ValidateSenderDomain operation with provider strategies


delano:
## Parent

Blocked by / part of #2792
Depends on: Phase 1a (DomainEmailConfig model)

## Summary

Create the `ValidateSenderDomain` operation using the strategy pattern to support DNS-based domain verification across SES, SendGrid, and Lettermint providers.

## Context

This is **Phase 2** of #2792. It can run in parallel with Phases 1b and 3 once Phase 1a is complete. The operation follows the existing `VerifyDomain` operation pattern.

### Core Operation
- [ ] `lib/onetime/operations/validate_sender_domain.rb` — orchestrator that delegates to provider strategy
- [ ] Strategy selection based on `DomainEmailConfig#provider`
- [ ] Generate required DNS records (DKIM, SPF) for user to configure
- [ ] Query DNS to check verification status
- [ ] Update `DomainEmailConfig#verification_status` and `#verified_at`

### Provider Strategies
- [ ] `strategies/ses_validation.rb` — AWS SES identity verification (DKIM + SPF + MX)
- [ ] `strategies/sendgrid_validation.rb` — SendGrid domain authentication (DKIM + SPF)
- [ ] `strategies/lettermint_validation.rb` — Lettermint domain verification (DKIM + SPF)

### Verification Flow
1. User saves config with provider selection
2. Operation generates DNS records to display
3. User configures DNS at their registrar
4. User triggers validation check
5. Operation queries DNS and updates status

### Tests
- [ ] Unit tests for each provider strategy
- [ ] Integration tests for the orchestrator
- [ ] DNS record generation tests
- [ ] Status transition tests (pending → verified, pending → failed)

## References

- Existing domain verification: `lib/onetime/operations/verify_domain.rb`
- Provider comparison table in #2792

| Provider | Validation Method | Records |
|----------|------------------|---------|
| AWS SES | DNS (DKIM, SPF, MX) | 3 CNAME + 1 MX + 1 TXT |
| SendGrid | DNS (DKIM, SPF) | 3 CNAME + 1 TXT |
| Lettermint | DNS (DKIM, SPF) | 2 CNAME + 1 TXT |

## Acceptance Criteria

- Each provider strategy generates correct DNS records for display
- DNS verification correctly reports pass/fail per record type
- Verification status is persisted on the config model
- Strategy pattern allows adding new providers without modifying orchestrator

PR #2838, opened 2026-03-30
[#2834] Move sender domain DNS validation to background job

### LINKED ISSUE #2834: Move sender domain DNS validation to background job
## Problem Statement

`ValidateSenderDomain#call` performs sequential DNS lookups synchronously — up to 5 queries for SES. On a responsive resolver this completes in ~50-100ms, but under degraded conditions (slow resolver, timeouts), latency compounds: `Resolv::DNS` has a 5-second initial timeout, so 2 timed-out records add ~10-15s to the request. Worst case (resolver unreachable) is ~25s, likely hitting Rack/nginx timeouts first.

The operation's existing design (immutable `Result`, `persist: true` default, `pending`/`verified`/`failed` status progression) already supports async execution — the web request just needs to enqueue and return `pending`.

## Early Diagnosis

The current synchronous path is acceptable for initial ship since most DNS lookups complete well within a web request budget. The risk is the degraded/timeout path where sequential lookups compound additively. A background job eliminates that risk and fits the existing persistence model naturally.

Likely approach: enqueue a job from the controller action, return `pending` status immediately, have the UI poll or use SSE for completion. The `Result` and `persist_verification` machinery already handles the write-back.

## Affected Areas

- [ ] `lib/onetime/operations/validate_sender_domain.rb` — already has the persistence path; needs a job wrapper
- [ ] `lib/onetime/domain_validation/sender_strategies/base_strategy.rb` — `verify_all_records` sequential loop is the hot path
- [ ] Controller/route that triggers validation — needs to switch from sync call to job enqueue

Parent: #2803

delano:
## Summary

`ValidateSenderDomain` performs sequential DNS lookups synchronously — up to 5 for SES. Under degraded DNS conditions each lookup can block for 5s (Resolv::DNS default timeout), compounding to 25s worst case. This moves validation to a RabbitMQ background job so the web request returns `pending` immediately.

The operation's existing design (immutable `Result`, `persist: true`, `pending`/`verified`/`failed` status progression) already supported async execution — the web request just needed to enqueue and return.

## Changes

- **New endpoint** `POST /:extid/email-config/validate` — sets `verification_status = 'pending'`, enqueues job, returns immediately. Auth/entitlement checks via existing `authorize_sender_config!`.
- **New worker** `DomainValidationWorker` — consumes `domain.validation.check` queue, loads MailerConfig, delegates to `ValidateSenderDomain` with retry logic (2 retries, 2s exponential backoff for transient DNS flakiness). Follows the NotificationWorker pattern: idempotency claim, ack/reject, DLQ on permanent failure.
- **New queue config** `domain.validation.check` — durable, with `dlx.domain.validation` / `dlq.domain.validation` dead letter routing.
- **Publisher convenience method** `enqueue_domain_validation(domain_id)` — sync fallback when jobs disabled (dev/test without RabbitMQ), raises when RabbitMQ is unavailable but jobs are enabled.
- **Bugfix** — `ValidateSenderDomain#strategy` called `private_class_method` via explicit receiver, which Ruby disallows. Changed to `self.class.send(:resolve_strategy, ...)`. Pre-existing issue never hit because all callers passed an explicit strategy override.

## How polling works

No new polling endpoint needed. The existing `GET /:extid/email-config` already serializes `verification_status`, so the frontend can poll it to detect the transition from `pending` to `verified` or `failed`.

## Test plan

- [x] 13 new test cases: queue config, DLX/DLQ, QueueDeclarator integration, Publisher API, sync fallback behavior
- [x] 35 existing ValidateSenderDomain tests still pass
- [x] Full unit suite (2803 pass, 13 pre-existing locale failures unrelated to this change)
- [ ] Manual: trigger validation via POST endpoint, observe status transition via GET endpoint
- [ ] Manual: verify worker picks up messages with RabbitMQ running

Closes #2834
Parent: #2803

### DIFF
--- a/lib/onetime/jobs/workers/domain_validation_worker.rb
+++ b/lib/onetime/jobs/workers/domain_validation_worker.rb
@@ -43,6 +43,9 @@ module Onetime
 
         QUEUE_NAME = 'domain.validation.check'
 
+        # Conservative defaults for initial rollout. DNS-bound workers spend
+        # most time in I/O wait, so 8-16 threads would be safe here. Tune
+        # via env vars once production telemetry shows DNS response distributions.
         from_queue QUEUE_NAME,
           **QueueDeclarator.sneakers_options_for(QUEUE_NAME),
           threads: ENV.fetch('DOMAIN_VALIDATION_WORKER_THREADS', 2).to_i,
@@ -76,7 +79,7 @@ module Onetime
           # Load the mailer config for this domain
           mailer_config = Onetime::CustomDomain::MailerConfig.find_by_domain_id(domain_id)
           unless mailer_config
-            log_error "MailerConfig not found for domain_id: #{domain_id}"
+            log_error "MailerConfig not found for domain_id: #{domain_id}", message_id: message_id, metadata: message_metadata
             return ack! # Don't retry -- config won't appear on its own
           end
 
@@ -87,6 +90,10 @@ module Onetime
               mailer_config: mailer_config,
               persist: true,
             ).call
+            # Re-raise so with_retry can retry transient DNS failures.
+            # ValidateSenderDomain#call rescues internally and returns a
+            # Result — without this, with_retry never sees an exception.
+            raise result.error if result.error
           end
 
           log_info "Sender domain validation complete: #{domain_id}",
--- a/apps/api/domains/logic/sender_config/validate_sender_config.rb
+++ b/apps/api/domains/logic/sender_config/validate_sender_config.rb
@@ -52,12 +52,21 @@ module DomainsAPI
           OT.ld "[ValidateSenderConfig] Triggering async DNS validation for domain #{@domain_id} by user #{cust.extid}"
 
           # Set status to pending before enqueuing so the UI reflects immediately
+          previous_status                    = @mailer_config.verification_status
           @mailer_config.verification_status = VERIFICATION_STATUS_PENDING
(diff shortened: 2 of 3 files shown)


lib/onetime/jobs/workers/domain_validation_worker.rb
```diff
@@ -0,0 +1,106 @@
...
+      class DomainValidationWorker
+        include Sneakers::Worker
+        include BaseWorker
+
+        QUEUE_NAME = 'domain.validation.check'
+
+        from_queue QUEUE_NAME,
+          **QueueDeclarator.sneakers_options_for(QUEUE_NAME),
+          threads: ENV.fetch('DOMAIN_VALIDATION_WORKER_THREADS', 2).to_i,
+          prefetch: ENV.fetch('DOMAIN_VALIDATION_WORKER_PREFETCH', 5).to_i
```
delano, line 52: **Insight:** These defaults are intentionally conservative for initial rollout — DNS-bound workers spend most of their time in I/O wait, so higher thread counts (8–16) and prefetch (10–20) would be safe here. Both are already configurable via `DOMAIN_VALIDATION_WORKER_THREADS` and `DOMAIN_VALIDATION_WORKER_PREFETCH` env vars, so this can be tuned per-environment without a code change. Worth bumping once we have production telemetry on actual DNS response distributions.

PR #2907, opened 2026-04-07
Use Familia staged relationships for invitation lifecycle

### LINKED ISSUE #2903: Invited members don't see organization in their org list after accepting
## Problem Statement

When a user is invited to an organization, creates an account via the invite link, and logs in, the inviting organization does not appear in their "Manage Organizations" list. They only see their auto-provisioned default workspace. Meanwhile, the organization owner correctly sees the invited user on the Team page with the correct role and join date.

## Early Diagnosis

The `OrganizationMembership#accept!` method (`organization_membership.rb:238-248`) bypasses `add_members_instance` and writes directly to `org.members.add(customer, score)`. The comment at line 239 explains why: calling `add_members_instance` would try to create a new through model via `ThroughModelOperations.find_or_create`, but one already exists (the invitation record itself).

The problem is that `add_members_instance` normally does two things in a MULTI transaction:
1. `ZADD organization:{org_objid}:members` — `accept!` does this manually
2. `SADD customer:{customer_objid}:participations` — `accept!` **does not do this**

The organization list page calls `cust.organization_instances` which reads from the `customer.participations` reverse index. Since `accept!` never populates it, the org is invisible to the invited member.

The SSO auto-join (`join_domain_organization.rb`) and CLI add-member paths do not have this bug because they use `add_members_instance` which writes both sides.

Likely fix: add `customer.track_participation_in(org.members.rediskey)` inside `accept!` after `org.members.add(customer, score)`.

## Affected Areas

- [ ] `lib/onetime/models/organization_membership.rb` — `accept!` method missing `customer.participations` write
- [ ] `apps/web/auth/operations/accept_invitation.rb` — orchestrates the acceptance flow
- [ ] `apps/api/organizations/logic/organizations/list_organizations.rb` — reads from `organization_instances` (participations-backed)

delano:
## Summary

Fixes #2903 — invited members didn't see the organization in their org list after accepting because the invitation `accept!` path bypassed Familia's `add_members_instance` and wrote directly to `org.members` without populating `customer.participations` (the reverse index that powers `organization_instances`).

The fix migrates the invitation lifecycle to Familia 2.4.0's new staged relationships API, which manages the three-structure invariant atomically:
- Active sorted set (`org.members`)
- Reverse index (`customer.participations`)
- Staging set (`org.pending_invitations`)

A new `ensure_membership` class method serves as the convergence point for all "add a known customer to an org" paths (SSO auto-join, CLI add-member, future join requests). It checks for pending invitations first — activating them if found, otherwise creating directly — so pending invitations are no longer orphaned when users arrive via SSO or CLI.

## Changes

- `organization_membership.rb`: `accept!` uses `activate_members_instance`, `create_invitation!` uses `stage_members_instance`, `revoke!` uses `unstage_members_instance`. New `ensure_membership` class method composes activate + add with domain-specific email lookup.
- `join_domain_organization.rb`: SSO auto-join now calls `ensure_membership` instead of `add_members_instance`
- `add_member_command.rb`: CLI add-member now calls `ensure_membership`, removed `warn_pending_invitation`/`check_pending_invitation` (dead code)
- `organization.rb`: Removed manual `sorted_set :pending_invitations` (now auto-generated by Familia)
- `customer.rb`: Added `staged: :pending_invitations` to `participates_in` declaration
- Familia gem updated from git branch reference to released `~> 2.4.0`

## Test plan

- [x] 22 tests for `ensure_membership` (direct add, idempotency, pending activation, expired/declined fallthrough, role precedence, reverse index)
- [x] 25 tests for staged relationship lifecycle (stage, activate, unstage, reverse index regression)
- [x] 17 tests for index lifecycle (token_lookup, org_email_lookup, org_customer_lookup transitions)
- [x] 93 integration tests for invitation and auth flows pass
- [x] 1306 model tests pass (pre-existing billing test failures unrelated)


delano:
NOTE: Feedback resolved on #2909.

PR #3002, opened 2026-04-15
Derive Sentry breadcrumb scrub patterns from API route metadata

### LINKED ISSUE #2981: Derive Sentry breadcrumb scrub patterns from API route metadata
## Problem Statement

The `beforeBreadcrumb` handler in `src/plugins/core/enableDiagnostics.ts` scrubs sensitive path segments from HTTP breadcrumbs (XHR/fetch) using a hardcoded regex: `/(secret|private|receipt|incoming)/[a-zA-Z0-9]+`. This pattern is front-line defense for API breadcrumbs and must be manually kept in sync with actual sensitive API routes. As new endpoints are added, the regex can silently fall out of date, leaving sensitive identifiers exposed in Sentry breadcrumbs.

## Early Diagnosis

The project already has the infrastructure to do this properly. Vue Router routes declare scrubbing intent via `sentryScrubParams` in route metadata — the `beforeBreadcrumb` handler uses this for navigation breadcrumbs. The same principle could apply to API endpoints: Otto routes already support arbitrary `name=value` metadata, which flows through `scripts/openapi/otto-routes-parser.ts` into the generated OpenAPI specs. A `sensitive=true` annotation on sensitive routes would let the frontend derive scrub patterns at build time instead of maintaining a separate regex.

## Affected Areas

- [ ] `src/plugins/core/enableDiagnostics.ts` — replace hardcoded `SENSITIVE_PATH_PATTERN` with generated patterns
- [ ] `scripts/openapi/otto-routes-parser.ts` — ensure `sensitive` (or equivalent) name=value is passed through
- [ ] `generated/openapi/` — `x-sensitive` extension on sensitive path operations
- [ ] `src/schemas/api/index.ts` — may need schema extension for the new field
- [ ] Otto routes `.txt` files — add `sensitive=true` or `sensitive=identifier,key` to routes with sensitive path params

## Approach

1. Add `sensitive=true` (or `sensitive=<param-list>`) to sensitive Otto route definitions
2. Thread it through the parser and into OpenAPI output as an `x-sensitive` extension
3. Add a build step that generates `src/generated/sentry-scrub-patterns.ts` — a typed set of path templates derived from flagged routes
4. Replace `SENSITIVE_PATH_PATTERN` in `enableDiagnostics.ts` with patterns imported from the generated file

## Metadata Format

The `sensitive` param supports two forms (mirroring `sentryScrubParams` in Vue Router):

- `sensitive=true` — scrub all path parameters in this route
- `sensitive=identifier,key` — scrub only the named parameters

This allows precise control while remaining general enough for other systems (logging, analytics) to consume the same metadata.

## Related

Follows from #2962 — the hardcoded regex introduced there is the immediate gap this resolves.

delano:
Generates Sentry breadcrumb URL scrub patterns automatically from API route metadata rather than maintaining them by hand. A new script (`scripts/generate-sentry-scrub-patterns.ts`) parses route definitions across all API versions and emits typed patterns into `src/generated/sentry-scrub-patterns.ts`, which the diagnostics scrubber then consumes. Includes a test suite covering the generated patterns.

Also regenerates the route listing files for v1/v2/v3 as a side effect of the updated OpenAPI generation script.

Closes #2981

### DIFF
--- a/apps/api/v1/routes.txt
+++ b/apps/api/v1/routes.txt
@@ -20,23 +20,23 @@ POST   /generate                             V1::Controllers::Index#generate    
 
 POST   /create                               V1::Controllers::Index#create      openapi_auth=basic,anonymous content=form response=default
 
-POST   /secret/:key                          V1::Controllers::Index#show_secret openapi_auth=basic,anonymous content=form response=default
+POST   /secret/:key                          V1::Controllers::Index#show_secret openapi_auth=basic,anonymous content=form response=default sensitive=true
 
 # Canonical receipt paths
 # NOTE: Specific sub-paths (burn) must precede the :key wildcard so
 # Otto's first-match-wins dynamic routing resolves them correctly.
 GET    /receipt/recent                       V1::Controllers::Index#show_receipt_recent openapi_auth=basic response=default
-POST   /receipt/:key/burn                    V1::Controllers::Index#burn_secret  openapi_auth=basic,anonymous content=form response=default
-GET    /receipt/:key                         V1::Controllers::Index#show_receipt openapi_auth=basic,anonymous response=default
-POST   /receipt/:key                         V1::Controllers::Index#show_receipt openapi_auth=basic,anonymous content=form response=default
+POST   /receipt/:key/burn                    V1::Controllers::Index#burn_secret  openapi_auth=basic,anonymous content=form response=default sensitive=true
+GET    /receipt/:key                         V1::Controllers::Index#show_receipt openapi_auth=basic,anonymous response=default sensitive=true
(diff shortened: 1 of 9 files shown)


delano:
Addressed review feedback in 1d943419c:

**CodeQL finding (critical)**
- Added backslash escaping in `pathToRegexPattern()` before other regex escapes

**Reviewer feedback**
- Fixed replace callback to properly slice off `offset` and `originalString` from args (was accidentally iterating over them due to `...groups` rest parameter)

**Test coverage**
- Added 6 test cases covering backslash escaping edge cases

The global regex `lastIndex` concern is mitigated by existing design:
- Patterns are `@internal` (exported only for testing)  
- Functions use `.replace()` which resets `lastIndex` automatically
- Generated patterns already have explicit `pattern.lastIndex = 0` resets

PR #3116, opened 2026-05-13
Fix #3107: resolve client IP behind RFC-1918 proxy chains

### LINKED ISSUE #3107: Client IP shows as reverse proxy address behind Kubernetes ingress
When OTS runs behind Kubernetes ingress (Traefik in our case), the app displays the ingress pod IP instead of the actual client IP. Traefik sets `X-Forwarded-For` correctly, but it never reaches the display.

Affected: Colonel "Banned IPs" page — "Your Current IP Address" shows the pod IP (`10.244.8.0` in our case) instead of the real visitor IP.

## Root cause

Rack's `request.ip` scans `X-Forwarded-For` for the leftmost IP that isn't a known proxy. In Kubernetes, both the real client and the ingress pod can be in RFC 1918 space. Rack can't find a non-private IP in the chain and falls back to `REMOTE_ADDR` — the pod IP.

Otto handles this correctly via `add_trusted_proxy` / `client_ipaddress`, but the code path for the ban IP display calls `request.ip` directly.

## Suggested fix

A global `TRUSTED_PROXY_DEPTH` or `TRUSTED_PROXIES` env var. `UI_HOMEPAGE_TRUSTED_PROXY_DEPTH` already does this for homepage mode detection — it just needs to be generalized to the rest of the app.

## Setup

- OTS v0.25.2
- Kubernetes, Traefik ingress
- All IPs in the forwarding chain are RFC 1918

syphernl:
When OTS runs behind Kubernetes ingress, the Colonel "Banned IPs" page shows the ingress pod IP instead of the visitor's actual IP — and bans enforced against that IP never match incoming requests.

Rack's `request.ip` looks through `X-Forwarded-For` for the leftmost non-private IP. In Kubernetes, everything in the chain can be RFC 1918, so it gives up and falls back to `REMOTE_ADDR` — also the pod IP. The homepage mode detection already solved this with `UI_HOMEPAGE_TRUSTED_PROXY_DEPTH`, but that setting was scoped to CIDR matching and never reached the ban or session paths.

There's a second bug on top of that: `list_banned_ips.rb` was reading `metadata[:ip]` from the strategy result, but `IdentityResolution` writes `metadata[:ip_address]`. The Colonel page fell through to `'unknown'` on every request regardless of proxy configuration.

## What this PR does

Adds `TRUSTED_PROXY_DEPTH` (default `0`, so existing deployments are unaffected) and `TRUSTED_IP_HEADER` to the site config. Set depth to `1` behind a standard reverse proxy, `2+` behind CDN + proxy chains.

Introduces `Onetime::ClientIpHelpers` with a `RackExtension` mixin that adds `trusted_client_ip` to `Rack::Request` at boot — no per-callsite setup needed. Crucially, this helper does not reject private IPs after extraction: once you've declared how many proxies to trust, the extracted IP is the client IP, full stop. That's what makes the all-RFC-1918 Kubernetes case work.

All `request.ip` / `req.ip` calls are replaced with `request.trusted_client_ip` across identity resolution, ban enforcement, session storage, auth hooks, audit logging, and error handling. Otto's `req.client_ipaddress` is untouched.

The metadata key bug in `list_banned_ips.rb` is fixed to read `metadata[:ip_address]`.

Homepage mode detection now delegates its header parsing to `ClientIpHelpers` but keeps the private-IP rejection it uses for CIDR matching.

## Testing

New spec at `spec/unit/onetime/helpers/client_ip_helpers_spec.rb` covers depth 0/1/2, all three header modes, all-RFC-1918 chains, missing and malformed headers, and the `Rack::Request` mixin. Existing homepage specs are unchanged.

To verify manually: start the server with `TRUSTED_PROXY_DEPTH=1`, curl any endpoint with `X-Forwarded-For: 203.0.113.42, 10.244.8.0`, and confirm the Colonel page reports `203.0.113.42`. With depth unset, it should still return `REMOTE_ADDR`.

Closes #3107


delano:
Not sure if you want a review yet so I just glanced over it, but it looks like a good changes. And thanks for catching the ip_address field bug as well.

There are some minor stylistic differences but I'll end up offloading a lot of this stuff to Otto at some point so I'm not going to nitpick about it. If you can get the one ruby test to pass and mark it ready, I'll merge it in.

syphernl:
Ran an end-to-end test with a local Docker build. Banned `5.5.5.5` in Redis, then fired requests with different depth settings:

- `TRUSTED_PROXY_DEPTH=1` + `X-Forwarded-For: 5.5.5.5` → 403 ✓
- `TRUSTED_PROXY_DEPTH=0` + `X-Forwarded-For: 5.5.5.5` → 200 (XFF ignored, REMOTE_ADDR used) ✓
- `TRUSTED_PROXY_DEPTH=1`, no XFF header → 200 (falls back to REMOTE_ADDR) ✓

Spent a bit of time confused about why the `[Otto] Auth strategy result` log kept showing the Docker bridge IP even with depth=1. Turns out that's just Otto's inner `IPPrivacyMiddleware` masking IPs before the log runs. The ban, session, and identity middleware sit earlier in the stack and get the real resolved IP, which is what matters.

syphernl:
@delano This is ready for final review & merge.

delano:
> Spent a bit of time confused about why the [Otto] Auth strategy result log kept showing the Docker bridge IP even with depth=1. Turns out that's just Otto's inner IPPrivacyMiddleware masking IPs before the log runs. The ban, session, and identity middleware sit earlier in the stack and get the real resolved IP, which is what matters.

Oof, sorry for the headache. Otto still has some rough edges to sort out.

delano:
@syphernl While sorting out the tests, I figured out a way to implement the trusted proxies using Rack internals that allows for using the regular `request.ip`. The limitation with the approach is that it filters to RFC-1918 ranges by default (plus any additional CIDRs configured) which removed the explicit depth control (your changes) so it supports both via `mode: filter | depth`. 

I took the opportunity to plan ahead with the configuration more generally and moved it under a new layer `site.network`. 

```yaml
site:
  # ...
  #
  # Settings for proxy trust, client IP resolution, and network-level behavior.
  network:
    # Trusted Proxy Configuration
    #
    # Controls how client IPs are resolved when behind reverse proxies
    # (Kubernetes ingress, cloud load balancers, CDNs, etc.).
    #
    trusted_proxy:
      # Master switch. When false, all forwarded headers are ignored
      # and REMOTE_ADDR is used directly.
      enabled: <%= ENV['TRUSTED_PROXY_ENABLED'] == 'true' %>

      # How to resolve client IP behind proxies:
      #
      # 'filter' (default): Rack walks X-Forwarded-For right-to-left,
      #   automatically skipping RFC1918 private IPs (10.x, 172.16-31.x,
      #   192.168.x). Returns first non-private IP. Works for most k8s/cloud
      #   deployments where proxies have internal IPs.
      #
      # 'depth': Position-based counting. Skip exactly N rightmost hops.
      #   Use when:
      #   - CDN has a public IP (not RFC1918)
      #   - Variable proxy depth that filter can't handle
      #   - You need deterministic hop selection regardless of IP class
      #
      mode: <%= ENV['TRUSTED_PROXY_MODE'] || 'filter' %>
      header: <%= ENV['TRUSTED_PROXY_HEADER'] || 'X-Forwarded-For' %>
      cidrs: <%= ENV['TRUSTED_PROXY_CIDRS']&.split(',')&.map(&:strip) || [] %>
      depth: <%= ENV['TRUSTED_PROXY_DEPTH']&.to_i || 1 %>
```
  
I'm going to merge now so we can keep building on it. As a bonus, I think it sets us up to simplify the homepage mode settings too.

This is a great contribution. If you want to keep iterating, or there's something you want to revisit, a follow-up issue or PR is welcome.

PR #3167, opened 2026-05-21
feat: add EMAILER_MODE=disabled to silence mail delivery

### LINKED ISSUE #3109: Add option to disable email delivery entirely
For self-hosted deployments that handle all communication through SSO or don't need email at all, there's no way to disable the emailer. `EMAILER_MODE` only accepts delivery modes (`smtp`, etc.) — there's no `none` or `disabled` value.

This means a misconfigured or intentionally SMTP-less deployment will throw errors on any code path that tries to send mail, rather than silently skipping it.

## Suggested fix

Add `EMAILER_MODE=disabled` (or `EMAILER_ENABLED=false`) that short-circuits mail delivery without errors. Useful for:

- SSO-only instances where email verification is turned off (`AUTH_AUTOVERIFY=true`)
- Air-gapped or internal deployments with no outbound SMTP
- Testing and staging environments

## Setup

- OTS v0.25.2, self-hosted, SSO-only (`AUTH_SSO_ONLY=true`, `AUTH_AUTOVERIFY=true`)

syphernl:
## Summary

- Adds `Delivery::Disabled` backend that silently discards all outbound mail
- Handles `EMAILER_MODE=disabled` and `EMAILER_MODE=none` in the delivery backend selector
- Logs a single info line per skipped delivery attempt for observability

## Use cases

- SSO-only deployments with `AUTH_AUTOVERIFY=true` (no email verification needed)
- Air-gapped or internal instances with no outbound SMTP
- Staging/testing environments where email delivery should be suppressed

## Configuration

```
EMAILER_MODE=disabled
```

Closes #3109

### DIFF
--- a/lib/onetime/mail/delivery/disabled.rb
+++ b/lib/onetime/mail/delivery/disabled.rb
@@ -18,8 +18,7 @@ module Onetime
       # All delivery attempts succeed silently with no side effects.
       #
       class Disabled < Base
-        def perform_delivery(email)
-          OT.info "[mail] Delivery skipped (emailer disabled) to=#{obscure_email(email[:to])}"
+        def perform_delivery(_email)
           nil
         end
 
--- a/lib/onetime/mail/delivery/disabled.rb
+++ b/lib/onetime/mail/delivery/disabled.rb
@@ -1,0 +1,38 @@
+# lib/onetime/mail/delivery/disabled.rb
+#
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module Onetime
+  module Mail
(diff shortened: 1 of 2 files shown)


delano:
Yeah this makes sense. 

There's the `UI_CAPABILITIES_RECIPIENT` to toggle display of the recipient field as well which is a purely display setting. In some universes it would make sense to not display the recipient field automatically when `EMAILER_MODE=disabled` or `EMAILER_MODE=none` but the way the configuration is structured I don't think it makes sense to do that via the backend.

PR #3168, opened 2026-05-21
feat: hide API key section in Account Settings when API_ENABLED=false

### LINKED ISSUE #3115: Hide API key section in Account Settings when API_ENABLED=false
When `API_ENABLED=false`, the API routes correctly return 404. But the "API Key" section in Account Settings is still visible.

Users see a key, copy it, try to use it, and get a 404 with no explanation. There's nothing in the UI that tells them the API is disabled.

Would be good if the section was hidden when `API_ENABLED=false`. Alternatively, a short message saying "the API is disabled for this instance" would also work.

## Steps to reproduce

1. Set `API_ENABLED=false`
2. Log in as any user
3. Open Account Settings — the API key section is still there

syphernl:
## Summary

- Exposes `api_enabled` from the Ruby config serializer to the frontend bootstrap payload
- Adds `api_enabled` to the Zod bootstrap schema (default: `true`)
- In Account Settings → API tab: shows a notice instead of the key form when the API is disabled
- Adds English i18n key `web.settings.api.api_disabled_notice`

## Before / After

**Before:** Users with `API_ENABLED=false` still see their API key, copy it, try to use it, and get a 404 with no explanation.

**After:** The API key section is replaced with a clear notice that the API is disabled for this instance.

Closes #3115

### DIFF
--- a/apps/web/core/spec/views/serializers/config_serializer_spec.rb
+++ b/apps/web/core/spec/views/serializers/config_serializer_spec.rb
@@ -70,7 +70,21 @@ RSpec.describe Core::Views::ConfigSerializer do
       'site' => {
         'host' => canonical_domain,
         'ssl' => true,
-        'interface' => { 'ui' => {} },
+        'interface' => {
+          'ui' => {},
+          'api' => {
+            'enabled' => true,
+            'guest_routes' => {
+              'enabled' => true,
+              'conceal' => true,
+              'generate' => true,
+              'reveal' => true,
+              'burn' => true,
+              'show' => true,
+              'receipt' => true,
+            },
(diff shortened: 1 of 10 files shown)


delano:
Thanks for all these nice additions to configuration btw. 

I just took the opportunity with this one to pass the entire `site.interface.api` object. When I added the API interface settings, I was hesitant to pass them at all. I wasn't sure yet whether there would be settings that we would not want to pass to the frontend. That was I think a year ago now and we actually had the opposite problem: we weren't passing the guest_routes through to the frontend so they were just hardcoded in the components/composables. That is fixed now as well.

PR #3169, opened 2026-05-21
feat: add WORKSPACE_FEEDBACK_ENABLED and FOOTER_VERSION_ENABLED env vars

### LINKED ISSUE #3108: Add env vars to hide feedback link and footer version independently
Two small configuration gaps for self-hosted deployments:

## Feedback link

The feedback link in workspace navigation can only be hidden by disabling all workspace links (`WORKSPACE_LINKS=false`). There's no way to hide just the feedback item while keeping other workspace links visible. A `WORKSPACE_FEEDBACK_ENABLED` (or similar) env var would cover this.

## Footer version string

The footer shows the OTS version number. For self-hosted deployments this leaks version info to end users. A `FOOTER_VERSION_ENABLED` env var to suppress it would be useful.

## Setup

- OTS v0.25.2, self-hosted

syphernl:
## Summary

- **`WORKSPACE_FEEDBACK_ENABLED`** (default: `true`): set to `false` to remove the Feedback link from the workspace footer while keeping all other workspace links intact. Previously the only way to hide feedback was to disable all workspace links via `WORKSPACE_LINKS`.

- **`FOOTER_VERSION_ENABLED`** (default: `true`): set to `false` to suppress the OTS version number in the workspace footer. Useful for self-hosted deployments where leaking version info to end users is undesirable.

## Implementation

- ERB conditional in `config.defaults.yaml` omits the feedback link entry when disabled
- `show_version` field added to `uiInterfaceSchema` (Zod) and the `ui` bootstrap payload
- `WorkspaceFooter.vue` reads `ui.show_version` from the bootstrap store to gate version display

Closes #3108


delano:
Good call. I just made sure the vars are used everywhere and added test coverage. I also updated `.env.reference`.

PR #3268, opened 2026-05-28
fix: cast numeric safe_dump fields to integers for V3 schema compatibility

syphernl:
## What broke and why

Recipients saw "That information is no longer available" immediately on opening a secret link, even for secrets the sender confirmed had never been viewed. The sender's dashboard showed the secret as "Previewed" but not "Viewed" -- the backend was processing the request, but the frontend was failing silently after getting the response back.

What happens:

1. Recipient opens the link
2. Frontend calls `GET /api/v3/guest/secret/:identifier`
3. Backend runs `ShowSecret`, calls `secret.previewed!`, returns HTTP 200
4. Frontend runs `gracefulParse` against `responseSchemas.secret`
5. Validation fails: `lifespan`, `secret_ttl`, `created`, and `updated` come back as strings, but the V3 schema uses `z.number()`, which does not coerce
6. `gracefulParse` throws, `record` stays null, the `UnknownSecret` component renders

The secret was never consumed, just unreachable.

## Root cause

Familia stores plain `field` values as strings in Redis. The V3 schema was written assuming native JSON types, but the `safe_dump` lambdas for numeric fields were returning raw Redis strings. `z.number()` is strict -- `"604800"` fails where `604800` passes.

## The fix

Added `.to_i` casting in the `safe_dump` lambdas for numeric fields in both `Secret` and `Receipt`:

- `lifespan`, `secret_ttl`, `metadata_ttl`, `receipt_ttl`: use `&.to_i` (nil-safe, keeps null for unset lifespans)
- `created`, `updated`: use `.to_i` (always set by Familia)

The cast is a no-op when the value is already an integer, so it is safe regardless of which Familia version is in use.

## How to reproduce

1. Deploy with a Familia version that stores plain fields as strings
2. Create a secret
3. Open the link in an incognito window
4. You see "That information is no longer available" with no confirmation form shown
5. Check the sender's dashboard: secret shows "Previewed", not "Viewed" -- it was never consumed

## Testing note

Not verified locally -- setting up a full env is still needed. The logic is straightforward (`.to_i` on a string integer is well-defined Ruby), but maintainer review and a test run before merging would be good.

The existing test `"rejects missing lifespan field from API (V3 requires number)"` in `secretStoreFieldHandling.spec.ts` already covers the null case -- a companion test for string-typed lifespan would close the gap.


delano:
@syphernl Oof, I thought I was past these type issues. There used to be a tonne of code like that. 

It shouldn't be happening now for primitives since we read and write to redis/valkey with a JSON serializer so they should be hydrating as numbers. That could mean A) something is writing as a string or B) something is converting it to a string after it's read, before safe_dump is called. 

I need to finish some stuff for Teams and SSO so I might not get a chance to investigate until Monday-ish.

delano:
@syphernl I haven't been able to reproduce this with Familia v2. I know this behaviour is totally possible, I just don't want to commit code that I can't reliably says fixes the problem. I'm going to close this PR, but please open an issue if this is an ongoing thing and I'll look into it further.

PR #3410, opened 2026-06-09
Complete Tailwind v4 CSS-first migration on develop (remove tailwind.config.ts)

delano:
Completes the Tailwind v4 CSS-first migration on `develop` and removes `tailwind.config.ts`, collapsing the theme into a single source of truth (`src/assets/style.css`).

Sibling of #3408, which does the same on `main`. This PR targets `develop`, where the brand palettes already live in `style.css` under `@theme static`, so only `content` / `darkMode` / `fontFamily` remain to port.

## Changes
- `darkMode: 'class'` → `@custom-variant dark (&:is(.dark *))`
- `content[]` → `@import "tailwindcss" source(none)` + `@source` globs
- `fontFamily.brand` → `@theme --font-brand`
- Remove the dead `--background-size-200%` var (generates no utility in v4; `bg-200%` was already a no-op)
- Repoint tooling at the CSS entry and drop references to the deleted file: eslint-plugin-tailwindcss `config` (absolute path), prettier `tailwindStylesheet`, Dockerfile `COPY`, `.dockerignore`, CI path filters, labeler, docs
- Delete `tailwind.config.ts`

## Branding safety
The brand color block stays **`@theme static`** (not plain `@theme`). The `static` modifier emits all 44 `--color-brand*` defaults into `:root`, so `useBrandTheme`'s per-custom-domain overrides on `<html>` — and the `clearOverrides()` fallback to the compiled defaults — both keep working. Brand color values are unchanged.

## Verification
Built `develop` before and after the migration on a worktree:
- Production CSS emits an **identical set of 2,201 utility classes** (zero lost, zero added).
- All **44 brand CSS variables** are still emitted to `:root`; default `--color-brand-500:#3b82f6` present and `bg-brand-500 → var(--color-brand-500)`, so the runtime branding bridge is provably intact.

https://claude.ai/code/session_012uqyngfgZ9JG8ETaG3YaEx

---
_Generated by [Claude Code](https://claude.ai/code/session_012uqyngfgZ9JG8ETaG3YaEx)_


delano:
**CI note:** the `T2 · TypeScript Unit Tests` failure is pre-existing on `develop`, not introduced by this PR.

Verified by running the failing specs on pristine `origin/develop` (30dee3a) without this PR's commit: the **identical 10 failures** reproduce — 7 in `DomainSsoConfigForm.spec.ts` ("Enforce SSO Only Toggle": `#domain-sso-enforce-only` never renders) and 3 in `ProfileSettings.spec.ts` (entitlements fetch mock not called). Same 10-failed / 33-passed result on both this branch and base.

Also closing the verification gap noted in the PR description: the branding suites (`useBrandTheme.spec.ts`, `brand-palette.spec.ts`, `useBranding.spec.ts`) now ran against this branch locally — **70 passed, 0 failed**, confirming the runtime palette override and `@theme static` fallback are intact.

The 10 inherited failures look like component/test drift on `develop` (SSO toggle gating, entitlements fetch) and are worth a separate fix.

https://claude.ai/code/session_012uqyngfgZ9JG8ETaG3YaEx

---

PR #3429, opened 2026-06-11
Fix: give the IP privacy middleware a trusted-proxy config

### LINKED ISSUE #3427: IP privacy middleware hides the real client IP when running behind a proxy
Behind a reverse proxy, the Colonel "Your Current IP Address" panel on `/colonel/banned-ips` shows the ingress hop with its last octet zeroed (for us `10.244.10.0`, a Traefik pod) instead of the visitor's IP. The same wrong value feeds ban checks, sessions, identity resolution, and audit logs, so bans never match real traffic. Enabling `site.network.trusted_proxy` does not fix it.

## Cause

`MiddlewareStack.configure` mounts the IP privacy middleware first with no security config:

```ruby
builder.use Otto::Security::Middleware::IPPrivacyMiddleware
```

With `@security_config` nil, Otto's `resolve_client_ip` returns on its first line:

```ruby
return remote_addr unless @security_config
```

So it takes `REMOTE_ADDR` (the proxy hop), never reads `X-Forwarded-For`, masks it, then overwrites `REMOTE_ADDR` / `X-Forwarded-For` / `X-Real-Ip` with the masked value. Everything after it reads `request.ip` from an env whose forwarded headers are already gone. The `.0` ending is the one-octet privacy mask.

The `site.network.trusted_proxy` settings from #3116 only configure `Rack::Request#ip`, which runs after the middleware has already rewritten the headers. The two paths use separate trust lists, and the Otto one (`Otto::Security::Config#trusted_proxies`) is never populated, since every `Otto.new` is called without `trusted_proxies:`.

## Fix

Pass the middleware an `Otto::Security::Config` that trusts the proxy hops (RFC1918/loopback), gated on `site.network.trusted_proxy.enabled`, so it resolves the real client from the forwarded headers before masking. PR incoming.

Seen on `v0.25.8` (otto `2.0.2`); same code path on `main` (otto `2.2.0`).

syphernl:
Fixes #3427.

The IP privacy middleware is mounted first in the common stack with no security config:

```ruby
builder.use Otto::Security::Middleware::IPPrivacyMiddleware
```

With `@security_config` nil, Otto's `resolve_client_ip` returns `REMOTE_ADDR` without reading `X-Forwarded-For`, masks it, and overwrites the forwarded headers with the masked value. Behind a proxy the real client IP is gone before any later middleware or auth strategy calls `request.ip`, so the `site.network.trusted_proxy` config from #3116 (which only configures `Rack::Request#ip`) runs too late to help. Details in #3427.

## Change

`MiddlewareStack` builds an `Otto::Security::Config` and passes it to the middleware. When `site.network.trusted_proxy.enabled` is true it trusts the private proxy ranges (RFC1918, loopback, link-local, IPv6 ULA/loopback), so the middleware resolves the client from the forwarded headers and masks that instead. When trusted proxy is off, the helper returns nil and behaviour is unchanged, so direct-connection deployments are unaffected.

## Limits

- Only RFC1918/loopback hops are trusted. Public proxy ranges (a CDN with public egress IPs) would need CIDR matching, which Otto's prefix-based trusted-proxy list doesn't do, so `site.network.trusted_proxy.cidrs` still only affects `Rack::Request#ip`. Noted in a code comment.
- The stored/displayed IP is still masked to a /24 by the existing privacy settings. This just makes it the correct /24 rather than the proxy's.

## Tests

Added `spec/unit/onetime/application/middleware_stack_spec.rb` for the absent, disabled, and enabled cases (documentation IPs only). Couldn't run the suite locally (no bundler in my env), so please let CI confirm.

### DIFF
--- a/lib/onetime/application/middleware_stack.rb
+++ b/lib/onetime/application/middleware_stack.rb
@@ -54,6 +54,33 @@ module Onetime
         'application/x-www-form-urlencoded' => proc { |body| Rack::Utils.parse_nested_query(body) },
       }.freeze
 
+      # IP ranges that are always treated as proxy hops, never as the client.
+      # IPv4: RFC1918 (10/8, 172.16/12, 192.168/16), loopback (127/8) and
+      # link-local (169.254/16). IPv6: loopback (::1), ULA (fc00::/7, the
+      # f[cd] branch) and link-local (fe80::/10, the fe[89ab] branch). Otto's
+      # IPPrivacyMiddleware walks X-Forwarded-For and returns the first
+      # address that does NOT match a trusted proxy, so trusting these private
+      # ranges is what lets the real public visitor IP surface in deployments
+      # where every proxy hop has an internal address (Kubernetes ingress,
+      # cloud load balancers).
+      #
+      # Defined at module scope (not inside `class << self`) so it is reachable
+      # as MiddlewareStack::PRIVATE_PROXY_RANGES from the Otto router wiring in
+      # OttoHooks, while remaining lexically visible to the singleton methods
+      # below.
+      PRIVATE_PROXY_RANGES = /
+        \A(?:
+          10\.|
+          127\.|
+          192\.168\.|
+          169\.254\.|
+          172\.(?:1[6-9]|2\d|3[01])\.|
+          ::1\z|
+          f[cd]|
+          fe[89ab]
+        )
+      /ix
+
       class << self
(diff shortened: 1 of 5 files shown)


delano:
@syphernl Good fix.

Just a little organizing and passing the same trusted-proxy list to Otto. Also added fe80::/10 to the private ranges and rounded things out with some IPv6 and router-wiring tests.

PR #3437, opened 2026-06-12
Fix MFA QR code to encode correct HMAC secret (issue #3431)

### LINKED ISSUE #3431: MFA QR code encodes wrong secret; manual setup key works
## Problem Statement

Customers cannot enable MFA by scanning the QR code: after scanning with their authenticator app (e.g. Google Authenticator), entering their password and the current OTP, they always get `"Authentication failed. Please verify your credentials and try again."`. Typing the **setup key manually** into the same app works fine. This means the QR code encodes a different secret than the manual setup key, so the authenticator computes OTPs against the wrong secret.

## Early Diagnosis

The QR code is generated client-side from the OTP secret, but the secret it encodes is **inconsistent** with the manually-displayed key, and the two QR-generation callsites disagree with each other:

- `OtpSetupWizard.vue:150` — manual entry displays `setupData.otp_setup`.
- `useMfa.ts:180` (standard/200 path) — QR encodes `otp_setup` (matches manual).
- `mfaHelpers.ts:100` (HMAC/422 path) — QR encodes `otp_raw_secret` (differs from manual).

Suspected root cause: in the HMAC setup path the QR encodes `otp_raw_secret` while the manual key shown to the user — and the value the server validates against — is `otp_setup`. Authenticator scans → wrong secret → wrong OTP → "Authentication failed". Manual typing of `otp_setup` works because it is the secret the server expects. The fix is to make the QR encode the *same* secret the manual key displays and the server validates, and to make both callsites consistent.

Secondary concern: `generateQrCode` (`mfaHelpers.ts:35-42`) builds the `otpauth://` URI by hand without `algorithm`/`digits`/`period` params and without base32 normalization (stripping spaces/uppercasing). Likely not the primary cause here, but worth hardening once the field mismatch is fixed.

```
otpauth://totp/${issuer}:${email}?secret=${secret}&issuer=${issuer}
```

## Affected Areas

- [ ] `src/shared/composables/helpers/mfaHelpers.ts` — `generateQrCode` (L35-42) + HMAC-path QR uses `otp_raw_secret` (L100)
- [ ] `src/shared/composables/useMfa.ts` — standard-path QR uses `otp_setup` (L180); reconcile which field is the TOTP secret
- [ ] `src/apps/session/components/OtpSetupWizard.vue` — manual key displays `otp_setup` (L150); must match the QR
- [ ] `apps/web/auth/routes/mfa.rb` / `apps/web/auth/config/features/mfa.rb` — confirm backend semantics of `otp_setup` vs `otp_raw_secret` (which is the raw base32 TOTP key vs HMAC'd validation value)

## Repro

1. Start MFA setup, scan the QR with an authenticator app.
2. Enter password + current OTP → "Authentication failed."
3. Instead, type the displayed setup key manually into the app → succeeds.

delano:
## Summary

Fixes #3431

The MFA QR code was encoding the wrong secret when HMAC mode was enabled. The backend generates two secrets during setup:
- `otp_raw_secret`: Raw key used only for the setup handshake
- `otp_setup`: HMAC'd key that the authenticator app must actually use

The frontend was incorrectly reconstructing the otpauth:// URI and encoding `otp_raw_secret` instead of `otp_setup`, causing scanned QR codes to never match what the server validates.

**Fix:** The backend now emits Rodauth's authoritative `otp_provisioning_uri` (built from the correct HMAC'd secret) in the otp-setup response as `provisioning_uri`. The frontend renders this URI directly as a QR code without reconstruction, ensuring the encoded secret always matches server validation.

### Backend (Ruby/Rodauth)
- **apps/web/auth/config/hooks/mfa.rb**: Added `before_otp_setup_route` hook to emit `provisioning_uri` from Rodauth's `otp_provisioning_uri` when HMAC is enabled
- **apps/web/auth/spec/config/features/mfa_provisioning_uri_spec.rb**: Added comprehensive regression tests covering:
  1. `provisioning_uri` embeds the correct HMAC'd secret, not the raw secret
  2. TOTP codes derived from `provisioning_uri` complete setup successfully
  3. TOTP codes derived from `otp_raw_secret` are rejected (the original bug)

### Frontend (TypeScript/Vue)
- **src/shared/composables/helpers/mfaHelpers.ts**: 
  - `generateQrCode()` now accepts the backend's `provisioning_uri` directly instead of reconstructing it from secret/issuer/email
  - `enrichSetupResponse()` renders QR from `provisioning_uri` instead of `otp_raw_secret`
  - Removed `siteName` and `email` parameters (no longer needed)
  
- **src/shared/composables/useMfa.ts**: 
  - `setupMfa()` no longer requires `siteName` and `email` parameters
  - Uses backend's `provisioning_uri` for QR rendering in both standard and HMAC flows
  
- **src/apps/session/components/OtpSetupWizard.vue**: Simplified to call `setupMfa()` without site name/email parameters

- **src/schemas/api/auth/responses/auth.ts**: Added `provisioning_uri` field to `otpSetupResponseSchema`

- **src/tests/composables/useMfa.spec.ts**: Updated tests to verify QR is rendered from `provisioning_uri` and removed site name/email parameter assertions

## Test Plan

Added comprehensive regression test suite (`mfa_provisioning_uri_spec.rb`) that exercises the real otp-setup route with the production hook configuration. Tests verify:
- The provisioning URI embeds the HMAC'd secret (not raw secret)
- Setup completes successfully with codes derived from the provisioning URI
- Setup fails with codes derived from the raw secret (confirming the bug is fixed)

Updated existing frontend unit tests to verify QR code generation uses the backend's `provisioning_uri` directly.

### DIFF
--- a/apps/web/auth/config/hooks/mfa.rb
+++ b/apps/web/auth/config/hooks/mfa.rb
@@ -32,6 +32,35 @@ module Auth::Config::Hooks
       # - Verification attempts (POST with OTP code)
       #
       auth.before_otp_setup_route do
+        # ----------------------------------------------------------------
+        # Emit Rodauth's authoritative TOTP provisioning URI (issue #3431)
+        # ----------------------------------------------------------------
+        #
+        # By the time this hook body runs, Rodauth's json feature has already
+        # executed (via super) its own before_otp_setup_route, which — in the
+        # HMAC JSON flow — generates the setup secret and populates the
+        # response with:
+        #   - otp_setup_param     => otp_user_key (the HMAC'd key the
+        #                            authenticator must actually use)
+        #   - otp_setup_raw_param => the raw key, used ONLY for the setup
+        #                            handshake (NOT a valid TOTP secret)
+        #
+        # The SPA historically reconstructed the otpauth:// URI itself and, on
+        # the HMAC path, encoded the raw key — so scanned codes never matched
+        # what the server validates (otp built from otp_user_key), while the
+        # manually-displayed otp_setup key worked. We instead emit Rodauth's
+        # own otp_provisioning_uri (built from otp_user_key, with the correct
+        # issuer/digits/period) as the single source of truth, so the frontend
+        # can render the QR directly without guessing the secret or params.
+        #
+        # Rebuild the tmp OTP key from the raw secret we just handed the client
+        # so otp_provisioning_uri reflects exactly that secret, independent of
+        # any later key regeneration in the route body.
+        if otp_keys_use_hmac? && (raw_secret = json_response[otp_setup_raw_param])
+          otp_tmp_key(raw_secret)
+          json_response[:provisioning_uri] = otp_provisioning_uri
+        end
+
         is_post      = request.post?
         has_otp_code = !param_or_nil(otp_auth_param).to_s.empty?
         has_password = !param_or_nil(password_param).to_s.empty?
--- a/apps/web/auth/spec/config/features/mfa_provisioning_uri_spec.rb
+++ b/apps/web/auth/spec/config/features/mfa_provisioning_uri_spec.rb
@@ -1,0 +1,151 @@
+# apps/web/auth/spec/config/features/mfa_provisioning_uri_spec.rb
+#
+# frozen_string_literal: true
+
+# Regression coverage for issue #3431 — "MFA QR code encodes wrong secret".
+#
+# Root cause
+# ----------
+# With otp_keys_use_hmac? enabled, the secret an authenticator app must use is
+# the HMAC'd key (Rodauth's otp_user_key, surfaced in the JSON otp-setup
+# response as otp_setup), NOT the raw key (otp_raw_secret, which exists only for
+# the setup handshake). The SPA used to rebuild the otpauth:// URI itself and,
+# on the HMAC path, encoded the raw key — so scanned codes never matched what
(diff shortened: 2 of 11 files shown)


delano:
Thanks for the review. Addressed in the latest push (force-update to `a1db683f`):

- **Silent no-QR when `provisioning_uri` is missing (P1):** `enrichSetupResponse` now fails fast on the HMAC path — it logs, pings Sentry via `captureMessage` when diagnostics is enabled, and returns `null`, so the caller surfaces an error instead of advancing to a blank scan step. Added a regression test (`surfaces an error when HMAC 422 omits provisioning_uri`).
- **"HMAC setup without password" missing the QR assertion:** added the `qr_code` + `expect(QRCode.toDataURL).toHaveBeenCalledWith(...)` assertions to that test.
- **Top-level `include AuthTestConstants` — leaving as-is.** This is the established convention for the auth feature specs here: `mfa_spec.rb:16` and `feature_combinations_spec.rb:15` both do a top-level `include AuthTestConstants`. The constants (e.g. `STATUS_VERIFIED`) are referenced from top-level `let` blocks, whose lexical constant lookup resolves via `Object`'s ancestry — which is exactly what the top-level include provides. Moving it inside `describe` only here would diverge from its siblings and risk breaking that resolution. Happy to revisit repo-wide if you'd prefer the scoped form everywhere.

---

PR #3438, opened 2026-06-12
Fix MFA QR code to encode correct HMAC secret (issue #3431) [develop]

### LINKED ISSUE #3431: MFA QR code encodes wrong secret; manual setup key works

delano:
## Summary

Fixes #3431 on `develop` — the develop-targeted twin of #3437 (which targets `main`).

The MFA QR code encoded the wrong secret when HMAC mode was enabled. The backend generates two secrets during setup:
- `otp_raw_secret`: raw key used only for the setup handshake
- `otp_setup`: the HMAC'd key the authenticator app must actually use (and the value the server validates against)

The frontend reconstructed the `otpauth://` URI and encoded `otp_raw_secret` instead of `otp_setup`, so scanned QR codes never matched what the server validates (typing the manual key worked, scanning didn't).

**Fix:** the backend now emits Rodauth's authoritative `otp_provisioning_uri` (built from the correct HMAC'd secret *and* the brand-configured issuer) as `provisioning_uri` in the otp-setup response. The frontend renders that URI directly as a QR code with no client-side reconstruction.

## Scope — fix only

This PR is the **bug fix only** (7 files). The now-redundant `brand_totp_issuer` plumbing — which existed solely to feed the client-side reconstruction this fix removes — is split into the follow-up #3439 so each change stays single-concern. The branding feature is otherwise untouched here: `brand_totp_issuer` is still defined and serialized; the wizard just no longer reads it. **Merge this PR before #3439.**

### Backend (Ruby/Rodauth)
- **apps/web/auth/config/hooks/mfa.rb**: `before_otp_setup_route` hook emits `provisioning_uri` from Rodauth's `otp_provisioning_uri`
- **apps/web/auth/spec/config/features/mfa_provisioning_uri_spec.rb**: route-level regression specs — the provisioning URI embeds the HMAC'd secret (not the raw secret); codes derived from it complete setup; codes derived from `otp_raw_secret` are rejected (the original bug)

### Frontend (TypeScript/Vue)
- **src/shared/composables/helpers/mfaHelpers.ts**: `generateQrCode()` renders the backend's `provisioning_uri` directly; drops `siteName`/`email` reconstruction
- **src/shared/composables/useMfa.ts**: `setupMfa()` no longer needs issuer/email; renders QR from `provisioning_uri` in both standard and HMAC flows
- **src/apps/session/components/OtpSetupWizard.vue**: calls `setupMfa()` with no issuer/email plumbing
- **src/schemas/api/auth/responses/auth.ts**: adds `provisioning_uri` to `otpSetupResponseSchema`
- **src/tests/composables/useMfa.spec.ts**: asserts the QR is rendered from `provisioning_uri`

## Relationship to #3437

Identical fix; the only difference is the wizard — on `develop` it drops the `brand_totp_issuer` read, on `main` (#3437) it drops `site_name`, because the two branches start from different wizard code.

## Test Plan

- Backend `mfa_provisioning_uri_spec.rb`: 3/3 (exercises the real otp-setup route with the production hook).
- Frontend `useMfa.spec.ts`: 33/33; type-check clean; full frontend suite green.

### DIFF
--- a/apps/web/auth/config/hooks/mfa.rb
+++ b/apps/web/auth/config/hooks/mfa.rb
@@ -32,6 +32,36 @@ module Auth::Config::Hooks
       # - Verification attempts (POST with OTP code)
       #
       auth.before_otp_setup_route do
+        # ----------------------------------------------------------------
+        # Emit Rodauth's authoritative TOTP provisioning URI (issue #3431)
+        # ----------------------------------------------------------------
+        #
+        # By the time this hook body runs, Rodauth's json feature has already
+        # executed (via super) its own before_otp_setup_route, which — in the
+        # HMAC JSON flow — generates the setup secret and populates the
+        # response with:
+        #   - otp_setup_param     => otp_user_key (the HMAC'd key the
+        #                            authenticator must actually use)
+        #   - otp_setup_raw_param => the raw key, used ONLY for the setup
+        #                            handshake (NOT a valid TOTP secret)
+        #
+        # The SPA historically reconstructed the otpauth:// URI itself and, on
+        # the HMAC path, encoded the raw key — so scanned codes never matched
+        # what the server validates (otp built from otp_user_key), while the
+        # manually-displayed otp_setup key worked. We instead emit Rodauth's
+        # own otp_provisioning_uri (built from otp_user_key, with the
+        # brand-configured otp_issuer and correct digits/period) as the single
+        # source of truth, so the frontend can render the QR directly without
+        # guessing the secret, issuer, or params.
+        #
+        # Rebuild the tmp OTP key from the raw secret we just handed the client
+        # so otp_provisioning_uri reflects exactly that secret, independent of
+        # any later key regeneration in the route body.
+        if otp_keys_use_hmac? && (raw_secret = json_response[otp_setup_raw_param])
+          otp_tmp_key(raw_secret)
+          json_response[:provisioning_uri] = otp_provisioning_uri
+        end
+
         is_post      = request.post?
         has_otp_code = !param_or_nil(otp_auth_param).to_s.empty?
         has_password = !param_or_nil(password_param).to_s.empty?
--- a/apps/web/auth/spec/config/features/mfa_provisioning_uri_spec.rb
+++ b/apps/web/auth/spec/config/features/mfa_provisioning_uri_spec.rb
@@ -1,0 +1,153 @@
+# apps/web/auth/spec/config/features/mfa_provisioning_uri_spec.rb
+#
+# frozen_string_literal: true
+
+# Regression coverage for issue #3431 — "MFA QR code encodes wrong secret".
+#
+# Root cause
+# ----------
+# With otp_keys_use_hmac? enabled, the secret an authenticator app must use is
+# the HMAC'd key (Rodauth's otp_user_key, surfaced in the JSON otp-setup
+# response as otp_setup), NOT the raw key (otp_raw_secret, which exists only for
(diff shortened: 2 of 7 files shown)


delano:
Thanks for the review. Addressed in the latest push (force-update to `4bf47983`):

- **Silent no-QR when `provisioning_uri` is missing:** `enrichSetupResponse` now fails fast on the HMAC path — it logs, pings Sentry via `captureMessage` when diagnostics is enabled, and returns `null`, so the caller surfaces an error instead of leaving the user on a blank scan step. Added a regression test covering the 422-with-secrets-but-no-`provisioning_uri` (version-skew) path.
- **Top-level `include AuthTestConstants` — leaving as-is.** It's the convention in the sibling auth feature specs (`mfa_spec.rb:16`, `feature_combinations_spec.rb:15`), and the constants are referenced from top-level `let` blocks whose lexical lookup needs them on `Object`'s ancestry. Moving it inside `describe` only here would diverge from those and risk constant-resolution breakage. Happy to do a repo-wide scoped-include pass separately if preferred.

---

PR #3439, opened 2026-06-12
Remove redundant brand_totp_issuer now the backend owns the QR (issue #3431)

delano:
## Summary

Follow-up cleanup to #3438. Removes the `brand_totp_issuer` value piped to the SPA, now that the backend's `provisioning_uri` is the single source of truth for the MFA QR code.

The SPA used to reconstruct the `otpauth://` URI client-side and needed `brand_totp_issuer` to do it. #3438 removes that reconstruction (the QR is rendered from the backend's `provisioning_uri`, which already carries the brand-configured issuer), so the piped value is now dead weight. This removes it end-to-end.

The brand totp_issuer **config source stays** — the backend's `otp_issuer` reads it and bakes it into `provisioning_uri`. Only the now-unused frontend plumbing and its serializer field are removed.

## ⚠️ Depends on #3438 — merge that first

This branch is built on top of #3438 because the cleanup is only safe once the wizard stops reading `brand_totp_issuer` (remove the field first and the unfixed SPA references a field that no longer exists → type-check fails).

As a result, **this PR's diff currently includes #3438's fix commit** (15 files / 2 commits). Merge #3438 first; GitHub then reduces this PR to the 8-file cleanup below. Note that because the fix commit lives on this branch too, merging *this* PR would also bring in the fix — so please land #3438 first.

(If you'd rather this PR show only the 8 cleanup files right now, say the word and I'll set its base to `claude/mfa-qr-provisioning-uri`.)

### Backend (Ruby)
- **apps/web/core/views/serializers/config_serializer.rb**: stop emitting `brand_totp_issuer`
- **apps/web/core/views/helpers/initialize_view_vars.rb**: drop the `brand_totp_issuer` view-var
- **apps/web/core/spec/views/serializers/config_serializer_spec.rb**: drop the `brand_totp_issuer` serializer assertions
- **try/unit/config/config_serializer_brand_color_try.rb**: drop `brand_totp_issuer` from the expected output

### Frontend (TypeScript)
- **src/schemas/contracts/bootstrap.ts**: remove `brand_totp_issuer` from the bootstrap schema
- **src/shared/stores/bootstrapStore.ts**: remove `brand_totp_issuer`
- **src/tests/contracts/bootstrap-serializer-fields.ts**: remove from `CONFIG_SERIALIZER_FIELDS`
- **src/tests/fixtures/bootstrap.fixture.ts**: remove `brand_totp_issuer`

## Test Plan

Full frontend suite green; backend config-serializer spec green (brand block reduced to the remaining fields); type-check clean. The brand totp_issuer config → backend `otp_issuer` → `provisioning_uri` path is exercised by #3438's `mfa_provisioning_uri_spec.rb`.

---
_Generated by [Claude Code](https://claude.ai/code/session_012j3QYu572Ms9rHPDXxGM4V)_

### DIFF
--- a/apps/web/auth/config/hooks/mfa.rb
+++ b/apps/web/auth/config/hooks/mfa.rb
@@ -32,6 +32,36 @@ module Auth::Config::Hooks
       # - Verification attempts (POST with OTP code)
       #
       auth.before_otp_setup_route do
+        # ----------------------------------------------------------------
+        # Emit Rodauth's authoritative TOTP provisioning URI (issue #3431)
+        # ----------------------------------------------------------------
+        #
+        # By the time this hook body runs, Rodauth's json feature has already
+        # executed (via super) its own before_otp_setup_route, which — in the
+        # HMAC JSON flow — generates the setup secret and populates the
+        # response with:
+        #   - otp_setup_param     => otp_user_key (the HMAC'd key the
+        #                            authenticator must actually use)
+        #   - otp_setup_raw_param => the raw key, used ONLY for the setup
+        #                            handshake (NOT a valid TOTP secret)
+        #
+        # The SPA historically reconstructed the otpauth:// URI itself and, on
+        # the HMAC path, encoded the raw key — so scanned codes never matched
+        # what the server validates (otp built from otp_user_key), while the
+        # manually-displayed otp_setup key worked. We instead emit Rodauth's
+        # own otp_provisioning_uri (built from otp_user_key, with the
+        # brand-configured otp_issuer and correct digits/period) as the single
+        # source of truth, so the frontend can render the QR directly without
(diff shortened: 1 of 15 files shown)


delano:
Thanks for the review. The **silent no-QR on missing `provisioning_uri`** case is fixed in the fix commit this branch is stacked on (#3438): `enrichSetupResponse` now fails fast on the HMAC path — logs, pings Sentry via `captureMessage` when diagnostics is enabled, and returns `null` (with a regression test), instead of returning success with no `qr_code`.

This branch has been rebased onto that updated fix (`4bf47983`); its own diff remains the `brand_totp_issuer` end-to-end removal. As before, please merge #3438 first.

---

PR #3450, opened 2026-06-13
Honour continue=false on burn endpoints (v1 and v2)

delano:
## Summary

The v1 and v2 `BurnSecret` logics both parse the confirmation flag into a real boolean in `process_params`:

```ruby
@continue = [true, 'true'].include?(params['continue'])
```

…but then compute the greenlight from the **raw** param instead of the parsed value:

```ruby
continue_result = params['continue']
@greenlighted   = viewable && correct_passphrase && continue_result
```

In Ruby every non-empty string is truthy, so a request carrying the string `"false"` — the common shape for form/query submissions — yields `continue_result = "false"` (truthy) and **burns the secret anyway**, destroying it against the caller's explicit intent. The whole point of the burn confirmation is to gate this destructive, irreversible action.

## Fix

Use the already-parsed `continue` boolean in the greenlight check (and in the v2 debug log) so only a genuine truthy confirmation burns the secret. This matches how `show`/`reveal` already parse and consume the flag (`params['continue'].to_s == 'true'`).

```ruby
@greenlighted = viewable && correct_passphrase && continue
```

## Notes

- Behaviour for the intended happy path is unchanged: `continue=true` / `continue: true` still burns.
- The only behavioural change is that `continue="false"` (and other non-truthy string values) now correctly does **not** burn — previously it did.
- Existing `apps/api/v1/spec/logic/secrets/burn_secret_spec.rb` sends `'continue' => 'true'` and still greenlights.
- `ruby -c` clean on both files.

🤖 Generated with [Claude Code](https://claude.ai/code)

---
_Generated by [Claude Code](https://claude.ai/code/session_0116X8RbQ4qVt7jbPLF1h6Ur)_

### DIFF
--- a/apps/api/v1/logic/secrets/burn_secret.rb
+++ b/apps/api/v1/logic/secrets/burn_secret.rb
@@ -33,11 +33,9 @@ module V1::Logic
 
           @correct_passphrase = !potential_secret.has_passphrase? || potential_secret.passphrase?(passphrase)
           viewable = potential_secret.viewable?
-          # Use the parsed @continue boolean (set in process_params), not the raw
-          # param. The raw value is a string for form/query requests, and every
-          # non-empty string — including "false" — is truthy in Ruby, so reading
-          # params['continue'] directly would burn the secret even when the
-          # client explicitly sent continue=false.
+          # Use the parsed boolean (true / 'true' only), not the raw param: the
+          # raw value treats any non-empty string as truthy, so a deliberate
+          # `continue=false` would burn the secret anyway.
           @greenlighted = viewable && correct_passphrase && continue
 
           if greenlighted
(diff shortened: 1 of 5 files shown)


delano:
Addressed the missing-v2-coverage note: added `apps/api/v2/spec/logic/secrets/burn_secret_spec.rb` (commit b1ef3f5). It exercises `V2::Logic::Secrets::BurnSecret#process` in isolation (`success_data` stubbed, since its URL/serialization chain is out of scope for the greenlight decision) and asserts `continue='false'` neither greenlights nor calls `burned!`, while `continue='true'` still burns. Confirmed the false-case examples fail against the pre-fix raw-param code and pass with the fix, on Ruby 3.4.9.

---

PR #3451, opened 2026-06-13
Log the real secret id on receipt reveal/burn transitions

delano:
## Summary

`Receipt#revealed!`, `Receipt#burned!`, and `Receipt#expired!` clear `secret_identifier` to `''` **before** building their audit-log payload, then log the now-empty field:

```ruby
self.secret_identifier = ''
save update_expiration: false

secret_logger.info 'Receipt state transition to revealed',
  {
    receipt_id: shortid,
    secret_id: secret_identifier,   # always "" at this point
    ...
  }
```

So every reveal/burn/expire event is logged with `secret_id: ""`, severing the link between the audit event and the secret it concerns — exactly the information you'd want during incident review on a secret-sharing service.

Only the sibling transition `orphaned!` already gets this right: it snapshots the id into a local (`original_secret_id`) before clearing.

> **Correction:** an earlier revision of this description claimed `expired!` was already correct. That was wrong — `expired!` had the identical defect (caught in review), and is now fixed here alongside `revealed!`/`burned!`.

## Fix

Capture `original_secret_id = secret_identifier` before clearing the field in `revealed!`, `burned!`, and `expired!`, and log that local — matching the existing pattern in `orphaned!`.

Log-only change; no behavioural or stored-data impact. Verified with `try/unit/models/receipt_state_terminology_try.rb` (47/47 pass, which exercises all three transitions) on Ruby 3.4.9.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/changelog.d/20260613_124500_claude_receipt_audit_log_secret_id.rst
+++ b/changelog.d/20260613_124500_claude_receipt_audit_log_secret_id.rst
@@ -3,10 +3,10 @@
 Fixed
 -----
 
-- The "Receipt state transition to revealed/burned" audit log lines now
-  record the actual secret identifier. Both ``Receipt#revealed!`` and
-  ``Receipt#burned!`` cleared ``secret_identifier`` to an empty string before
-  building the log payload, so every reveal/burn event was logged with
-  ``secret_id: ""`` — defeating the trail for incident review. The identifier
-  is now captured before it is cleared (matching ``orphaned!`` and
-  ``expired!``), so the log reflects which secret the event refers to.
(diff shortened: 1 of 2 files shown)


delano:
Good catch on `expired!` — you're right, it had the identical defect and my original description was wrong to say it was already correct. Fixed in 77883ff: `expired!` now captures `original_secret_id` before clearing the field, same as `revealed!`/`burned!`. Updated the changelog and PR description accordingly. `try/unit/models/receipt_state_terminology_try.rb` (which exercises all three transitions) stays green at 47/47 on Ruby 3.4.9.

---

PR #3455, opened 2026-06-13
MFA hardening: accurate recovery-code docs + loud failure on missing QR URI

delano:
Two related MFA robustness/clarity fixes surfaced while reviewing the #3431 QR rework.

## 1. Recovery-code comment was factually wrong (`apps/web/auth/config/features/mfa.rb`)

The comment claimed "8-character alphanumeric codes (like: 4k9m-x2pq) — ~1.7 billion (36^8)", but `new_recovery_code` uses `Familia.generate_trace_id`, which emits a **CSPRNG-backed 64-bit base36 token** (~13 chars, ~1.8e19 ≈ 2⁶⁴, no dash). Corrected the comment to the real format/entropy and recorded why 64-bit is the deliberate choice (server-side, rate-limited, account-bound verification; shorter is easier to type when locked out). Generator unchanged — bumping to a higher tier doesn't escape Familia's "resist intentional guessing" caveat and only lengthens the code.

## 2. Blank-QR gap on the non-HMAC (200) setup path (`useMfa.ts` / `mfaHelpers.ts`)

The 422 (HMAC) path already failed loudly when `provisioning_uri` was missing; the 200 path set `setupData` with an undefined `qr_code` → wizard bound to a **blank scan step with no error**. Extracted the guard into a shared `renderSetupQr` helper used by both paths; the 200 path now surfaces an error. Latent today (HMAC always on), but removes the asymmetry.

## Tests + validation (run locally)

- New `useMfa.spec.ts` case: a 200 response with secrets but no `provisioning_uri` returns null, leaves `setupData` null, surfaces `internal_error`, and never calls `QRCode.toDataURL`.

✅ **`vitest run src/tests/composables/useMfa.spec.ts` → 35/35 pass** (incl. the new case) and **`rspec mfa_spec.rb` → 21/21 pass** (the Ruby change is comment-only; spec confirms boot/config untouched). The earlier "couldn't run locally" caveat is resolved.

https://claude.ai/code/session_01Jt6ne55BRYMKMkfSb3nrAV

### DIFF
--- a/apps/web/auth/config/features/mfa.rb
+++ b/apps/web/auth/config/features/mfa.rb
@@ -62,8 +62,15 @@ module Auth::Config::Features
       # after_login hook in apps/web/auth/config/hooks/login.rb which checks
       # uses_two_factor_authentication? and sets json_response[:mfa_required] = true
 
-      # Generate 8-character alphanumeric codes (like: 4k9m-x2pq)
-      # This provides ~1.7 billion possible codes (36^8)
+      # Recovery codes are CSPRNG-backed (SecureRandom) 64-bit values rendered
+      # in base36 — roughly 13 characters, e.g. "3w5e11264sgsf", with ~1.8e19
+      # (2**64) possibilities. 64 bits is deliberately chosen over a longer
+      # token: recovery codes are verified server-side, are rate-limited, and
+      # are bound to a single account, so they are not an offline-guessable
+      # artifact — and a shorter code is far less error-prone for a user to type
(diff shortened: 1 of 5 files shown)


delano:
Superseded by #3459, which cherry-picks these 3 commits (5447f68, 569e6c6, b015c90) onto `main` for the v0.25.10 release. `develop` will receive the change when `main` is merged back into it, so this develop-targeted PR is redundant. Closing.

PR #3518, opened 2026-06-23
Correlate API error responses with logs via request_id

delano:
## Problem

A backend JSON API error (e.g. a v3 `404`) returns a payload like:

```json
{"error_type":"RecordNotFound","error_id":"5955f4b712ce5475"}
```

That `error_id` is minted **inside the Otto gem** (`Otto::Core::ErrorHandler`, `SecureRandom.hex`), and it's a poor support handle:

- **Dev-only in the body** — Otto only adds `error_id` to the response body when `RACK_ENV` is dev/development. In production the client never receives it.
- **Logged on a different logger, uncorrelated** — it goes to the `Otto` logger at the handler's level (`:info` for a 404), and that log context carries **no `request_id`**. Meanwhile `RequestLogger` logs the request on the `HTTP` logger (4xx→`:warn`) **with `request_id` but no `error_id`/`error_type`**. No single log line carries both, so you can't pivot `error_id ↔ request_id` — and where the `Otto` logger sits above `:info`, the line disappears entirely.

The reliable correlation id we already have is **`request_id`** (`x-request-id`, a Familia trace id): it's in the response header *and* the request log. Otto's `error_id` was the wrong thing to chase.

## Change

Lean into `request_id` as the one correlation id:

- **`OttoHooks`** — every typed error handler now runs its body through a nil-safe `with_error_correlation` helper that (a) echoes the request's `x-request-id` into the JSON error body and (b) stashes `error_type` into `env`.
- **`RequestLogger`** — includes `error_type` in the request-log payload when an error handler ran, so the single line that already carries `request_id` also names *what* failed (e.g. `RecordNotFound`), regardless of capture mode.
- **Docs** — documents the `request_id` correlation field in ADR-013 and the OpenAPI error schema.

Now an API consumer can quote the `request_id` from the error body (or the `x-request-id` header) and we grep it straight out of the request log — which also tells us the `error_type`.

### Deliberately unchanged

- The static `router.not_found`/`router.server_error` fallbacks (`BaseJSONAPI`) are request-independent, so they can't carry `request_id` in the body — the `x-request-id` response header still covers them, and the ADR-013 regression spec stays green.
- This does **not** try to surface Otto's own `error_id` in logs (it's dev-only and uncorrelated); standardizing on `request_id` is the robust fix. Propagating Otto's `error_id` would belong in the `delano/otto` gem.

## Testing

Run with Ruby 3.4.9 + Redis on `:2121`:

- `spec/unit/onetime/application/otto_hooks_spec.rb` — new correlation cases (request_id in body, error_type stashed in env, nil-safe) ✅
- `spec/unit/onetime/application/request_logger_spec.rb` — new; error responses log `error_type` next to `request_id` ✅
- `spec/integration/api/error_response_shape_spec.rb` — static fallbacks still ADR-013 ✅
- `spec/unit/onetime/errors_spec.rb`, `spec/integration/api/v2/plan_cache_miss_handling_spec.rb` — handler paths green ✅

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qb98MpWiA5kTmkM8DsMS1M)_


delano:
Thanks @greptile-apps — addressed in `7aabcd1`:

- **Issue 2 (minimal-mode loses correlation):** Fixed. `RequestLogger` now forces `request_id` onto error lines even under `:minimal` capture (the YAML default), so `error_type` and the client-visible id stay on the same greppable line. Unit coverage added.
- **Issue 1 (FormError without `error_type`):** Fixed. `with_error_correlation` falls back to the exception class name (e.g. `FormError`) for the stashed log `error_type` when `to_h` compacts it away. The response **body** contract is unchanged — this only enriches the request log. A FormError carrying a custom type (e.g. `email_mismatch`) is still preferred.
- **Issue 3 (schema requires `message`):** Fixed. Dropped `required: ['message']`. Error shapes genuinely vary across versions (frozen V1 is message-based; V2+ is ADR-013 `error`/`error_type` with no `message`), so no field is universally required and forcing `message` rejected valid V2+ bodies.
- **Issue 4 (Roda `/auth` surface skipped):** Deferred by design, not a fix in this PR. Extending `request_id` correlation to `Auth::ErrorTranslator` is a separate, larger change; I documented the current boundary in ADR-013 (request_id-in-body is an **Otto-app** guarantee today; `/auth` still relies on the `x-request-id` response header). Happy to do that as its own follow-up PR.

(Heads-up: this review ran against the `main`-merge commit `838fb90`; the fixes are in `7aabcd1` on top.)

---

PR #3529, opened 2026-06-24
fix(oauth): grant offline_access so refresh tokens survive scope intersection

delano:
## Summary

Addresses the greptile **P1** review feedback on #3239: `offline_access` was missing from the OAuth scope allow-lists, so refresh tokens are never issued in the real `/authorize` flow.

rodauth-oauth intersects requested scopes at **two** levels, and `offline_access` was absent from both:

1. **Server-level allow-list** — `oauth_application_scopes` in `apps/web/auth/config/features/oauth.rb` was `%w[openid profile email]`. Anything outside this set is stripped before the grant row is written.
2. **Per-application column** — the seeded dev SP's `scopes` in `apps/web/auth/initializers/seed_dev_oauth_client.rb` was `'openid email profile'`. rodauth-oauth runs a second intersection against this column.

With `:oidc` enabled, the gem only issues a `refresh_token` when `offline_access` survives into the granted scopes. The SP (`omniauth.rb`) requests `offline_access`, but it was stripped at both gates — so no refresh token was ever issued end-to-end, regardless of `grant_types`.

## Changes

- `features/oauth.rb` — add `offline_access` to `oauth_application_scopes`.
- `seed_dev_oauth_client.rb` — add `offline_access` to the seeded SP's `scopes` column.
- `oauth_idp_lifecycle_spec.rb` — remove the now-redundant `grant_types` workaround in the `before(:all)` block (the seed already sets `'authorization_code refresh_token'`) and its stale "real seed fix needed" comment.

## Notes

- The other open review threads on #3239 need no code change — they're already handled (RSA `\n` unescaping at `oauth.rb` load, `oauth_require_pkce true`, https-only URI default in prod) or outdated.
- Branched from the current `feature/3104-idp` tip, so the diff is exactly these three files.
- Not verified via rspec in this environment (container Ruby is 3.3.6; Gemfile requires ≥3.4.7). All three files pass `ruby -c`.

https://claude.ai/code/session_019dXw3KmSsz8tXuK66vjuX6

---
_Generated by [Claude Code](https://claude.ai/code/session_019dXw3KmSsz8tXuK66vjuX6)_

### DIFF
--- a/apps/web/auth/config/features/oauth.rb
+++ b/apps/web/auth/config/features/oauth.rb
@@ -100,7 +100,14 @@ module Auth::Config::Features
       auth.oauth_refresh_token_protection_policy 'rotation'
 
       # ─── Scopes ────────────────────────────────────────────────────────
-      auth.oauth_application_scopes %w[openid profile email]
+      # `offline_access` must be present here (the server-level allow-list) or
+      # rodauth-oauth strips it from the request before the grant row is
+      # written. With :oidc enabled, oidc.rb only issues a refresh_token when
+      # `offline_access` survives into the granted scopes, so dropping it here
+      # silently breaks refresh tokens for every client (the SP in omniauth.rb
+      # requests it). The seeded SP's per-application `scopes` column must list
+      # it too — see seed_dev_oauth_client.rb.
+      auth.oauth_application_scopes %w[openid profile email offline_access]
 
       # ─── PKCE enforcement ──────────────────────────────────────────────
       # :oauth_pkce (enabled above) makes PKCE *available* but not *mandatory*:
--- a/apps/web/auth/initializers/seed_dev_oauth_client.rb
+++ b/apps/web/auth/initializers/seed_dev_oauth_client.rb
@@ -12,8 +12,11 @@ module Auth
     # is reserved for local development — real production SP clients are
     # provisioned through a separate (manual or admin-tooled) path.
     #
-    # Idempotent: checks for an existing row by client_id and skips the insert
-    # if found. Safe to run on every boot.
+    # Idempotent: looks up the row by client_id. If it already exists, the
+    # drift-prone columns (scopes, grant_types) are reconciled to the current
+    # desired values so an environment seeded before a config change self-heals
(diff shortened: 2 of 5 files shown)


delano:
Thanks for the reviews — addressed all the feedback in `6abf947` and `4420b99`.

**greptile P2 — stale seed rows won't pick up `offline_access`** ✅
Rather than just documenting a manual reseed, the seed initializer now **self-heals**: instead of skipping on an existing `onetimesecret-sp-dev` row, it reconciles the drift-prone columns (`scopes`, `grant_types`) to the current desired values on boot. The reconcile is targeted (leaves the bcrypt secret untouched) and idempotent, so concurrent boots still converge and the unique-index/insert race path is preserved. `scopes`/`grant_types` are now shared constants (`DEV_CLIENT_SCOPES` / `DEV_CLIENT_GRANT_TYPES`) so insert and reconcile use one source of truth. Added a spec (`seed_dev_oauth_client_idempotency_spec`) that seeds a stale row and asserts it gains `offline_access` after the next run.

**claude[bot] — scope ordering inconsistency** ✅
Aligned the seed's ordering with `oauth_application_scopes` (`openid profile email offline_access`). Functionally irrelevant (treated as a set) but reads cleaner.

**claude[bot] — no E2E test for the real `/authorize` scope intersection** ✅ (partially)
Good catch — this was the one path with no coverage, since the lifecycle/protocol specs seed `oauth_grants` directly. The discovery document advertises `scopes_supported` derived from `oauth_application_scopes`, so I extended the discovery spec to assert `offline_access` is advertised. That's an HTTP-level regression guard for the **server allow-list** gate — a revert of the `oauth_application_scopes` change now fails a test. Combined with the stale-row self-heal spec guarding the **per-application** gate, both intersection points are now covered.

I deliberately did **not** add a full authenticated `/authorize` → `/token` flow test: it requires standing up a logged-in rodauth session, and I can't execute rspec in this environment (container Ruby is 3.3.6; the Gemfile requires ≥3.4.7), so I'd be committing an unverified, likely-flaky test. Happy to add the full-flow E2E test as a follow-up if you'd like it — just say the word.

---

Issue #3491, opened 2026-06-20
Clarify separation between plan entitlements and role capabilities in entitlement system


delano:
## Summary

The entitlement system currently serves dual purposes:

1. **Plan entitlements** (feature access): What the org's subscription includes (`incoming_secrets`, `custom_branding`, `audit_logs`)
2. **Role capabilities** (user permissions): What the user can do within the org (`manage_org`, `manage_members`, `manage_sso`)

Both are stored in the same `ENTITLEMENTS` constant, materialized into the same `membership.entitlements` array, and checked via the same `can?()` interface. This works but creates conceptual confusion.

## Problem Statement

The naming and structure obscure the authorization model:

- Frontend sees `ENTITLEMENTS.MANAGE_ORG` alongside `ENTITLEMENTS.INCOMING_SECRETS` — both look like feature flags
- Role capabilities like `manage_org` are defined in `OWNER_ENTITLEMENTS`, making roles appear to be plan-gated
- The two-stage authorization pattern (plan check + role check) is implicit rather than explicit

## Current Architecture (Works Correctly)

```ruby
# lib/onetime/models/organization_membership.rb
MEMBER_ENTITLEMENTS = Set['create_secrets', 'view_receipt', ...]     # feature access
ADMIN_ENTITLEMENTS  = Set['manage_members', 'incoming_secrets', ...] # mixed
OWNER_ENTITLEMENTS  = Set['manage_org', 'manage_sso', ...]           # role capabilities

# Materialization: org.entitlements ∩ ROLE_ENTITLEMENTS[role]
```

```ruby
# lib/onetime/logic/base.rb - require_entitlement_in!
membership = OrganizationMembership.find_by_org_customer(org.objid, cust.objid)
return true if membership.can?(entitlement)  # Already checks membership, not org
```

The backend correctly checks `membership.can?()` (user's role-scoped entitlements). The architecture is sound — the naming is confusing.

## Evidence

Issue #3479 demonstrates the confusion: the frontend checked `org.entitlements.includes('incoming_secrets')` without checking `manage_org`, because both looked like plan features. The backend's two-gate check (manage_org + incoming_secrets) wasn't obvious from the entitlement names.

## Proposal

Document and potentially restructure to make the two authorization dimensions explicit:

### Option A: Naming Convention (Low Effort)
- Prefix role capabilities: `role:manage_org`, `role:manage_members`
- Keep plan entitlements unprefixed: `incoming_secrets`, `custom_branding`
[...]

Issue #3479, opened 2026-06-15
Incoming Secrets panel shows contradictory enable prompt under entitlement error


delano:
## Problem Statement

On the custom-domain config page (Incoming Secrets section), the panel renders two contradictory messages at once:

- A red error: **"Error: Feature requires manage org entitlement"**
- An info box: **"Incoming secrets are currently disabled. External users cannot send secrets to recipients on this domain. Enable the feature below to allow incoming secrets."**

The info box directs the user to "enable the feature below," but the error indicates they lack the entitlement to do so. The two states should be mutually exclusive: if the user lacks the `manage org` entitlement, the enable affordance and its instructional copy should not be shown.

Observed on `secrets.example.com` (domain also flagged "DNS Incorrect", likely unrelated).

## Early Diagnosis

Likely the Incoming Secrets component renders the entitlement-error branch and the "feature disabled / enable below" branch independently rather than gating one on the other. Suspect the enable prompt + toggle are shown whenever the feature is off, without first checking entitlement. The error appears to come from the gated API call failing, while the UI copy still assumes an enable path is available.

## Affected Areas

- [ ] Frontend domain-config / Incoming Secrets panel component (Vue) — renders both states; needs entitlement gate before showing enable prompt
- [ ] Entitlement check feeding the panel — should hide/disable enable affordance and swap copy when `manage org` entitlement is absent

Issue #3225, opened 2026-05-25
Membership-level entitlements, materialized


delano:
## Problem

Entitlements and roles are two independent authorization systems in this codebase. They never intersect. There is no mechanism for differentiating what an organization **owner** or **admin** can do from what an ordinary **member** can do via the entitlement system.

The role hierarchy exists on `OrganizationMembership` (`lib/onetime/models/organization_membership.rb:89`):

```
owner > admin > member
```

But the entitlement check is org-scoped:

```ruby
# lib/onetime/logic/base.rb:174
def require_entitlement!(entitlement, error_key: nil)
  # ...
  return true if auth_org.can?(entitlement)
  # ...
end
```

`auth_org.can?(entitlement)` reads the org's `materialized_entitlements` Redis set. The result is identical for every member of the org regardless of role. The middleware gate at `lib/middleware/entitlement_check.rb:62` has the same shape.

The only role-aware checks are three discrete guards in `lib/onetime/application/authorization_policies.rb`:

- `verify_organization_owner` (line 183) — used by `update_organization`, `delete_organization`, `update_member_role`, domain config
- `verify_organization_admin` (line 218) — used by `create_invitation`, `list_invitations`, `resend_invitation`, `revoke_invitation`, `add_domain`
- `verify_organization_member` (line 200) — basic membership check

These guards must be called explicitly in each logic class's `raise_concerns`. They are not composed with `require_entitlement!`, and there is no convention or test that forces them to be paired.

### 1. Present-day exploit (active regression)

`apps/api/domains/logic/domains/{get,update,remove}_domain_{brand,image}.rb` inherit `DomainsAPI::Logic::Base` directly — **not** the `*_config/base.rb` classes that include `DomainConfigAuthorization`. They gate only on `require_entitlement!('custom_branding')`. No `verify_organization_*`.

In self-hosted mode (`STANDALONE_ENTITLEMENTS` at `lib/onetime/models/features/with_entitlements.rb:72` grants `custom_branding` to every org), any member of any org can mutate branding on the org's domains. `custom_privacy_defaults` has the same latent shape — `validate_privacy_defaults_entitlement` at `update_domain_brand.rb:170` calls `require_entitlement!('custom_privacy_defaults')` but its only invoker at line 48 is commented out.

Closing this is a small focused PR independent of the broader RBAC work: include `DomainConfigAuthorization` (or call `verify_organization_owner` directly) in those five classes.

### 2. Privilege escalation by omission (latent)

[...]

PR #3546, opened 2026-06-29
docs(entitlements): prepare full separation of plan entitlements from role capabilities (#3491)

delano:
Read-only preparation/design research for issue #3491. Maps the current
entitlement system end-to-end and lays out a staged plan to fully separate
organization plan-entitlements (subscription features) from membership
role-capabilities (role permissions) — Option C in the issue.

Key findings documented:
- Plan features and role capabilities share one undifferentiated ENTITLEMENTS
  namespace across the billing catalog, ROLE_ENTITLEMENTS sets, the materialized
  membership blob, the can?()/require_entitlement! interface, the wire, and the
  frontend can() composable; the system cannot tell why a check failed.
- The single org.entitlements ∩ ROLE_ENTITLEMENTS[role] intersection strips role
  capabilities in billing mode (plans never list manage_*); standalone mode only
  masks this by stuffing manage_* into STANDALONE_ENTITLEMENTS.
- Two divergent enforcement paths read the same namespace (membership.can? vs
  org.can?), and the one explicit two-gate site admits the paths can diverge.
- Frontend can() ignores role entirely (standalone short-circuits true for all),
  the structural root of bug #3479.
- Drift inventory: ADR-012 tiers vs shipped code, orphaned manage_orgs/manage_teams,
  manage_billing missing from the FE constant, stale source citations, free-tier
  limit mismatches.

Includes an authoritative plan-feature/role-capability/limit taxonomy, a
severity-ranked drift and coupling audit, the Option C target design (two
registries, two predicates, two gates, two wire fields, two composables), an
A→B→C migration/sequencing plan with backward-compat and data-migration notes,
test-impact analysis, risks/open questions, and acceptance-criteria mapping.

Follow-up revisions on this branch:
- Ratified decisions recorded in §8: D1 (manage_org stays owner-exclusive; new
  admin+ manage_domains capability for the domain-config surface) and D2
  (per-member feature overrides dropped; per-member capability overrides stay).
- Independent re-analysis folded in: §1.2 on why the fusion exists (ADR-012's
  deliberate cure for unpaired role guards, #3225) and a required pairing
  countermeasure for the two-gate pattern; a Stage B tier-preservation decision
  list (role-tiering of plan features is live enforcement — re-typing
  add_domain.rb to a bare plan gate would silently relax privileges); explicit
  colonel-bypass/anonymous rulings and HTTP-status semantics for the split
  gates; clarification that the materialize_for_role! header/body divergence is
  deliberate and self-documented (fix the header only); L5 (the unmaterialized
  fallback re-implements the intersection and skips grants−revokes) and L6
  (colonel preview mode pinned to the feature axis); Stage 0 docs sweep
  extended beyond ADR-012 (membership-entitlements.md, FE quick reference,
  preview-mode diagram).

Refs #3491, #3479, ADR-012.

Co-Authored-By: Claude Opus 4.8 
Claude-Session: https://claude.ai/code/session_014BBsqDyLrwKiTZ6gi5Yeit


delano:
Thanks for the thorough reviews (@copilot-pull-request-reviewer, @greptile-apps, and the Claude review bot). All feedback addressed in 77d9544 — since this is a spec that will drive implementation, accuracy matters, so I verified each disputed fact against source before editing.

**Corrections (factual)**
- **Taxonomy counts (Greptile P2 — confirmed wrong).** Re-counted from source: the billing catalog has **23** `entitlements:` definitions (not 24) and `ENTITLEMENTS` has **22** runtime keys (not 23). Updated §3 to `17 plan_feature + 6 role_capability = 23 = catalog defs`, `9` limits, and reconciled M8 to `KNOWN_ENTITLEMENTS 23 / STANDALONE 22 / catalog 23`. Added an explicit note that the Stage A CI assertion must **recompute** partitions from source rather than hard-code totals — exactly so a catalog edit can't silently invalidate them.
- **`materialize_for_role!` citation inconsistency (Copilot + bot).** The "claims `org.materialized_entitlements`" text is the module-header *Materialization formula* comment, while the method body uses `org.entitlements`; the method itself is well below the `:13-18` header. Re-cited both by **symbol name** (not line number) in §2.2 and Appendix B.

**Provenance / citation drift (Copilot).** Reworked the provenance block: line numbers are now stated as best-effort snapshots that will drift, with **file path + symbol name** as the durable anchor and a note to `grep` the symbol to relocate.

**Open question resolved (bot Q3).** Verified `Billing::Controllers::Entitlements#show` returns `entitlements: org.entitlements_for_request(session)` — **org-level plan entitlements, no role filter**. So the endpoint's "org-level" framing is correct and the fix is a **re-label** (`org.entitlements` → `org.plan_entitlements`) plus a **new `role_capabilities` field**, not a re-scope. Folded into §8.

**Clarifications added**
- **§5.1** — comment that owners inherit `manage_members` transitively via the `ROLE_CAPABILITIES` union (owner ⊇ admin ⊇ member), so it's intentionally not re-listed in `OWNER_CAPABILITIES`.
- **L1** — added the recommended dual-natured-token pattern: plan flag for "org has X" + a distinct capability for "role may use X"; only split where both questions are real.
- **§6** — added the wire conflict rule: when both new fields and the legacy `entitlements` array are present, **new fields win, legacy ignored** (never union/intersect).
- **§7** — made explicit that `entitlement_keys_spec.rb`'s derivation source changes from `STANDALONE` to the two constants `PLAN_FEATURES` + `ROLE_CAPABILITIES`.

**Product questions (bot) — recommended, non-binding, deferred to the superseding ADR**
- **`manage_org` owner-vs-admin (M3):** recommend keeping `manage_org` owner-exclusive for org-lifecycle (rename/delete/billing/SSO) and adding a distinct **admin+** `manage_domains` capability for the domain-config surface admins already reach via route guards — aligning the model with shipped behavior instead of widening `manage_org`.
- **Per-member plan-feature overrides:** recommend treating per-member *feature* overrides as **unsupported** (features are org/subscription-scoped) while keeping per-member *capability* overrides; no current spec/CLI exercises a per-member feature revoke, so Stage C can drop that path — pending confirmation against the colonel/operator workflows.

I've left these two as recommendations rather than binding decisions since they're auth-semantics calls for @delano to ratify in the ADR that supersedes ADR-012.

---

delano:
Closing the loop on the two product questions I'd left as non-binding recommendations above — both are now **ratified** and folded into the spec as decisions (`73ab10f`):

- **D1 — `manage_org` stays owner-exclusive; add an admin+ `manage_domains` capability.** The domain-config surface admins already reach via route guards gets its own capability rather than widening `manage_org`, which remains owner-only for org-lifecycle (rename/delete/billing/SSO). Reflected in §5.1 (`ADMIN_CAPABILITIES` gains `manage_domains`), M3, and §8.
- **D2 — per-member *feature* overrides are unsupported; per-member *capability* overrides stay.** Features are a property of the org's subscription, not the member; the colonel grant/revoke path stays capability-only. Reflected in L4, Stage C, and §8.

No behavior change — these are decisions the superseding Option C ADR will carry forward. With the taxonomy counts corrected against source and the citations re-anchored by symbol, the doc is internally consistent and, from my side, ready to merge as the design input for #3491.

---

PR #3552, opened 2026-06-30
Add self-service verification email resend endpoint

delano:
## Summary

Implements a self-service "resend verification email" recovery flow for unverified accounts that cannot log in. This endpoint allows users to request a fresh verification email without authentication, following an anti-enumeration contract that returns identical responses across all account states.

### Changes

**Backend (Ruby)**
- Added `POST /api/account/resend-verification-email` endpoint via `AccountAPI::Logic::Account::ResendVerifyAccount`
- Delegates to Rodauth's `verify_account_resend` internal request method
- Returns uniform `{ sent: true }` response for all outcomes (nonexistent, verified, unverified, throttled, errors)
- Emits differentiated server-side audit events (`verify_account_resend_sent`, `verify_account_resend_blocked`, `verify_account_resend_error`, `verify_account_resend_noop`) without leaking state in HTTP response
- Handles feature-disabled gracefully (no-op in non-full modes)

**Frontend (Vue/TypeScript)**
- Added `resendVerificationEmail()` method to `useAuth` composable
- Sends login email, CSRF token, and locale to the endpoint
- Returns `true` on accepted requests (uniform 200 response)
- Returns `false` and sets error on malformed requests
- Added UI in `VerifyAccount.vue` with email input and neutral confirmation message
- Added i18n strings for help text and confirmation

**Testing**
- Comprehensive integration test suite (`resend_verification_email_spec.rb`) covering:
  - Enumeration uniformity across account states
  - Throttling behavior
  - Malformed request handling
  - CSRF exemption for `/api/` paths
  - Server-side audit log differentiation
- Frontend unit tests (`useAuth.resendVerification.spec.ts`) verifying:
  - Correct endpoint and payload
  - Uniform success response handling
  - Error handling and field errors
  - Network/server failure resilience
- Mechanism guard test for Rodauth internal request availability

## Test Plan

All tests pass:
- Integration tests verify the anti-enumeration contract (identical 200 + body across all account states)
- Integration tests confirm server-side audit events differentiate outcomes
- Frontend unit tests verify composable behavior and error handling
- Mechanism test guards the Rodauth `verify_account_resend` internal request method exists when feature is enabled
- Tests skip gracefully when `verify_account` feature is disabled at boot (RACK_ENV=test default)

https://claude.ai/code/session_0137pMAX5bnvaa5y8AnBkQZo


delano:
Thanks for the thorough review (Greptile + Claude). Addressed in `cc5a677`. Summary of each finding:

### P1 — fixed

**SSO-Only Mode Breaks Uniformity** (`resend_verify_account.rb`) — `require_non_sso_only!` (which raises `Forbidden` → 403) is removed from `raise_concerns`. SSO-only is now folded into the uniform `{ sent: true }` no-op via a new `verify_account_resend_available?` (`full_enabled? && verify_account_enabled? && !sso_only_enabled?`), alongside the verify-account-disabled and non-full-mode cases. An unauthenticated probe can no longer distinguish deployment auth config. Added integration Scenario 10 asserting SSO-only returns 200, not 403.

**Failed Resend Shows Success** (`VerifyAccount.vue`) — the confirmation now replaces the form **only** when the request was accepted; network/5xx failures keep the form and show a retryable error.

### Claude review items

2. **No error feedback** — the resend form now renders a neutral, retryable error (`web.auth.verify.resend_error`) on failure.
3. **Enter-key submit** — the resend UI is now a real `<form @submit.prevent>`, so Enter submits and `type="email"` gets native validation.
5. **Status-code inconsistency** — confirmed: `raise_form_error` → `Onetime::FormError` → **422** (`otto_hooks.rb:59`), so the integration spec's "blank login → 4xx" was correct and the frontend test's `200`-error-body mock was not. The composable spec now exercises the realistic **422** path and the misleading comments are corrected.
6. **Resend missing from the no-key state** — good catch. The resend form is now a reusable `ResendVerificationForm` component used in **both** the verify-error and missing-key states (a natural entry point for users who never clicked their link). New component spec covers accepted→confirmation, failed→error+retry, empty→no-call, loading→disabled.

4. **Email-format validation** — **intentionally not adding this.** `valid_email?` uses Truemail (network-capable validation); on an endpoint whose entire purpose is a byte-identical, account- and config-opaque response, a per-request email validation introduces a domain-level network/timing signal — the opposite of what we want here. The malformed case is already uniform-safe (it routes through the same no-op/blocked path), and `type="email"` gives native client-side validation. The minor audit-log conflation (malformed vs blocked) isn't worth that trade-off. Happy to revisit with a local regex-only check if you'd prefer the audit separation.

Frontend checks green locally (vitest incl. the new specs, `vue-tsc`, eslint). The Ruby integration suite runs in CI.

---

Issue #3557, opened 2026-06-30
Extract a shared exception → {status, log_level} registry for the Otto and Roda error layers


delano:
## Context

The typed-error → HTTP-status / log-level mapping is currently encoded twice, once per routing stack:

- **Otto apps** register it imperatively in `Onetime::Application::OttoHooks#configure_otto_request_hook` — each `router.register_error_handler(Klass, status:, log_level:) { ... }` carries the status and log level inline.
- **Roda `/auth`** encodes the same data declaratively in `Auth::ErrorTranslator::STATUS_BY_CLASS` and `LOG_LEVEL_BY_CLASS`.

`Auth::ErrorTranslator` already documents the coupling: *"The status-code mapping intentionally parallels otto_hooks.rb. If a typed exception's status changes in one place, it should change here as well."* The two are kept in sync by hand; `otto_hooks_spec` and `router_error_shape_spec` each independently assert the values, so a drift turns a spec red — but only after the fact, not structurally.

This is the registry the ADR-013 follow-up note gestures at (*"a future refactor may extract a shared registry consumed by both layers"*). The #3520 work already extracted the **correlation** half of the Otto↔Roda duplication into `Onetime::Application::ErrorCorrelation`; this issue tracks the **status/log-level** half.

## Shared classes (must agree in both layers today)

`RecordNotFound` (404/info), `MissingSecret` (404/info), `FormError` (422/info), `LimitExceeded` (429/warn), `EntitlementRequired` (403/info), `GuestRoutesDisabled` (403/info), `Forbidden` (403/warn), `Unauthorized` (401/warn).

## Proposal

Introduce a single source of truth — e.g. `Onetime::Application::ErrorPolicy` — mapping exception class → `{ status:, log_level: }` for the shared typed errors, consumed by both layers:

- `OttoHooks` reads `status:`/`log_level:` from the registry when registering each handler (the per-error body blocks stay — they differ per class and can't be data-driven).
- `Auth::ErrorTranslator` derives `STATUS_BY_CLASS` / `LOG_LEVEL_BY_CLASS` from the registry instead of hand-maintained literals.

## Constraints / non-goals

- **Preserve each layer's matching semantics.** Otto matches by exact class (subclasses register individually); Roda walks ancestors (`ancestor_status` / `ancestor_level`). The registry is **data only** — matching stays in each consumer, so the change is behavior-preserving.
[...]

PR #3559, opened 2026-06-30
Add request_id correlation to error responses (#3520)

### LINKED ISSUE #3520: Extend request_id error correlation to the Roda /auth surface
## Context

PR #3518 added `request_id` correlation to the **Otto** JSON API error path: `OttoHooks#with_error_correlation` echoes the request's `x-request-id` into every typed error body and stashes `error_type` into the Rack env so `RequestLogger` records *what* failed on the same line that carries `request_id`.

That helper only runs for the Otto error handlers. The **Roda `/auth` surface** renders typed errors through `Auth::ErrorTranslator` and is not covered, so its error responses don't match the same correlation behavior.

ADR-013 was updated in #3518 to document this boundary explicitly (request_id-in-body is currently an Otto-app guarantee; `/auth` relies on the `x-request-id` response header). This issue tracks closing that gap.

## Evidence

From greptile's T-Rex runtime check on #3518: a typed `Onetime::FormError` raised on a Rack-compatible `/auth` route and translated through the real `Auth::ErrorTranslator` under the real `RequestLogger` produced a `422` whose **body had `error_type` but no `request_id`**, while the **request log had `request_id` but not `error_type`**, and `env['otto.error_type']` remained `null`. So neither side of the correlation lands on the auth surface.

## Desired behavior

For typed errors on `/auth`, match the Otto surface:
- Echo `request_id` (from `env['HTTP_X_REQUEST_ID']`) into the JSON error body.
- Stash the `error_type` (class-specific value, or the exception class name as fallback) into `env['otto.error_type']` so `RequestLogger` logs it alongside `request_id`.

## Implementation pointers

- The auth app already runs through the universal `MiddlewareStack` (`Auth::Application < Onetime::Application::Base`; `Base#build_rack_app` → `MiddlewareStack.configure`), so `Rack::RequestId` and `RequestLogger` are active and `env['HTTP_X_REQUEST_ID']` is present.
- Error handling/logging lives in `apps/web/auth/router.rb` (the error handler around lines ~78–112) and `apps/web/auth/error_translator.rb` (currently pure / logging-free — it does not echo `request_id`).
- Mirror `Onetime::Application::OttoHooks#with_error_correlation` (`lib/onetime/application/otto_hooks.rb`): merge `request_id` into the body and set `env['otto.error_type']`. `RequestLogger` already picks up `otto.error_type` and forces `request_id` onto error lines (added in #3518), so no logger change is needed.
- The auth flow exercises `Onetime::Customer` (a `Familia::Horreum`) heavily, so once correlation lands here those auth-triggered DB ops also become tied to the originating request via `request_id`.

## Acceptance

- A typed `/auth` error (e.g. a `422` FormError) returns `request_id` in the JSON body.
- The corresponding `RequestLogger` line carries both `request_id` and `error_type`.
- Tests cover the auth error path (mirroring the Otto coverage added in #3518).

## References

- PR #3518 (Otto-side implementation)
- ADR-013 — API 4xx/5xx Error Response Wire Format (Future Considerations: "Extend `request_id` correlation to the Roda auth surface")

delano:
## Summary

Fixes #3520

Implements request_id correlation for error responses across both the Otto apps and the Roda /auth surface. Errors now echo the `x-request-id` header into the JSON response body and stash the error classification into the request env, allowing RequestLogger to record what failed alongside the request_id on a single log line.

### Changes

**New module: `Onetime::Application::ErrorCorrelation`**
- Centralizes the logic for echoing `request_id` into error bodies and stashing `error_type` into the request env
- Dependency-free design allows use by both Otto hooks and the Roda router without coupling
- Includes fallback to exception class name when body omits `error_type`
- Nil-safe on env for unit test compatibility

**Updated `OttoHooks#with_error_correlation`**
- Refactored to delegate to the shared `ErrorCorrelation` module
- Maintains existing behavior while reducing duplication

**Updated `Auth::Router` error handler**
- Calls `ErrorCorrelation.apply()` on translated error bodies
- Stashes error classification into request env for RequestLogger to read one frame up the middleware stack

**Updated `RequestLogger`**
- Reads error classification from `env['otto.error_type']` (the key written by ErrorCorrelation)
- Ensures request_id is included on error log lines even under `:minimal` capture mode

**Test coverage**
- Added comprehensive unit tests for `ErrorCorrelation` module covering request_id echoing, error_type stashing, fallback behavior, and nil-safety
- Extended auth router error-shape spec with new describe block testing request_id correlation through the shared request env
- Refactored existing router tests to use a shared stub app that mirrors production error-handling wiring

## Related Issues

Fixes #3520

## Test Plan

All new and existing tests pass. The unit tests for `ErrorCorrelation` cover the core logic in isolation. The auth router spec tests verify the integration with Roda's error handler and request env correlation. The existing error-shape assertions remain unaffected by the correlation additions.

https://claude.ai/code/session_01H7ATpkazkyv4XmnrfREhYX


delano:
Thanks for the thorough review. Addressing each point:

**RequestLogger spec coverage** — already covered, so the loop is closed: `spec/unit/onetime/application/request_logger_spec.rb:76–88` exercises `:minimal` capture on an error response and asserts both `error_type` and `request_id` land on the same line. This PR's `request_logger.rb` change swaps the bare `'otto.error_type'` literal for `ErrorCorrelation::ENV_ERROR_TYPE` (identical value), so that existing test drives the exact path — no new case needed. The PR didn't touch that spec, which is why it wasn't in the diff.

**X-Request-ID injection / length** — capped upstream. `Rack::RequestId` only accepts `/\A[\x20-\x7E]{1,200}\z/` (printable ASCII, ≤200 chars) and generates a fresh id otherwise, so a client can't inject arbitrary bytes or unbounded length into the echoed body.

**Docstring notes (anonymous-class `nil` fallback + `respond_to?` vs `req&.env` rationale)** — folded in as `21cfdee` (doc-only).

**`short_class_name` duplication** — intentional for now: it's a pure one-liner, and deduping would couple `Auth::ErrorTranslator` (auth app) to `ErrorCorrelation` (lib) solely for it. It folds naturally into the shared exception→`{status, log_level}` registry tracked in #3557.

**404 fallbacks bypassing correlation** — intentional and symmetric with Otto: the `status_handler(404)`/catch-all bodies are request-independent (returned from routing, not raised), mirroring Otto's static `router.not_found`/`router.server_error` fallbacks per ADR-013 — the `x-request-id` response header remains their correlation handle.

**`payload[:request_id] ||= …`** — agreed, a no-op in practice since `nil` and an absent key are equivalent to `||=` here.

---

Issue #3558, opened 2026-06-30
CSP header not emitted on web surface despite nonce machinery


delano:
## Problem Statement

The web application generates CSP nonces and stamps them on every `<script>` tag, but no CSP header is emitted to enforce them. This leaves the nonce-based XSS protection inert. The hardened CSP policy from #3498 (commit 40431511d9) only applies to the opt-in API surface, not the user-facing HTML pages where inline scripts actually execute.

## Early Diagnosis

Two disjoint CSP mechanisms exist:

| Surface | File | Mechanism | Gate | Emits Header? |
|---------|------|-----------|------|---------------|
| API | `apps/api/v1/controllers/helpers.rb:163` | `res.headers['content-security-policy'] = …` (line 212) | opt-in (`site.security.csp.enabled == true`, line 171) | **yes** |
| Web | `apps/web/core/application.rb:121` | `router.enable_csp_with_nonce!` | always-on flag | **no** |

The web path (`enable_csp_with_nonce!`) only sets `@csp_nonce_enabled = true` (otto `config.rb:364`). The nonce is generated (`apps/web/core/middleware/request_setup.rb:37`, `env['onetime.nonce']`) and applied to script tags (`vite_manifest.rb`, `sanitizer.rb`), but nothing calls Otto's `Response#send_csp_headers` (otto `response.rb:123`) — the header never materializes.

#3498 strengthened the API helper's CSP, but the web rendering path was not touched.

## Affected Areas

- [ ] `apps/web/core/application.rb` — `enable_csp_with_nonce!` call exists but no header emission wired
- [ ] `apps/web/core/middleware/request_setup.rb` — nonce generation (line 37)
- [ ] `apps/api/v1/controllers/helpers.rb` — working CSP emission gated by opt-in flag (line 171+)
- [ ] otto `lib/otto/response.rb` — `send_csp_headers` defined but uncalled
- [ ] otto `lib/otto/config.rb` — `default_security_headers` (line 630) excludes CSP

PR #3561, opened 2026-07-01
refactor(security): non-CSP #3498 hardening — YAML safe-load Date/Time + CLI regression coverage

delano:
## Summary

[#3498](https://github.com/onetimesecret/onetimesecret/issues/3498)

The #3498 coordination branch (PR #3544) bundled the CSP work together with two unrelated hardening fixes. This PR carves out **only the non-CSP fixes** so they can be reviewed and merged on their own; the CSP work is handled separately (report-endpoint in #3560, broader rollout in #3558, being reworked from the Otto layer up). The net diff against `main` is exactly 8 files.

**YAML safe-load now permits `Date` and `Time`.** Four loaders passed `permitted_classes: [Symbol]`, so any config carrying a timestamp scalar raised `Psych::DisallowedClass` at load. They now pass `[Symbol, Date, Time]` (with `require 'date'`) — timestamp-bearing config loads, while arbitrary object deserialization stays rejected (no `!ruby/object`, etc.):
- `lib/onetime/config.rb`
- `lib/onetime/operations/config/validate.rb`
- `lib/onetime/utils/enumerables.rb`
- `lib/onetime/initializers/setup_loggers.rb`

**CLI security regression coverage** for the two already-merged (44f55d3) hardening items that shipped without dedicated regression tests:
- `spec/cli/session_command_security_spec.rb` — item 1: session data is never `Marshal.load`-ed from Redis bytes.
- `spec/cli/billing/test_trigger_webhook_command_spec.rb` — item 3: the Stripe CLI is invoked via array-form `system(*cmd)`, so a shell-metacharacter argument is passed literally, never through a shell.

## Related Issues

Part of #3498 (does not close it). Overlaps with PR #3544, which currently contains a superset (CSP + these fixes) — see note below on de-duplication.

## Test Plan

- `spec/unit/onetime/config_yaml_safe_load_spec.rb` — asserts the loaders accept `Symbol`/`Date`/`Time`/aliases and still reject `!ruby/object` payloads (`Psych::DisallowedClass`).
- `spec/cli/session_command_security_spec.rb`, `spec/cli/billing/test_trigger_webhook_command_spec.rb` — regression specs for items 1 and 3.

Verification: 44 examples, 0 failures; `rubocop --fail-level warning` clean on the four changed production files; no `Gemfile`/`Gemfile.lock` changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_017trt2xM4hEQr3zLqddcEeV)_


delano:
Fixed in 10a5f144f.

**Item #1 — one correction on the mechanism.** A bare `Date` (`2026-01-02`) does *not* silently become a long TTL: `Date#to_i` raises `NoMethodError`, so that path already fails loud (the same loud failure as the pre-#3498 `Psych::DisallowedClass` crash). The regression this PR actually introduced is a bare **`Time`** (`2026-01-02 10:00:00`): `Time#to_i` returns epoch seconds ≈ 56 years, **silently**.

The fix adds `coerce_ttl_seconds` at the config-normalization boundary (`after_load`), rejecting any `Date`/`Time` on the two numeric TTL fields (`site.secret_options.default_ttl`, `features.incoming.default_ttl`) with an actionable `OT::ConfigError`. `String`/`Float` stay on lenient `.to_i` so live deployments aren't regressed. This restores the validate⇒boot symmetry the loader comment promises — the JSON schema already types these `integer`, so a config that validates now also boots.

**Item #2 — fixed.** Added `require 'stripe'` before the spec stubs `Stripe.api_key`. To be precise, this is explicitness/robustness rather than a live bug: the full-app boot currently pulls `stripe` in transitively, so the stub target happens to resolve today — but the spec shouldn't lean on incidental boot order. Verified the spec passes run in isolation.

+6 regression examples (Date/Time rejected; Integer/Float/String still coerce). 40 examples, 0 failures.

---

Separately, this surfaced a structural gap I'll track as a **stacked follow-up** rather than widen this PR: boot never runs the schema validator, so the `after_load` coercions and `static.schema.json` are two independent type systems that can drift. The same class also turns `default_ttl: "7d"` into `7` seconds. Closing that (schema-validate-at-boot, or generate the coercions from the schema) is its own change.

PR #3658, opened 2026-07-05
Refactor scope switchers to use shared ScopeSwitcher engine

delano:
## Summary

Extract the common dropdown interaction logic from `DomainContextSwitcher` and `OrganizationScopeSwitcher` into a reusable `ScopeSwitcher` engine component. This eliminates code duplication and centralizes the fix for a subtle but critical bug: the gear icon's `stopPropagation()` must be paired with an explicit `close()` call to dismiss the menu, since `stopPropagation()` suppresses HeadlessUI's built-in MenuItem auto-close.

### Changes

**New shared engine:**
- `src/shared/components/navigation/ScopeSwitcher.vue` — featurized HeadlessUI Menu dropdown that owns trigger styling, panel layout, row rendering, and the interaction contract (select/open-settings events emitted *after* menu closes)
- `src/shared/components/navigation/scopeSwitcher.ts` — `ScopeSwitcherItem` interface defining the normalized row shape

**Refactored adapters:**
- `src/apps/workspace/components/navigation/DomainContextSwitcher.vue` — moved from `src/shared/components/navigation/`, now a thin adapter that maps domains to `ScopeSwitcherItem[]` and handles domain-specific navigation
- `src/apps/workspace/components/navigation/OrganizationScopeSwitcher.vue` — refactored to use the engine, eliminating ~100 lines of duplicated dropdown/interaction code

**Tests:**
- `src/tests/shared/components/navigation/ScopeSwitcher.spec.ts` — unit tests for the engine in isolation (real HeadlessUI, stubbed OIcon)
- Moved domain switcher tests from `src/tests/shared/` to `src/tests/apps/workspace/` to reflect the new location
- Updated import paths in existing tests

### Why

The two switchers had nearly identical dropdown logic with a known bug: clicking the gear icon would fail to close the menu because `stopPropagation()` suppresses HeadlessUI's auto-close. Both components had to manually call `close()` to work around this, but the pattern was easy to miss or regress. By centralizing the interaction in one place, we:

1. Fix the bug once and prevent future regressions
2. Reduce maintenance burden (one source of truth for dropdown behavior)
3. Make it easier to add new scope switchers in the future

### Scope: behavior-preserving

This is a refactor, not a behavior change. The two real-HeadlessUI `*.close.spec` suites — the regression net for the original "dropdown stays open after navigation" bug — pass **unchanged** against the rewritten adapters, which is the evidence that behavior is preserved. Two intentional, benign DOM deltas: the organization dropdown header moved from a plain `<div>` to a flex container + `<span>` (visually identical; needed so the shared header can host the domain switcher's `[+]` action), and org rows without an `extid` now use test id `org-menu-item-<objid>` instead of the previous `org-menu-item-undefined`.

### Deliberately out of scope (follow-ups)

- **Nested `<button>` accessibility.** The row is a `<button>` containing the gear `<button>` — invalid HTML, flagged on the earlier fix PR. This refactor **does not** fix it; it makes it a **one-place** change instead of two. The proper fix restructures the row so the gear is a sibling of the clickable row element, which is a structural (not behavior-preserving) change best done on its own.
- **Navigation dedup.** The `onOrgSwitch` / `onDomainSwitch` route-meta resolution (`'same'` / `:extid` / static path) is still duplicated across the two adapters. It is the most-divergent piece (named-route push vs. path-string replace), so a `useScopeSwitchNavigation()` extraction is left as a phase-2 step.

## Test Plan

- Unit tests for `ScopeSwitcher` engine verify the interaction contract: `select` and `open-settings` are emitted after the menu closes, disabled rows don't select, and the gear button's `stopPropagation()` + `close()` dance works correctly
- Existing domain and organization switcher tests pass with updated imports
- E2E tests for domain context switching continue to pass

https://claude.ai/code/session_01WxiCsDnAFVC6w6xvfvFZce


delano:
Thanks — both are fair points on logic that's now centralized. Addressed in f16ecb3.

**1. `String.replace()` first-occurrence only (`DomainContextSwitcher.vue`)** — switched the placeholder substitutions to `replaceAll` for `:extid` and `:orgid` (both the `matched.path` and the `switchTarget` branches). No observable change for current routes, which carry at most one of each placeholder, but it can no longer leave a stray `:extid`/`:orgid` segment if a route pattern ever repeats one.

**2. Silent no-op when `route.name` is falsy (`OrganizationScopeSwitcher.vue`)** — the `'same'`-switch branch now `console.warn`s on an anonymous route instead of doing nothing, matching the existing missing-`extid` branches.

Both are behavior-neutral for every existing route — the real-HeadlessUI close-behaviour suites assert the exact navigation targets and still pass (25/25 across the four switcher suites), `vue-tsc` clean, ESLint 0 errors.

---

Issue #3625, opened 2026-07-04
Receipt state transitions can resurrect a TTL-evicted receipt with no expiration (immortal receipt)


delano:
## Summary

The Receipt state transitions (`previewed!`, `revealed!`, `burned!`, `orphaned!`, `expired!` in `lib/onetime/models/receipt/features/deprecated_fields.rb`) persist via `save update_expiration: false`. If the receipt's Redis key has been TTL-evicted (expired) between the time the object was loaded and the time a transition runs, that save **recreates the hash with no TTL** — an immortal receipt — and re-registers it in the `receipt:instances` index.

This is the Receipt analogue of the Secret `previewed!` resurrection bug fixed in PR #3615: a non-conditional write that recreates a key which is supposed to be gone. Severity is lower than the Secret fix — receipts hold no plaintext — but it leaks permanent keys and lets a stale in-memory instance revert an already-advanced persisted state.

## Root cause

Reading the familia 2.10.1 gem source:

- `save(update_expiration: true)` (the default) → `persist_to_storage(true)` → `hmset(...)` **then** `update_expiration`, which issues `EXPIRE default_expiration`. Correct.
- The receipt transitions call **`save update_expiration: false`** (deliberately, so viewing a receipt doesn't reset its TTL). On that path `persist_to_storage(false)` runs `hmset(...)` and **skips `update_expiration` entirely** (`persistence.rb:935-950`).

`hmset` on a missing key **creates** it, and with the `EXPIRE` step suppressed the recreated hash has no TTL. `hmset` can't distinguish "recreating an evicted key" from "updating a live one," and the caller explicitly opted out of touching the TTL — so the combination yields an immortal key. This is not a bug in `save`/`update_expiration` themselves; it's the interaction of the non-conditional write with `update_expiration: false` on the resurrection path.

## Reproduction

```
fresh receipt        state="new"  ttl=7200          (positive TTL, as expected)
evict the hash (DEL) EXISTS=0     ttl=-2            (key gone, as if TTL-expired)
call previewed!      EXISTS=1     ttl=-1            (key back, NO expiry)
VERDICT: immortal receipt (recreated with no TTL)
```

The transition's in-memory `state?(:new)` fast-path passes (the loaded instance still believes it's `:new`), so the guarded `save` runs against the now-absent key.

## Impact

- **Immortal keys**: an evicted receipt is resurrected without a TTL and never expires, and is re-added to the `receipt:instances` zset — slow leak of permanent keys.
[...]

Issue #3633, opened 2026-07-04
A GET on the secret status/show endpoints mutates secret state (safe-method surprise + weak access signal)


delano:
## Summary

Reading a secret's status changes the secret. `GET /secret/:identifier/status` (and `GET /secret/:identifier`) advance the secret's persisted lifecycle `state` from `new` → `previewed` as a side effect of the read. This is surprising on two counts — it violates the HTTP "safe method" expectation for GET, and it's a Command/Query Separation break — and the signal it produces for the person who created the secret is weaker than it looks.

This ticket is about the **problem space**. Concrete solutions are sketched at the end as *options* to discuss, not a chosen fix.

## Where it happens

- Route: `GET /secret/:identifier/status` → `V2::Logic::Secrets::ShowSecretStatus` (`apps/api/v2/routes.txt:22`)
- Mutation: `apps/api/v2/logic/secrets/show_secret_status.rb:35` calls `secret.previewed!` inside `process`
- Same pattern on the read path: `apps/api/v2/logic/secrets/show_secret.rb:97` (`GET /secret/:identifier`)
- Transition itself: `Secret#previewed!` in `lib/onetime/models/secret/features/secret_state_management.rb` (CAS `new` → `previewed`)
- Observable in output: `state`, `is_previewed`, `is_viewed` are all in the secret's safe-dump (`lib/onetime/models/secret/features/safe_dump_fields.rb`)

The `ShowSecretStatus` docstring describes it as checking existence and state *"without consuming it."* That's true in the narrow sense (it doesn't burn/reveal), but it does mutate lifecycle state, which a reader of "show status" would not expect.

## Why it's a problem

**1. Least-surprise / HTTP semantics.** RFC 9110 defines GET as a safe method — clients, crawlers, and intermediaries assume a GET has no meaningful server-side effect on the resource. A `/status` read advancing a state machine breaks that assumption, and the endpoint's own name/docstring reinforce the wrong mental model.

**2. The creator-facing signal is noisy.** Because the transition fires on any GET, it cannot distinguish a human recipient from mechanical fetches: link-preview unfurlers (Slack/Teams/iMessage), email security scanners, browser prefetch, uptime monitors. The creator sees "viewed" when nobody may have actually opened it — a false positive on the exact "was my secret seen?" signal a secret-sharing product is trusted for.

**3. The signal is also lossy / inconsistent.**
[...]

PR #3674, opened 2026-07-06
fix(receipt): guard state transitions with atomic CAS to stop TTL-evicted resurrection (#3625)

delano:
Receipt state transitions (revealed!, orphaned!, burned!, expired!)
persisted via `save update_expiration: false`, which issues an
unconditional HMSET. HMSET on a missing key CREATES it, and with the
EXPIRE step suppressed the recreated hash had no TTL. If a receipt's
Redis key was TTL-evicted between load and transition, the in-memory
`state?(:new)` fast-path still passed and the save resurrected the hash
as an immortal (TTL-less) key, re-registering it in receipt:instances.
A stale instance could likewise revert an already-advanced state.

Gate each transition on an atomic compare-and-set on the persisted
`state` field, mirroring the Secret STATE_CAS_SCRIPT from #3615. The
Lua HGET matches nothing on a missing key, so the claim fails closed:
an evicted receipt is never recreated and a terminal state is never
reverted. Only the winner persists the accompanying fields; its HSET
lands on the live key, preserving the original TTL.

Add regression tryouts reproducing the eviction window and asserting
the fail-closed outcome (no resurrection, no reversion, single winner
under concurrency).

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01MaKi63TWxNMcTKS9x7JMrN

### DIFF
--- a/apps/api/v2/spec/logic/secrets/state_backward_compat_spec.rb
+++ b/apps/api/v2/spec/logic/secrets/state_backward_compat_spec.rb
@@ -60,8 +60,11 @@ RSpec.describe 'V2 API State Backward Compatibility' do
       end
     end
 
-    context 'when previewed (new internal state)' do
-      before { receipt.previewed! }
+    context 'when previewed (telemetry-derived, #3633)' do
+      # previewed! was retired: viewing a link no longer mutates lifecycle
+      # state. A receipt now reads as "previewed" once its link has been
+      # fetched, i.e. once an access-timeline event exists.
+      before { receipt.record_access_event('secret_get') }
 
       it 'viewed timestamp is populated' do
         dump = receipt.safe_dump
@@ -117,7 +120,7 @@ RSpec.describe 'V2 API State Backward Compatibility' do
 
     context 'when previewed then revealed' do
       before do
-        receipt.previewed!
+        receipt.record_access_event('secret_get')
         @viewed_timestamp = receipt.safe_dump[:viewed]
         receipt.revealed!
       end
@@ -151,22 +154,10 @@ RSpec.describe 'V2 API State Backward Compatibility' do
       receipt.destroy! if receipt.exists?
     end
 
-    describe '#previewed!' do
-      it 'sets state and timestamp' do
-        receipt.previewed!
-        expect(receipt.state).to eq('previewed').or eq('viewed')
-        # Check the new canonical field (safe_dump provides backward compat for :viewed)
-        expect(receipt.previewed.to_i).to be > 0
-      end
-
-      it 'only transitions from :new state' do
-        receipt.state = 'burned'
-        receipt.save
-        receipt.previewed!
-        # Should not change since guard prevents it
-        expect(receipt.state).to eq('burned')
-      end
-    end
+    # #previewed! was retired in #3633: viewing a secret link no longer mutates
+    # lifecycle state. The telemetry-derived "previewed" signal (is_viewed /
+    # viewed timestamp) is covered by the safe_dump contexts above; the
+    # receipt-page-view audit event is covered in organization_audit_trail_spec.
 
     describe '#revealed!' do
       it 'sets state and timestamp' do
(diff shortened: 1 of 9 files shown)


delano:
Thanks for the review. Pushed `4171859` in response.

### Review suggestions

**1. Add an `expired!` eviction test — done.** The three sibling eviction cases skipped `expired!` because it carries the extra `secret_expired?` time gate. New case back-dates the in-memory `created` stamp so `secret_expiration` lands in the past, and asserts `secret_expired?` is `true` *before* the transition — so the case provably reaches and exercises the CAS rather than short-circuiting on the guard. (Note: `spawn_pair` rejects a negative lifespan — `Default expiration must be non-negative` — so the eviction is forced via the timestamp rather than a negative TTL.) Receipt tryout now 9/9, Secret 10/10.

**2. Hoist `STATE_CAS_SCRIPT` to the module level — done.** It now lives on `StateCas` and is reachable as `StateCas::STATE_CAS_SCRIPT`, giving a future `SCRIPT LOAD`/`EVALSHA` path one canonical source to hash. Instance methods still resolve it via lexical nesting, and both models continue to own `compare_and_set_state!` solely through `StateCas::InstanceMethods` (verified at runtime).

**3. Align the losing-caller in-memory state in the receipt transitions — deferred, intentionally.** In `Secret#win_reveal_claim!` the loser sets `@state`/`@ciphertext` because `viewable?`/`safe_dump` are read *after* a lost claim to withhold the plaintext — a security-relevant read. The receipt transitions have no equivalent post-loss reader: a losing caller just returns falsy and the object isn't inspected further. Adding the mutation would be a behavior change (a stale instance would suddenly report the terminal state) with no call site that needs it, which is outside this PR's fail-closed-persistence scope. Happy to align it in a follow-up if a reader that benefits turns up. The other notes (the `orphaned!` `secret_identifier` business-rule guard, the layered `secret_expired?` check, `eval` letting Redis errors bubble) all read as "confirmed correct, just noting" and I've left them as-is.

### CI: `ruby-unit` failure is pre-existing and unrelated to this PR

The 13 failures are all `undefined method 'previewed!'` in `state_backward_compat_spec.rb` and `organization_audit_trail_spec.rb`. `previewed!` was **retired** in #3633 (base commit `260be4a` of this branch) — `access_timeline.rb` even documents it as "the retired `previewed!` state mutation." Those two spec files still call `previewed!` (11 references already present at `260be4a`, where the method is likewise undefined), so they fail on `main` exactly as they do here. This PR's diff touches neither spec and neither defines nor removes `previewed!`:

```
docs/architecture/encryption-at-rest-dpa-audit.md
lib/onetime/models/features/state_cas.rb          (new)
lib/onetime/models/receipt.rb                      (+1: feature :state_cas)
lib/onetime/models/receipt/features/deprecated_fields.rb
lib/onetime/models/secret.rb                       (+1: feature :state_cas)
lib/onetime/models/secret/features/secret_state_management.rb
try/unit/models/receipt_ttl_resurrection_race_try.rb
```

So the red check isn't something this PR introduced or can fix on its own — it's stale-spec cleanup left over from the #3633 retirement. I'd rather not fold that into a receipt-CAS PR, but I'm glad to send a separate PR that deletes/rewrites the `previewed!` specs if you'd like the tree green.

---

delano:
Went ahead and fixed the pre-existing `previewed!` unit failures here rather than in a separate PR (`f35b155`). They were the 13 `NoMethodError: undefined method 'previewed!'` in `state_backward_compat_spec.rb` and `organization_audit_trail_spec.rb` — stale calls to the state mutation #3633 retired. Realigned them with the current lifecycle/telemetry split:

- **Telemetry "previewed"** — the safe_dump backward-compat contexts now drive `record_access_event('secret_get')` (which is what sets `is_viewed` / the `viewed` timestamp today) instead of `previewed!`.
- **Receipt-page-view audit** — uses `record_receipt_view!`, whose claim-once semantics match the "exactly one `receipt_viewed`" assertion (the test's own comment already described that behavior).
- **Reveal from `:previewed`** — constructs the legacy `:previewed` state directly, since the CAS still accepts it as a from-state for legacy receipts even though no method advances into it anymore.
- **Dropped** the obsolete `#previewed!` method describes on Receipt and Secret, with a pointer to where the replacement behavior is covered.

Both files now pass (43 examples, 0 failures). The v1 `previewed!` references are RSpec doubles/stubs (`double(…, previewed!: nil)`, `not_to receive(:previewed!)`), not real calls, so they were already green and I left them alone. That clears the only 13 failures the `unit / simple` run reported.

---

PR #3676, opened 2026-07-07
Add configurable secret content size limit

delano:
## Summary

Implement a server-enforced and configurable maximum length limit for secret content, shared across both regular (V2) and incoming (V3) secret creation paths. This limit is now the single source of truth for the client-side textarea hint, preventing drift between server and client validation.

## Changes

- **Server validation**: Added `validate_secret_size()` method in `lib/onetime/logic/base.rb` to enforce the configured maximum length on secret creation
- **Configuration**: Added `site.secret_options.content.maximum_length` (default: 10,000 characters) to `config.defaults.yaml` with environment variable override support (`SECRET_MAX_LENGTH`)
- **Incoming API**: Updated `GetConfig` response to include `secret_max_length` from the global configuration, exposed via `RecipientResolver`
- **V3 CreateIncomingSecret**: Added validation call to reject secrets exceeding the configured limit
- **V2 ConcealSecret**: Added validation call for consistency across both paths
- **Frontend**: Updated Vue components and composables to use the dynamic `secretMaxLength` from the incoming store instead of hardcoded 10,000
- **TypeScript schemas**: Added `content` object to `publicSecretOptionsSchema` with `maximum_length` field and corresponding shape definitions
- **Tests**: Added comprehensive test coverage for validation behavior (rejection above limit, acceptance at limit) and store integration

## Test Plan

Automated tests added covering:
- `GetConfig` includes `secret_max_length` from configuration
- `CreateIncomingSecret` rejects content exceeding the configured maximum
- `CreateIncomingSecret` accepts content exactly at the configured maximum
- Frontend store correctly reflects `secret_max_length` from API response
- Frontend components use dynamic limit instead of hardcoded value

Existing test suite passes with new validation integrated into both secret creation paths.

https://claude.ai/code/session_01TiUiRTLFSxARgAk7BwSig6

### DIFF
--- a/apps/api/incoming/logic/create_incoming_secret.rb
+++ b/apps/api/incoming/logic/create_incoming_secret.rb
@@ -77,6 +77,7 @@ module Incoming
 
         # Validate required fields (memo is optional)
         raise_form_error 'Secret content is required' if secret_value.empty?
+        validate_secret_size(secret_value)
         raise_form_error 'Recipient is required' if @recipient_hash.to_s.empty?
 
         # Now safe to perform resolver I/O (entitlement verified)
--- a/apps/api/incoming/logic/get_config.rb
+++ b/apps/api/incoming/logic/get_config.rb
@@ -17,6 +17,7 @@ module Incoming
     #     config: {
     #       enabled: true,
     #       memo_max_length: 50,
+    #       secret_max_length: 10000,
     #       default_ttl: 604800,
     #       recipients: [
     #         { hash: "abc123...", name: "Support Team" }
--- a/apps/api/v2/logic/secrets/conceal_secret.rb
+++ b/apps/api/v2/logic/secrets/conceal_secret.rb
@@ -27,6 +27,7 @@ module V2::Logic
         require_guest_route_enabled!(:conceal)
         super
         raise_form_error 'You did not provide anything to share', field: :secret, error_type: :missing if secret_value.to_s.empty?
+        validate_secret_size(secret_value)
       end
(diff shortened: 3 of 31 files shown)


delano:
Pushed `d741024`.

**Copilot review (Integer coercion) — addressed.** All three comments were the same underlying issue: `maximum_length` was normalized only when the config value was a `String`, so an unquoted YAML/ERB scalar like `10000.0` (a `Float`) would flow through and violate the frontend's Zod `int()` contract / render as `"10000.0"` in the error. Fixed at all three sites:
- `config.rb` `after_load`: coerces any non-`Integer` (String **or** Float) → Integer.
- `RecipientResolver#secret_max_length`: `to_i` guard before returning.
- `Onetime::Logic::Base#validate_secret_size`: normalizes once so the comparison and the error message agree.
- Added a tryout asserting `secret_max_length` is `Integer` even when config holds a `Float`.

**Failing `CI` check is pre-existing and unrelated to this PR.** The 13 failures are all `NoMethodError: undefined method 'previewed!'` in `apps/api/v2/spec/logic/secrets/state_backward_compat_spec.rb` and `apps/api/v2/spec/models/organization_audit_trail_spec.rb`. `previewed!` was retired by the receipt-state work already in this branch's base (`260be4a`), but those two spec files still call it. This PR touches none of those files, and the tryouts step (which runs first under `set -e`, including the new size-limit boundary cases) passed. Happy to fix those two specs in a separate PR if useful, but it's out of scope here.

---

PR #3686, opened 2026-07-08
Update AddDomain flow to helpfully suggest alternatives to an apex domain

delano:
## Summary

Redesign the domain form submission flow to provide guided UX for apex (registrable) domains. When a user enters a bare domain like `acme.com`, they now choose where secret links should live (via subdomain options or root domain). Deeper hostnames like `secrets.acme.com` are confirmed and used verbatim.

### Key Changes

- **New `analyzeDomain()` utility** (`src/utils/parse/domain.ts`): Pure, dependency-free hostname parser that classifies input as empty, valid subdomain, valid apex, or invalid. Handles multi-part suffixes (e.g., `co.uk`) and normalizes scheme/case/path. Server remains authoritative via `addDomainRequestSchema`.

- **Refactored DomainForm component**: 
  - Replaced simple validation with guided UX: shows echo block for subdomains, address-choice cards for apex domains
  - Added step rail (Add › Verify › Brand) to contextualize the form in the larger flow
  - Errors only surface after a submit attempt, clearing on keystroke
  - Submit button disabled until an apex choice is made; label updates dynamically
  - Removed `ErrorDisplay` component in favor of inline error paragraphs

- **Updated DomainInput**: Added `describedby` prop to link help/error text; parent now owns the label for proper accessibility

- **New i18n keys**: Field label, help text, echo confirmation, apex question, subdomain/root options, error messages, step labels, and dynamic continue button text

- **Comprehensive test suite**: 
  - `domain.spec.ts`: 147 tests covering empty input, subdomains, apex domains, invalid suffixes, malformed input, and cleaning logic
  - `DomainForm.spec.ts`: Refactored from 20 to 30+ tests covering rendering, empty/invalid submission, subdomain echo path, and apex address-card selection

## Test Plan

All new and refactored tests pass:
- `pnpm exec vitest run src/tests/utils/parse/domain.spec.ts` — 147 tests for hostname parsing
- `pnpm exec vitest run src/tests/apps/workspace/components/domains/DomainForm.spec.ts` — 30+ tests for form rendering, validation, and submission flows

The form's emit contract remains unchanged (`emit('submit', hostname)` and `emit('back')`), ensuring backward compatibility with parent components.

https://claude.ai/code/session_01JK58qMAP8wbV3gq8o5FbJC

### DIFF
--- a/lib/onetime/config.rb
+++ b/lib/onetime/config.rb
@@ -482,7 +482,7 @@ module Onetime
       # an Integer so the value the frontend receives satisfies its int()
       # contract and never renders as "10000.0".
       content_config = conf.dig('site', 'secret_options', 'content') || {}
-      content_max = content_config['maximum_length']
+      content_max    = content_config['maximum_length']
       unless content_max.nil? || content_max.is_a?(Integer)
         conf['site']['secret_options']['content']['maximum_length'] = content_max.to_i
       end
@@ -650,6 +650,19 @@ module Onetime
         candidate  = legacy_brand_value(ENV.fetch(legacy[:env], nil)) ||
                      legacy_brand_value(dig_path(conf, legacy[:path]))
         brand[key] = candidate if candidate
+      end
+
+      # A bare-relative logo path (e.g. 'img/logo.svg') resolves against the
+      # browser's current-route directory, so it loads on '/' but 404s on a
+      # nested route like '/receipt/:id' (the img then renders its alt text
+      # instead). Root-relativize it here so the one install logo resolves
+      # identically on every surface. Absolute URLs (scheme: or protocol-
+      # relative //) and already-root-relative paths pass through untouched.
+      logo_path = brand['logo_url']
+      if logo_path.is_a?(String) && !logo_path.empty? &&
+         !logo_path.start_with?('/') &&
+         !logo_path.match?(/\A[a-z][a-z0-9+.-]*:/i)
+        brand['logo_url'] = "/#{logo_path}"
       end
 
       # brand.logo_url is now the one install logo for every surface, but the
--- a/locales/content/en/workspace-domains.json
+++ b/locales/content/en/workspace-domains.json
@@ -123,6 +123,62 @@
     "text": "Add domain",
     "content_hash": "d86476ae"
   },
+  "web.domains.add.field_label": {
+    "text": "Custom domain",
+    "content_hash": "d7d4c1e4"
+  },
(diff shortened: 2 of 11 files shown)


delano:
Thanks for the thorough reviews (@copilot + the Claude review bot). Addressed in `3b333a7`. Summary:

**Correctness / schema alignment**
- **Underscores** — `analyzeDomain`'s `okChars` now permits interior underscores, matching `addDomainRequestSchema` (`[a-zA-Z0-9-_.]`). The parser gates submission, so it must not be stricter than the schema. New test covers `my_app.acme.com`.
- **Punycode / non-alpha TLDs** — the TLD check was `/^[a-z]{2,}$/`, which rejected `xn--p1ai`. Now `last.length >= 2 && /[a-z]/.test(last)`: accepts punycode and any letter-bearing gTLD, still rejects a single-char TLD (`example.c`) and a purely-numeric last label (`acme.123`). New tests for both.

**Accessibility**
- The inline error now has `id="domain-error"`, and the input's `aria-describedby` appends it (`"domain-help domain-error"`) only while an error is shown — otherwise just `"domain-help"`. Tests assert both.
- Step-rail `<nav>` now uses a dedicated `step_rail_label` ("Domain setup steps") instead of reusing "Add".

**i18n**
- Empty-input error switched from `please_enter_a_domain_name` with `['']` (rendered a double space) to a dedicated placeholder-free key `web.domains.add.error_empty`.

**DomainInput**
- `autofocus` is now bound to the real `<input>` via a template ref + `onMounted` focus (the attribute previously fell onto the wrapper `<div>` and did nothing in the SPA). The non-functional `required` passthrough was removed since custom validation owns the empty state.
- Dev-only `console.error` added to the previously-silent schema `catch`.

**Tests / docs**
- Added `www.acme.com` → apex → `root` submit coverage (emits bare `acme.com`) and a `full`-preservation assertion; fixed the stale vitest path in a doc comment.

**Intentionally left as-is:**
- *Shared `data-testid="domain-error"` on both error `<p>`s* — the two are mutually exclusive (`v-if`/`v-else-if`), so only one is ever in the DOM; a single id/testid is unambiguous at runtime.
- *`adding_ellipses` "double ellipsis"* — the key value is `"Adding"` (no trailing `…`), so `{{ ... }}...` renders `"Adding..."` as intended.
- *`choice` v-model typing* — the radio `value`s are hardcoded literals of the `Choice` union, so the runtime value is always in range and it typechecks; a cast would be cosmetic.

---

PR #3711, opened 2026-07-11
ci(C7): fresh-clone contributor lane + docs-command drift guard

delano:
## What

First slice of **C7 — Clean-room harness + CI lanes** (install-onboarding [testing-strategy §2c](docs/specs/install-onboarding/install-onboarding-testing-strategy.md), [work-chunks C7](docs/specs/install-onboarding/install-onboarding-work-chunks.md)). Two CI lanes + one committed harness script.

### `fresh-clone.yml` — the Zulip shape (2c)
A clean `ubuntu-24.04` runner runs *nothing but the documented contributor path*, from zero:

```
./install-test.sh                       # the documented setup artifact
pnpm run test:rspec:fast && pnpm test   # documented suites
./install-test.sh                       # second run = idempotency contract
git status --porcelain == empty         # litter check
```

- Toolchain comes from the pinned version **files** (`.ruby-version` 3.4.9, `.node-version` 22) to prove the *documented floors* — the Node-22 gate fix is the whole point of the onboarding branch, so this lane deliberately does **not** use `setup-node-env` (defaults to 25 + force-installs deps).
- No config-seeding composites, no Valkey service container: `install-test.sh` does all app-level setup itself. Reproducing that in YAML would test an idealized path, not the artifact contributors run.
- No `pnpm run build` before the suite — TR-01 is that `spec:fast` is green on a pristine clone with no build.
- Job duration is surfaced as the TTFHW proxy.

**This lane is the end-to-end clean-room re-run of the C1–C3 fixes (NF-1…NF-5) on a fresh Ubuntu host** — the item the clean-room validation doc flagged as outstanding.

### `docs-command-drift.yml` + `scripts/test-install/check-docs-commands.sh`
A 10-line guard that fails when a documented command references a script, `bin/` entrypoint, or `pnpm run` target that no longer exists. Same artifact runs locally and in CI. Passes green locally today.

## Not in this PR (later C7 slices)
- `proof-of-life.sh` (shared boot assertion)
- compose-smoke lane + app compose healthcheck — 2a (pulls forward CP-11)
- installer-matrix lane + `run.sh` container driver — 2b

## Verification
YAML validated, shellcheck clean, drift guard green locally. The fresh-clone lane can only be proven green on a real runner — watching this PR's Actions is the point.

PR #3712, opened 2026-07-11
ci(C7 2a): compose-smoke lane + app healthcheck (CP-11)

delano:
## What

C7 slice **2a** (install-onboarding [testing-strategy §2a](docs/specs/install-onboarding/install-onboarding-testing-strategy.md)) — guards the #1 self-host funnel.

> **Stacked on #3711** (uses `scripts/test-install/proof-of-life.sh` from that PR). Base is `feature/c7-clean-room`; merge #3711 first, or this retargets to `integration/onboarding` after.

### App healthcheck (CP-11) — `docker-compose.{simple,full}.yml`
The `app` service had no compose `healthcheck:`, so `docker compose up --wait` couldn't use it as a readiness gate. Added one reusing the image's role-aware `bin/healthcheck.sh` (web role → `/health/advanced` then `/health`), tuned for faster `--wait` polling. Upgraded full's proxy `depends_on: app` to `service_healthy`.

### `compose-smoke.yml`
- **`compose-smoke`** (PR on `docker/**` + weekly cron): `config -q` on every documented combo (root / simple / full / simple+mailpit), then `up` the simple stack with `--wait` (the healthcheck is the assertion) + shared `proof-of-life.sh`. Includes a CP-5 data-uid (1001) workaround — the real fix is a named volume/chown in C5.
- **`docker-run-readme`** (cron + manual, not per-PR): the README `docker run` quick start run **verbatim** against the exact pinned tag *scraped from README* so it can't drift from the docs.

## Verification
YAML valid, `config -q` parses clean client-side. The `up --wait` + proof-of-life path can only be proven on a runner with Docker — watching this PR's Actions is the point. Known risk being tested: whether the published `:latest` image goes healthy under the simple stack in CI (data perms / boot config).

### DIFF
--- a/.github/workflows/compose-smoke.yml
+++ b/.github/workflows/compose-smoke.yml
@@ -42,11 +42,19 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
-      # Compose reads SECRET (${SECRET:?...}) even for `config`, so seed a .env
-      # at the repo root (where the compose files' `env_file: ../../.env`
-      # resolves to). A throwaway secret is fine — this stack is torn down.
-      - name: Seed .env with a throwaway SECRET
-        run: echo "SECRET=$(openssl rand -hex 32)" > .env
+      # Compose interpolates ${SECRET:?...} even for `config`. Interpolation
+      # variables come from the shell environment and the project-directory
+      # .env — NOT from a service's `env_file:` (that only feeds the container
+      # at runtime). Because `-f docker/compose/<file>` shifts the project dir
+      # to docker/compose/, a repo-root .env alone would be missed. So export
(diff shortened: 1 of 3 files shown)


delano:
✅ **compose-smoke green on a real runner.** `docker compose up (simple)` passes every step: `config -q` on all 4 documented combos → `up --wait` (the published `:latest` image reaches **healthy** via the new app healthcheck) → **proof-of-life passes end-to-end** (status, homepage + `/dist/assets/*.js`, and the create→reveal→404 at-most-once round-trip against the live instance). CP-5 data-uid workaround held.

Notes:
- This also validates `proof-of-life.sh` (from #3711) against a real running OTS instance for the first time.
- `docker-run-readme` is cron/dispatch-only (skipped on PR by design) — unexercised until a scheduled run.
- The pending checks are unrelated `ci.yml` jobs on the combined tree, not this change.

Issue #3516, opened 2026-06-23
update-locale 500s for anonymous users (NoMethodError on nil cust)


delano:
## Problem Statement

`POST /api/account/update-locale` returns **500** for anonymous (unauthenticated) requests. The endpoint is explicitly designed to support anonymous locale changes (session-only), but any `noauth` request crashes before validation with `undefined method 'locale' for nil (NoMethodError)`.

Repro:
```
curl 'https://dev.onetime.dev/api/account/update-locale' \
  -H 'accept: application/json' --data-raw '{"locale":"nl"}'
# => 500
```

## Early Diagnosis

`AccountAPI::Logic::Account::UpdateLocale#process_params` runs `@old_locale = cust.locale` unconditionally. `process_params` is invoked from `Onetime::Logic::Base#initialize`, so it executes on **every** request — including anonymous ones where `cust` is `nil`.

This is inconsistent with the rest of the class: `raise_concerns`, `perform_update`, and `log_update` all branch on `anonymous_user?` to support the anonymous (session-only) path. Only `process_params` missed the guard. For anonymous users the "old" locale lives on the session, not a Customer record.

Existing tryouts only ever exercised the authenticated path (their comments note *"process_params reads cust.locale, so we need an authenticated user"*), which is why the gap shipped.

## Affected Areas

- [ ] `apps/api/account/logic/account/update_locale.rb` — `process_params` calls `cust.locale` without an `anonymous_user?` guard (line 13)
- [ ] `apps/api/account/logic/account/update_account_field.rb` — base class whose `initialize` triggers `process_params`
- [ ] `lib/onetime/logic/base.rb` — `initialize` invokes `process_params` for all requests
- [ ] `try/unit/logic/account/account_operations_try.rb` — no anonymous-path coverage for UpdateLocale

PR #3807, opened 2026-07-17
[#3516] Move login rate-limit check ahead of argon2 comparison

delano:
## Summary

Stacked on `security/audit-2026-07-06-high-medium`. Closes the **M-4 known residual** that was documented as "accepted" in the 2026-07-06 audit.

A locked/blocked login subject previously still burned **one argon2 hash per request**: `check_login_rate_limit!` ran at the top of `raise_concerns`, but `process_params` (which fires from the `Base` constructor) had *already* performed the `passphrase?` comparison by then. The limiter bounded sustained guessing but did not bound CPU on the locked path — a DoS-amplification gap.

The in-code comment claimed closing it required "restructuring the credential pipeline." It doesn't. The rate-limit keys never depended on the credential lookup:

- `login_rate_limit_email` -> `params['login']`
- `login_rate_limit_ip` -> `strategy_result.metadata[:ip]`

Both are in hand at the top of `process_params`, before `find_by_email`. So the check simply moves ahead of the comparison.

## Change

- **`check_login_rate_limit!` moves** to the top of `process_params`, before `Customer.find_by_email` / `passphrase?`. A locked subject now raises `LimitExceeded` from `AuthenticateSession.new` — no password hash computed.
- **`record_failed_login_attempt!` stays** in `raise_concerns` (it must count *after* the credential check — a read-only probe can't sit after the failure it records, so the two halves are split).

## Why surfacing is unchanged (429)

`Onetime::LimitExceeded < Forbidden`, **not** `< FormError`. `execute_with_error_handling` only rescues `FormError`, so it never caught this exception even in the old location — it always propagated to Otto's top-level `handle_error` (`otto_hooks.rb`: `LimitExceeded => 429, log_level: :warn`). Raising from the constructor (controller line 82, inside Otto's `@app.call` rescue) lands in the exact same handler. Verified end-to-end by the rodauth-expert review: same 429, same ADR-013 wire shape, same log level.

## Why checking before existence confirmation is safe (no downside)

- **No enumeration delta** — identical 429 whether or not the email maps to a real account; no timing fork (short-circuits before the DB lookup in both cases).
- **No new state** — the check is a read-only lockout probe; it writes nothing. Failed-attempt *records* for nonexistent emails already happened in `raise_concerns` pre-change.
- **No side-effect interaction** — the only side effect in `process_params` (`migrate_password_hash_if_needed`) is gated on `passwd_matches`, which a locked attacker never hits.

## Tests

Added to `try/unit/logic/authentication/authenticate_session_try.rb` (spy modules count `find_by_email` / `passphrase?`):

1. Locked per-IP subject raises `LimitExceeded` at **construction** (core regression guard).
2. Locked path **never reaches** `find_by_email` or `passphrase?` (the point of the fix — no hash burned).
3. Locked **nonexistent** email also raises at construction, spies still 0 (no enumeration delta).
4. Happy path still authenticates **and clears** rate-limit state.
5. Wrong-password path still **records** a failed attempt and raises the non-enumerating error.

```
try --agent try/unit/logic/authentication/authenticate_session_try.rb   # 14 passed, 0 failed
bundle exec rspec spec/integration/all/authentication_security_spec.rb  # 22 examples, 0 failures
```

## Follow-up (not in this PR)

The M-4 status in `docs/security/security-audit-2026-07-06.md` should flip from "accepted residual" to "fixed." Left out here because the doc currently carries unrelated in-flight **M-1** (compose Valkey hardening) edits in the worktree; fold the M-4 status update into whichever branch owns that doc change.

## Team

- **qa-automation-engineer** — test coverage, ran both suites green.
- **rodauth-expert** — confirmed constructor-raise -> 429 surfacing with file:line evidence.

### DIFF
--- a/apps/web/core/logic/authentication/authenticate_session.rb
+++ b/apps/web/core/logic/authentication/authenticate_session.rb
@@ -23,6 +23,17 @@ module Core::Logic
         @stay                    = true # Keep sessions alive by default
         @session_ttl             = (stay ? 30.days : 20.minutes).to_i
 
+        # M-4/#3516: gate BEFORE the argon2 passphrase comparison below, so a
+        # locked-out subject never triggers an expensive password hash. Both
+        # rate-limit keys derive from data already in hand here — the submitted
+        # email (params) and the client IP (strategy metadata) — so the check
+        # needs neither the customer lookup nor the comparison it precedes. This
+        # runs inside the Base constructor (process_params fires there, see
+        # #3516); a lockout therefore raises LimitExceeded from `.new`, which the
+        # Otto request hook / Roda ErrorTranslator surface as 429 exactly as they
+        # did when this check lived in raise_concerns.
+        check_login_rate_limit!(login_rate_limit_email, login_rate_limit_ip)
+
         potential = Onetime::Customer.find_by_email(potential_email_address)
 
         return unless potential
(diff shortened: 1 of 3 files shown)


delano:
Both Copilot comments are false positives: tryouts loads each `_try.rb` once per process (no reload loop), and the spy counters are reset before each locked-path assertion (L131-132, L150-151) which expect empty/zero because the spied methods are never reached. Unguarded prepend matches the existing pattern in `try/unit/cli/domains/reconcile_sender_command_try.rb`. No changes.

Issue #3840, opened 2026-07-21
Orchestrate SSO account-linking remediation (sequencing for #3838)


delano:
## Problem Statement

#3838 captures the full SSO account-linking resolution (H-3 follow-through) as seven work items, but does not sequence them. Meanwhile #3836 reporters are locked out of SSO **now**, and the durable flows are weeks of work. This ticket is the execution order — severity- and dependency-driven — so each slice can be branched, shipped, and checked off independently. Scope and acceptance criteria live in #3838; this ticket only orders them.

## Ordering Rationale

- **Item 5 first.** The `(provider, uid)` key concern is treated as confirmed, not investigative — schedule the schema fix, not the audit. It is also a hard dependency: items 1–3 all *write* identity rows, so migrating the key after building the flows means a second backfill and re-testing every flow.
- **Items 4+3 are the fast unblock** for #3836's install class (self-hosted single-tenant) and ship as v0.26.2.
- **Settings panel (1a) before interstitial (1b):** simpler surface to prove the bind primitive (user already authenticated, no callback state to carry across a password challenge), it is the tracked H-3 follow-up, and it gives item 4's error message its link target.
- **Mailbox proof (2) last of the flows:** builds on 1b's pending-link token machinery and has consent-copy requirements to get right.
- **Item 6 is read-only and independent** — runs in parallel anytime; findings feed #2729.
- **#3499 is revisited only after 1–4 land** (per #3838 item 7).

*Escape valve:* if the Phase 0 migration takes more than a few days, ship Phase 1 first — item 3's boot guard already excludes the multi-tenant surface, and the flag's auto-linked rows get backfilled with everything else.

### Phase 0 — schema authority (#3838 item 5) — blocks Phases 2–4

- [x] Migrate `account_identities` unique key to `(issuer, provider, uid)`; backfill `issuer` from SSO config at link time
- [x] Regression test: two OIDC configs, different `issuer`, colliding `sub`
- [x] Verify against `rodauth-omniauth`'s identity lookup path

### Phase 1 — v0.26.2 patch (#3838 items 4 + 3) — unblocks #3836

- [x] `Login.vue`: map `auth_error=account_exists_link_required` to a specific message (message-only; settings link added in Phase 2)
- [x] `sso.trust_email_for_linking` per-provider flag, default `false`, `warn`-level audit event per auto-link
- [x] Boot guard: refuse / hard-warn when any `CustomDomain::SsoConfig` exists
[...]

Issue #3838, opened 2026-07-21
SSO: proper resolution for account-linking (H-3 follow-through + class fix)


delano:
## Context

The reporter of #3836 upgraded a v0.25 install to v0.26 and lost the ability to sign in with Entra/OIDC — the SSO callback bounces back to `/signin` with a generic "SSO authentication failed" message. Root cause: the AUDIT-2026 security fix H-3 (commit `aa3d70b2`, *"Refuse OmniAuth email auto-link to existing accounts"*) refuses any callback whose `(provider, uid)` has no row in `account_identities`, where previously (rc0 and all of v0.24/v0.25) the hook silently auto-linked by email.

**H-3 is the right fix, not the bug.** Email-based auto-linking is a well-known SSO anti-pattern (see #3499's sibling concern about trusting IdP-provided email): whoever controls the IdP can mint a token bearing any victim's email and take over that account. What shipped without H-3 was a takeover vector; what shipped with H-3 was a lockout. Neither is acceptable, and the H-3 commit itself explicitly deferred the recovery UI ("no authenticated 'link SSO from account settings' flow exists yet").

This ticket is the v0.26.2 work: make the refusal recoverable without reopening the takeover vector.

Related:
- Reporter: #3836
- Sibling / class: #3499 (SSO accounts when IdP omits the email claim)
- Prior work: #3478 (invalid-email guard at JIT-create)
- H-3 commit: `aa3d70b2` — *[AUDIT-2026] Refuse OmniAuth email auto-link to existing accounts*
- Owning hook: `apps/web/auth/config/hooks/omniauth.rb` (`account_from_omniauth`)
- Identities table: `apps/web/auth/migrations/006_omniauth_identities.rb`
- Provider registration: `apps/web/auth/config/features/omniauth.rb`

## Blast radius (corrected)

An earlier draft claimed `account_identities` is new in v0.26 and therefore "empty on every v0.25 → v0.26 upgrade, so every pre-existing SSO user is refused." **That is wrong**:

- Migration `006_omniauth_identities` shipped with **v0.24.0** (`5ec84d7532`, "Add OmniAuth SSO infrastructure"). It is present at v0.25.11; `features/omniauth.rb` is byte-identical v0.25.11 → v0.26.0; `rodauth-omniauth` is pinned at 0.6.2 across both.
[...]

PR #3845, opened 2026-07-21
[#3840 Phase 0] Issuer-scoped SSO identity key — close cross-tenant takeover (#3838 item 5)

delano:
**Phase 0 of the #3840 stacked-PR program** (sequences #3838). Targets the orchestration branch. **Blocks Phases 2–4** (they write identity rows against this key).

## The vulnerability (#3838 item 5)
`account_identities` was uniquely keyed on `(provider, uid)`. `provider` is the strategy **name** (`'oidc'`, `'entra'`) — identical across every tenant using that strategy via `CustomDomain::SsoConfig`. Two IdPs (issuers) can each assert the same `sub`. Under the old key, tenant B's callback asserting `sub=user-123` matched tenant A's row → **tenant B signs in as tenant A's user**. Re-keying on `(issuer, provider, uid)` makes colliding identities distinct rows.

## Approach A — platform grace + lazy upgrade (ratified)
Existing rows can't be given a real issuer (#3838 bars log-scraping; it isn't stored on the row), so migration 008 backfills the `''` sentinel unconditionally and the **read path** resolves it:
1. Exact lookup `(provider, resolved_issuer, uid)` first.
2. **Platform** (env-configured) callbacks fall back to the legacy `(provider, '', uid)` row and **lazily upgrade** its issuer (self-heal).
3. **Tenant** (`CustomDomain::SsoConfig`) callbacks are **issuer-exact only** and never touch the `''` row — the legacy fallback on the tenant path *is* the takeover.

The platform-vs-tenant signal (`session[:validated_omniauth_domain_id]`, set by the tenant hook in `before_omniauth_callback_route`) is **verified in the gem** to be present before `retrieve_omniauth_identity` runs — this was the critical uncertainty and it's resolved with evidence, not assumption.

## Changes
- `migrations/008_issuer_scoped_identities.rb`: `issuer NOT NULL DEFAULT ''`, drop `(provider,uid)` unique (backend-branched `drop_constraint`), add `(provider,issuer,uid)` unique. Reversible `down`.
- `config/features/omniauth.rb`: pure `resolve_issuer` / `platform_path?` / `lookup_identity` functions wired via `retrieve_omniauth_identity` + insert/update hashes (persist + self-heal issuer).
- `spec_helper.rb#create_omniauth_tables` mirrors 008 (specs build schema from here, not the migration).

## Verification
- Unit `omniauth_spec.rb`: **91 examples, 0 failures** (re-run independently).
- Full-mode integration `omniauth_issuer_scoped_identity_spec.rb`: **7 examples, 0 failures** — takeover regression (two tenants, colliding uid, different issuer → distinct rows, no cross-bind), platform grace + lazy upgrade, tenant no-grace.
- **Postgres `up` path verified empirically** on a scratch PG DB: `account_identities_provider_uid_key` dropped correctly, `issuer` backfilled to `''`, colliding `(provider,uid)`/different-issuer both insert, exact `(provider,issuer,uid)` dup rejected.

## Ratified consequence (not a defect)
Existing **tenant** SSO users whose rows predate this migration miss the exact lookup and are refused (`account_exists_link_required`) until they re-link via a later phase's authenticated (Phase 2) or mailbox (Phase 4) flow. A tenant issuer backfill is **impossible** — rows carry no domain association, so a legacy `(oidc, uid, '')` row can't be mapped to a specific tenant's issuer. Platform users self-heal via the grace path. This is the tradeoff of the chosen approach; the Phase 1 trust flag does not apply here (it's platform-only + boot-guarded off multi-tenant).

## Follow-up
- Down-migration verified on SQLite; PG-down not exercised (rollback-only path).

### DIFF
--- a/apps/web/auth/config/features/omniauth.rb
+++ b/apps/web/auth/config/features/omniauth.rb
@@ -52,6 +52,153 @@ module Auth::Config::Features
       configure_entra_id_provider(auth)
       configure_github_provider(auth)
       configure_google_provider(auth)
+
+      # Issuer-scoped identity lookup (#3840 Phase 0 / #3838 item 5).
+      configure_issuer_scoped_identities(auth)
+    end
+
+    # ========================================================================
+    # Issuer-scoped SSO identities — cross-tenant takeover fix (#3838 item 5)
+    # ========================================================================
+    #
+    # account_identities is keyed on (provider, issuer, uid). `provider` is the
+    # strategy NAME ('oidc', 'entra'), identical across every tenant using that
+    # strategy, so (provider, uid) alone let a second IdP asserting the same
+    # `sub` match the FIRST tenant's row → account takeover. Adding `issuer`
+    # makes colliding identities distinct rows.
+    #
+    # APPROACH A — platform grace + lazy upgrade. Pre-existing rows have the
+    # sentinel issuer '' (migration 008 backfills unconditionally; the real
+    # issuer is unreconstructable per #3838). The read path resolves it:
+    #   1. Exact lookup (provider, resolved_issuer, uid).
+    #   2. PLATFORM path only: fall back to the legacy (provider, '', uid) row
+    #      and lazily upgrade its issuer to resolved_issuer (self-heal).
+    #   3. TENANT path: issuer-exact ONLY — NEVER the '' fallback. The legacy
+    #      fallback on the tenant path IS the item-5 takeover.
+    #
+    # The pure decision functions below are driven verbatim by the auth-class
+    # helpers wired in configure_issuer_scoped_identities, and unit/integration
+    # tested directly.
+
+    # Sentinel issuer for identities with no known IdP issuer. ALWAYS '' — never
+    # nil (a NULL vs '' split breaks the (provider, issuer, uid) unique index).
+    ISSUER_SENTINEL = ''
+
+    # Resolve the issuer for the current callback.
+    # Precedence: strategy option (authoritative) > ENV['OIDC_ISSUER'] for OIDC
+    # > '' sentinel (non-OIDC / unknown).
+    #
+    # @param strategy_options [Hash, nil] omniauth_strategy&.options
+    # @param provider [String, Symbol] omniauth_provider (route name)
+    # @param oidc_route_name [String] configured OIDC route name (OIDC_ROUTE_NAME)
+    # @param env_oidc_issuer [String, nil] ENV['OIDC_ISSUER']
+    # @return [String] resolved issuer or the '' sentinel
+    def self.resolve_issuer(strategy_options:, provider:, oidc_route_name:, env_oidc_issuer:)
+      option_issuer = strategy_options && strategy_options[:issuer]
+      return option_issuer.to_s if option_issuer && !option_issuer.to_s.empty?
+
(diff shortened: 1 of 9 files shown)


delano:
## Review follow-up — thread disposition (`edcf61a`, `a14a25b`)

Working through the review surfaced that the earlier `up->down->up` fix was **postgres-only** — the SQLite branch still had the bug. Detail below.

**P0 · Entra tenants share identity scope** (Greptile + Copilot) — **Fixed** in `373875141`. `resolve_issuer` fell through to `''` for Entra (its strategy exposes `:tenant_id`, not `:issuer`); it now falls back to the validated id-token `iss` claim, so Entra identities are issuer-scoped. Entra unit coverage added.

**P1 · `up->down->up` drops a missing constraint** (Greptile + Copilot) — **Fixed** in `edcf61a`. The previous commit only made the **postgres** branch idempotent; the **SQLite** branch still re-imposed the old key. On SQLite, `down` restores `(provider, uid)` as a *standalone* unique index, which `up`'s `drop_constraint` rebuild cannot see — so an up→down→up of migration 008 leaves that stale index in place and re-imposes the pre-issuer `(provider, uid)` key, rejecting the colliding-issuer inserts that issuer-scoping exists to permit.

**Scope — narrow, not a live exploit.** Production runs PostgreSQL, where the original fix was already correct. This only affects SQLite, and only on the uncommon path of rolling migration 008 back and re-applying it — i.e. the test suite, plus any SQLite self-hoster who performs that specific rollback/re-migrate. So: low probability, real consequence if hit — worth fixing because it's cheap and correct, not because it's an open door on the primary path. SQLite now drops the index idempotently (`DROP INDEX IF EXISTS`), mirroring postgres, with regression guards pinning the cycle on **both** backends.

**P1 · rollback rejects valid issuer collisions** (Greptile + Copilot) — **Working as intended, now guarded.** Collapsing two issuers back onto one `(provider, uid)` key is impossible by construction — that is precisely the takeover this schema keeps distinct rows to prevent. A new SQLite test pins that such a rollback fails *loudly* (`Sequel::DatabaseError`) and preserves the rows rather than silently dropping one. Documented in the migration `down` block.

**P1 · constraint name assumes PG convention** (Greptile) — **Out of scope.** This project ships PostgreSQL and SQLite only; the MySQL/MSSQL `else` branch is best-effort and would fail loudly on a mismatched name anyway.

**Nit · migration comment path** (Copilot) — **Fixed** (`edcf61a`). Full-pathed the `lookup_identity` reference.

**Nit · silent-rescue cleanup** (Copilot) — **Fixed** (`edcf61a`). Removed; the deletes are idempotent and FK-ordered, so a cleanup failure now surfaces instead of leaving stray rows in the shared in-memory DB.

**Follow-up · test-helper duplication** — **Fixed** (`a14a25b`). `issuer_migration_version` (filename→version derivation) and `insert_account` were duplicated across the two migration specs; both already include `MigrationTestHelpers`, so the pair is hoisted there. Behavior-preserving extraction — identical call sites and signatures.

**Verified:** SQLite migration spec 21/0 (re-run after the helper hoist) · PG migration spec 30/0 (real Postgres, pre-hoist — the hoist is a behavior-preserving extraction with identical call sites) · issuer-scoped integration 7/0 · rubocop clean on CI-inspected paths.

PR #3860, opened 2026-07-23
Add design system readiness audit and rolling-wave plan

delano:
## Summary

Adds `docs/planning/design-system-readiness.md`: an audit of all four frontend apps (`secret`, `workspace`, `session`, `admin`) plus the shared layer, tokens, and packaging, assessing readiness to extract a proper design system library.

Key findings captured in the document:
- The token/runtime-branding foundation (`@theme static`, `useBrandTheme`, `brand-palette`) is the strongest asset (5/5); the atomic primitive layer (`Button`, `Input`, `Card`, `Table`, `Dialog` shell) is entirely absent, driving most duplication.
- Per-app maturity: admin 3.5/5 (its `kit/` is the proto-design-system seed), secret 2.7/5, session 2.0/5, workspace 2.0/5.
- Plan is written rolling-wave style: Wave 1 (semantic neutral tokens + atoms with pilot adoption) is elaborated to task level; Waves 2–4 (kit promotion + migration, branded-fork collapse, package boundary/Storybook) are intentionally coarse and get elaborated at each wave boundary.

## Related Issues

None.

## Test Plan

Documentation-only change; no code paths affected. Verified internal file references (paths/lines cited in the audit) against the current tree.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EuwAjk7dHWgdCYCCWc5YuN)_ _(Fable 5 Low)_

### DIFF
--- a/docs/planning/design-system-readiness.md
+++ b/docs/planning/design-system-readiness.md
@@ -1,0 +1,148 @@
+# Design System Readiness — Audit & Rolling-Wave Plan
+
+**Date:** 2026-07-23
+**Scope:** All four frontend apps (`src/apps/{secret,workspace,session,admin}`), the shared
+layer (`src/shared/`), tokens (`src/assets/style.css`), and packaging.
+**Planning style:** Rolling wave / progressive elaboration — Wave 1 is specified to
+task-level detail; later waves are intentionally coarse and get elaborated as the
+preceding wave lands and teaches us more.
+
+---
+
+## 1. Executive summary
+
+The codebase is **not yet ready to extract a design system library**, but the gap is
+narrower than it looks:
+
+- **Best-in-class foundation:** the design-token and runtime custom-domain branding
+  stack (`@theme static` in `src/assets/style.css`, `useBrandTheme.ts`,
+  `brand-palette.ts` with WCAG contrast checking) is production-grade and already
+  token-driven. This is the asset to build around.
+- **Missing primitive layer:** there is **no shared `Button`, `Input`, `Field`,
+  `Card`, `Table`, or generic `Dialog` shell anywhere** in `src/shared/components`.
+  Every app hand-rolls these (hundreds of raw `<button>`/`<input>` sites). This one
+  gap drives most of the duplication and inconsistency found below.
+- **The seed exists:** `src/apps/admin/components/kit/` is a deliberate,
+  documented proto-design-system with library-grade typed props/slot APIs.
+  `OIcon` is the one primitive that already behaves like a design system
+  (universally adopted, ~470 uses in workspace alone).
+
+**Verdict:** build tokens-for-neutrals and the atomic primitives *inside*
+`src/shared` first; only then draw a package boundary. Extracting a package today
+would ship the duplication as a library.
+
+## 2. Maturity scorecard
+
+### 2.1 Per-app
+
+| App | Overall | Strengths | Weaknesses |
+|---|---|---|---|
+| `admin` | **3.5/5** | Documented UI kit (`components/kit/`: DataTable, AdminModal, StatCard, FilterBar, KitPagination) with typed generic props, slots, `v-model` contracts; zero hardcoded hex; disciplined shared-component reuse | ~1,650 raw `gray-*` utilities; 96 raw `<button>` / 24 raw `<input>` because no atom exists beneath the kit's molecules; thinner per-surface aria coverage (~38 `aria-*`) |
+| `secret` | **2.7/5** | Heaviest brand-token consumer (~122 `brand-*` uses); strong a11y (`aria-live`, `sr-only` status, focus rings); 82% of files have `dark:` coverage | `canonical/` vs `branded/` component forks are copy-and-diverge duplication; 10+ `SecretLinksTableRow*` variants; no variant API — raw `cornerClass` string props (`ConcealButton.vue`) |
+| `session` | **2.0/5** | Best a11y craft in the repo (~189 `aria-*` refs; full `aria-invalid`/`describedby` wiring; sr-only announcements); consistent dark mode | Zero abstraction: the same `bg-brand-600` submit-button block appears in 14 files; the ~8-line input class string is duplicated in every form (`SignInForm.vue:101-109`); one hardcoded hex `#d45a2a` bypasses tokens |
+| `workspace` | **2.0/5** | Heavy healthy reliance on shared logic (`OIcon` ×469, shared stores/composables ×324 imports); near 1:1 light/dark utility pairing | Worst duplication: 159 raw buttons, 51 raw inputs, 11 hand-rolled tables, 6 feature modals rebuilding dialog chrome; ~5:1 hardcoded-vs-brand color ratio; split folder roots (`billing/` vs `components/billing/`, three `account` locations); focus styles on <50% of interactive files |
+
+### 2.2 Foundation
+
+| Area | Rating | Notes |
+|---|---|---|
+| Runtime domain branding | **5/5** | `src/shared/composables/useBrandTheme.ts` + `src/utils/brand-palette.ts`: 44-key CSS-var palette injection, memoization, WCAG `contrastRatio()`, clean fallback to compiled `@theme static` defaults; 3-step fallback chain in `src/shared/constants/brand.ts` |
+| Design tokens | **4/5** | Tailwind v4 CSS-first (`src/assets/style.css:38` `@theme static`): six 11-shade brand scales (`brand`, `branddim`, `brandcomp`, `brandcompdim`, `brand2`), `--font-brand-*` stacks, `--radius-brand`. **Gap: neutrals are not tokenized** — ~4,000 raw `gray-*`/`slate-*` utilities repo-wide, no semantic `surface`/`ink`/`muted` layer |
+| Icon system | **4/5** | `OIcon` sprite system, 9 collections, lazy registry, a11y-conscious. Universally adopted |
+| Shared components | **3/5** | 94 components in coherent families (modals on headlessui + focus-trap, split global/inline notification system, skeletons) — but no atoms |
+| Docs / Storybook / tests | **2/5** | No Storybook; ~10% component spec coverage; `.interface-design/system.md` has drifted from code (documents flame-orange `#dc4a22` brand vs the neutral-blue default in `brand.ts`) |
+| Library packaging | **2/5** | `pnpm-workspace.yaml` has no `packages:` field (no real workspaces); no Vite lib build, no `exports` map; `@/shared/*` alias would need rewiring |
+
+## 3. Key evidence (for future reference)
+
+- Duplicated input class string: `src/apps/session/components/SignInForm.vue:101-109` (and again at `:132-140`) — the canonical template for a future `BaseInput`.
+- Library-grade API precedent: `src/apps/admin/components/kit/DataTable.vue:29-65` (generic `<T>`, typed columns, controlled sort, cell/header slots); `AdminModal.vue:30-57` (`v-model:open`, `dismissable`, typed emits).
+- Kit intent & isolation: `src/apps/admin/components/kit/index.ts:1-25`.
+- Copy-and-diverge fork: `src/apps/secret/components/canonical/SecretDisplayCase.vue` (263 lines) vs `branded/SecretDisplayCase.vue` (195 lines).
+- Token system: `src/assets/style.css:31-152` (incl. the documented `@theme static` rationale); extended tokens (`brand2-*`, `brandbg`, `brandtext`, `rounded-brand`) are injected but not yet consumed by any view (`style.css:99-109`).
+- Missing atoms: `src/shared/components/ui/` contains only specialized buttons (`CopyButton`, `SplitButton`, `ButtonGroup`) — no generic `Button`/`Input`.
+
+---
+
+## 4. Rolling-wave plan
+
+Waves are sequenced by dependency, not calendar. Each wave ends with a review that
+elaborates the next wave to task-level detail and updates this document
+(progressive elaboration — later waves below are deliberately coarse).
+
+### Wave 1 — Atoms + neutral tokens (elaborated now, ready to execute)
+
+Goal: create the missing primitive layer inside `src/shared` and the semantic
+neutral tokens it needs. No package extraction yet. No mass migration yet —
+each primitive is proven in one pilot surface per app.
+
+1. **Semantic neutral tokens** in `src/assets/style.css` `@theme` block:
+   `--color-surface-*` (page/raised/sunken), `--color-ink-*` (default/muted/faint),
+   `--color-edge-*` (border/divider), mapped to the existing gray ramp with
+   light/dark values so `dark:` pairs collapse into the token layer. Follow the
+   existing `@theme static` pattern so custom domains can override them later.
+2. **`BaseButton`** (`src/shared/components/base/BaseButton.vue`):
+   typed `variant` (`primary | secondary | ghost | danger`), `size` (`sm | md | lg`),
+   `loading`, `disabled`. Merge session's a11y treatment (sr-only busy status,
+   `aria-busy`) with admin-kit's API style. Built-in focus-visible ring —
+   this fixes workspace's missing-focus-state gap by construction. Absorb the
+   `cornerClass` smell into a `rounded` variant driven by `--radius-brand`.
+3. **`BaseInput` + `FormField`** wrapper (label, hint, error, `aria-invalid`/
+   `aria-describedby` wiring lifted verbatim from `SignInForm.vue`). Cover
+   `input`, `textarea`, `select` via one field contract.
+4. **`BaseCard`** and a generic **`Dialog` shell** (extract the overlay/panel
+   chrome the 6 workspace feature modals each rebuild; keep
+   `ConfirmDialog`/`SimpleModal` as compositions of it).
+5. **Pilot adoption, one surface per app** (proof, not migration):
+   session `SignInForm`, workspace `DomainSsoConfigForm`, secret
+   `ConcealButton`/`GenerateButton`, admin one kit view. Each pilot deletes its
+   local copies of the duplicated class strings.
+6. **Tests + doc truth:** component specs for each new atom (pattern:
+   `src/tests/shared/components/`); reconcile `.interface-design/system.md`
+   with the actual code defaults (resolve the flame-orange vs neutral-blue drift)
+   and record the decision in its Decisions Log.
+
+Exit criteria: atoms exist with typed variant APIs, pilots merged, zero new raw
+`<button>`/`<input>` in touched files, Wave 2 elaborated from pilot learnings.
+
+### Wave 2 — Promote molecules & migrate (coarse; elaborate at end of Wave 1)
+
+- Promote admin `kit/` upward to `src/shared` (drop the `Admin` prefix; move
+  console-specific "heavy-rule" defaults into variants). DataTable replaces the
+  11 hand-rolled workspace tables and informs a slot-driven replacement for the
+  `SecretLinksTableRow*` sprawl.
+- Codemod raw `gray-*`/`slate-*` utilities onto the Wave 1 semantic tokens,
+  app by app (expected order: session → secret → workspace → admin, smallest
+  and cleanest first).
+- Migrate remaining raw buttons/inputs onto the atoms; normalize workspace's
+  split folder roots (`billing/` vs `components/billing/`, the three `account`
+  locations) as files are touched.
+- Open questions to resolve during elaboration: codemod tooling choice;
+  whether `brand2` neutrals or Tailwind grays back the semantic tokens;
+  per-domain overridability of neutrals.
+
+### Wave 3 — Collapse the branded fork (coarse)
+
+- Merge secret's `canonical/` vs `branded/` component pairs into single
+  token-driven components using the existing `useBrandTheme` runtime pipeline —
+  the infrastructure already supports this; the fork predates it.
+- Wire up the injected-but-unused extended tokens (`brand2-*`, `brandbg`,
+  `brandtext`, `rounded-brand`) as the fork collapses.
+
+### Wave 4 — Package boundary & tooling (coarse; only after primitives are proven)
+
+- Real pnpm workspaces (`packages:` field), a Vite library build with an
+  `exports` map for the extracted `src/shared` design layer; rewire the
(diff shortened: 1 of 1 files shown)


docs/planning/design-system-readiness.md
```diff
@@ -0,0 +1,148 @@
...
+would ship the duplication as a library.
+
+## 2. Maturity scorecard
+
+### 2.1 Per-app
+
+| App | Overall | Strengths | Weaknesses |
+|---|---|---|---|
+| `admin` | **3.5/5** | Documented UI kit (`components/kit/`: DataTable, AdminModal, StatCard, FilterBar, KitPagination) with typed generic props, slots, `v-model` contracts; zero hardcoded hex; disciplined shared-component reuse | ~1,650 raw `gray-*` utilities; 96 raw `<button>` / 24 raw `<input>` because no atom exists beneath the kit's molecules; thinner per-surface aria coverage (~38 `aria-*`) |
+| `secret` | **2.7/5** | Heaviest brand-token consumer (~122 `brand-*` uses); strong a11y (`aria-live`, `sr-only` status, focus rings); 82% of files have `dark:` coverage | `canonical/` vs `branded/` component forks are copy-and-diverge duplication; 10+ `SecretLinksTableRow*` variants; no variant API — raw `cornerClass` string props (`ConcealButton.vue`) |
```
delano: Accuracy nit: "10+ `SecretLinksTableRow*` variants" overcounts. There are **four** visual variants — `SecretLinksTableRowTimeline/Console/Ledger/SlotMachine.vue` — plus the `SecretLinksTableRow.vue` async loader and `SecretLinksTableRowActions.vue` (6 files total, not 10+). The point stands strongly regardless (~1,750 lines of parallel logic across the four variants, each re-implementing memo editing, clipboard, `itemState` priority, TTL formatting), but I'd correct the count so the scorecard is defensible.

Worth adding the underlying diagnosis: these are documented A/B design experiments, so Wave 3's "slot-driven replacement" should be paired with an explicit *decision to retire the losing variants* — otherwise consolidation just relocates the sprawl.

---

docs/planning/design-system-readiness.md
```diff
@@ -0,0 +1,148 @@
...
+| `session` | **2.0/5** | Best a11y craft in the repo (~189 `aria-*` refs; full `aria-invalid`/`describedby` wiring; sr-only announcements); consistent dark mode | Zero abstraction: the same `bg-brand-600` submit-button block appears in 14 files; the ~8-line input class string is duplicated in every form (`SignInForm.vue:101-109`); one hardcoded hex `#d45a2a` bypasses tokens |
```
delano: Two refinements on the `#d45a2a` claim:

1. **It's a JS fallback constant, not a hardcoded utility class.** The only occurrence is `src/apps/session/views/AcceptInvite.vue:128` — `invitation?.branding?.primary_color || bootstrapStore.domain_branding?.primary_color || '#d45a2a'`. That's a different (and more defensible) category than the copy-pasted class strings this row otherwise indicts; it's the last link of a branding fallback chain, so I'd characterize it as "a stray brand default outside the token fallback chain" rather than "bypasses tokens."

2. **It's actually a *third* brand default**, which strengthens your drift finding on line 52. You have neutral-blue `#3b82f6` compiled into `@theme static`, flame-orange `#dc4a22` in `.interface-design/system.md`, and now flame-ish `#d45a2a` here. Three different "brand primary" defaults across three sources is the real story — I'd cross-link this cell to the Docs/Storybook drift row and fold "reconcile the brand-default constant" into Wave 1 item 6.

---

docs/planning/design-system-readiness.md
```diff
@@ -0,0 +1,148 @@
...
+| Library packaging | **2/5** | `pnpm-workspace.yaml` has no `packages:` field (no real workspaces); no Vite lib build, no `exports` map; `@/shared/*` alias would need rewiring |
+
+## 3. Key evidence (for future reference)
+
+- Duplicated input class string: `src/apps/session/components/SignInForm.vue:101-109` (and again at `:132-140`) — the canonical template for a future `BaseInput`.
+- Library-grade API precedent: `src/apps/admin/components/kit/DataTable.vue:29-65` (generic `<T>`, typed columns, controlled sort, cell/header slots); `AdminModal.vue:30-57` (`v-model:open`, `dismissable`, typed emits).
+- Kit intent & isolation: `src/apps/admin/components/kit/index.ts:1-25`.
+- Copy-and-diverge fork: `src/apps/secret/components/canonical/SecretDisplayCase.vue` (263 lines) vs `branded/SecretDisplayCase.vue` (195 lines).
+- Token system: `src/assets/style.css:31-152` (incl. the documented `@theme static` rationale); extended tokens (`brand2-*`, `brandbg`, `brandtext`, `rounded-brand`) are injected but not yet consumed by any view (`style.css:99-109`).
```
delano: Correction: the four extended tokens aren't uniformly unconsumed. **`rounded-brand` is already wired up** — see `src/apps/workspace/components/dashboard/SecretPreview.vue:68` and `src/shared/components/base/BaseUnknownSecret.vue:29`, both of which return `'rounded-brand'` and back it with an injected `--radius-brand`. It's `brand2-*`, `brandbg`, and `brandtext` that are injected-but-unconsumed (confirmed by `SimpleBrandPanel.vue:17`'s own comment noting `--color-brand2-*` is injected "but no [view consumes it]").

Suggest tightening to: "`brand2-*`/`brandbg`/`brandtext` are injected but not yet consumed by any view; `rounded-brand` is partially wired (SecretPreview, BaseUnknownSecret)." This matters because Wave 3's "wire up the injected-but-unused extended tokens" is smaller than stated for radius and larger for the color scales.

---

docs/planning/design-system-readiness.md
```diff
@@ -0,0 +1,148 @@
...
+Goal: create the missing primitive layer inside `src/shared` and the semantic
+neutral tokens it needs. No package extraction yet. No mass migration yet —
+each primitive is proven in one pilot surface per app.
+
+1. **Semantic neutral tokens** in `src/assets/style.css` `@theme` block:
+   `--color-surface-*` (page/raised/sunken), `--color-ink-*` (default/muted/faint),
+   `--color-edge-*` (border/divider), mapped to the existing gray ramp with
+   light/dark values so `dark:` pairs collapse into the token layer. Follow the
+   existing `@theme static` pattern so custom domains can override them later.
+2. **`BaseButton`** (`src/shared/components/base/BaseButton.vue`):
```
delano, line 89: `BaseButton` is the right first atom, but the plan should say what happens to the **existing specialized buttons** you list on line 63 (`CopyButton`, `SplitButton`, `ButtonGroup`, and `CopyToClipboardButton`). If `BaseButton` lands without refactoring those to compose it, you ship *two* button systems and the inconsistency you're trying to kill persists in the shared layer itself. Recommend an explicit Wave 1 sub-task: "reseat `CopyButton`/`SplitButton` on `BaseButton` (or document why they stay bespoke)." Same reasoning applies to `BaseInput` vs the existing `forms/PasswordStrengthChecker` / `DetailField`.

Also — a `danger` variant on `BaseButton` implies a semantic danger color, but Wave 1 item 1 only defines `surface`/`ink`/`edge` neutral tokens. The status ramp (`danger`/`warning`/`success`) that `secret`'s `statusConfig` maps and every `red-50` alert box need is missing from the token plan. I'd add semantic status tokens to Wave 1 item 1 so `BaseButton`'s `danger` variant and the alert primitives have something to reference.

---

docs/planning/design-system-readiness.md
```diff
@@ -0,0 +1,148 @@
...
+`<button>`/`<input>` in touched files, Wave 2 elaborated from pilot learnings.
+
+### Wave 2 — Promote molecules & migrate (coarse; elaborate at end of Wave 1)
+
+- Promote admin `kit/` upward to `src/shared` (drop the `Admin` prefix; move
+  console-specific "heavy-rule" defaults into variants). DataTable replaces the
+  11 hand-rolled workspace tables and informs a slot-driven replacement for the
+  `SecretLinksTableRow*` sprawl.
+- Codemod raw `gray-*`/`slate-*` utilities onto the Wave 1 semantic tokens,
+  app by app (expected order: session → secret → workspace → admin, smallest
```
delano: The codemod order `session → secret → workspace → admin` conflates two different orderings. "Smallest and cleanest first" argues for **admin first** — it already has zero hardcoded hex and the most disciplined token usage (§2.1 rates it highest), so it's the cleanest *reference implementation* of what "migrated" looks like. But you've placed it last. session is smallest but its color is mostly already on `brand-*` tokens, so a neutral-token codemod there exercises little.

Suggest reframing: migrate **admin first as the golden-path template** (small delta, proves the token mapping), then tackle **workspace** (largest raw-palette surface, ~3,200 utilities — where the codemod earns its keep), leaving session/secret to fold in behind the proven pattern. Also add a guardrail note: neutral codemods silently drop `dark:` pairs if the token doesn't encode both sides — the plan's strongest existing asset is near-universal dark-mode coverage, and a mechanical `gray-*`→token sweep is exactly where it can regress. Snapshot/visual-diff gating belongs in the exit criteria here.

---

docs/planning/design-system-readiness.md
```diff
@@ -0,0 +1,148 @@
...
+- Migrate remaining raw buttons/inputs onto the atoms; normalize workspace's
+  split folder roots (`billing/` vs `components/billing/`, the three `account`
+  locations) as files are touched.
+- Open questions to resolve during elaboration: codemod tooling choice;
+  whether `brand2` neutrals or Tailwind grays back the semantic tokens;
+  per-domain overridability of neutrals.
+
+### Wave 3 — Collapse the branded fork (coarse)
+
+- Merge secret's `canonical/` vs `branded/` component pairs into single
```
delano, line 143: One under-stated cost here: collapsing `canonical/` and `branded/` isn't just a token swap — the two trees share the **`BaseSecretDisplay` name but expose incompatible slot vocabularies**. `canonical/BaseSecretDisplay.vue` is a ~23-line slot passthrough (`header/content/warnings/cta/footer`); `branded/BaseSecretDisplay.vue` is ~175 lines with real logic and a *different* slot set (`logo/content/action-button/footer`). So "merge into single token-driven components" first requires designing a unified slot API and reconciling the branded tree's font-remeasure / show-more-clamp logic against canonical's marketing chrome (sign-in CTAs, `NeedHelpModal`). Worth calling out as its own elaboration task in Wave 3, because it's an API-design problem, not a styling one — and the runtime `useBrandTheme` pipeline solves the theming half but none of the slot-contract half.

---

delano (commented):
Reviewed against an independent audit of all four apps + the shared/token layer, and this document holds up well — the core diagnosis matches what I found independently.

**Strongly agree with:**
- **The central thesis** — the token/runtime-branding stack is the standout asset, and the *missing atom layer* (`Button`/`Input`/`Field`/`Card`/`Dialog`) is the single root cause of most duplication. I independently confirmed there is no `BaseButton`/`BaseInput`/`Card`/`Table` anywhere in `src` (raw `<button>` in 135 files, raw `<input>` in ~62). "Extracting a package today would ship the duplication as a library" is exactly right — build atoms in `src/shared` first, draw the boundary last.
- **admin's `kit/` as the seed.** It's the most mature code in the repo (typed controlled APIs, frozen contracts, an *enforced* isolation test), and promoting it in Wave 2 is the correct sequencing.
- **Neutrals-not-tokenized as the #1 token gap**, and the rolling-wave discipline of proving each primitive in one pilot before mass migration.

**Verified counts** (spot-checked): `OIcon` ×469 in workspace ✓, ~122 `brand-*` in secret ✓, the `SignInForm.vue:101-109` duplicated input block ✓.

**Corrections / gaps** (details inline):
1. "10+ `SecretLinksTableRow*` variants" → there are **4** (+loader +actions = 6 files). Count is off; the argument isn't.
2. `#d45a2a` is a **JS branding-fallback constant** in `AcceptInvite.vue:128`, not a hardcoded class — and it's a *third* brand default alongside `#3b82f6` and `#dc4a22`, which reinforces the drift row.
3. `rounded-brand` is **already consumed** (SecretPreview, BaseUnknownSecret); only `brand2-*`/`brandbg`/`brandtext` are unconsumed — resizes the Wave 3 wiring task.
4. **Semantic *status* tokens (`danger`/`warning`/`success`) are missing from Wave 1's token plan**, yet `BaseButton`'s `danger` variant and every alert box need them. Add them alongside the neutral tokens.
5. **Existing specialized buttons** (`CopyButton`/`SplitButton`/`ButtonGroup`) need an explicit "reseat on `BaseButton`" task, or you ship two button systems.
6. **Codemod ordering** — lead with admin (cleanest, = golden-path template) then workspace (largest payoff), and gate the neutral sweep with visual-diff snapshots so it can't silently regress the near-universal `dark:` coverage.
7. **secret's branded/canonical merge** is an API-design problem (incompatible `BaseSecretDisplay` slot vocabularies), not just a theming one — worth its own Wave 3 elaboration task.

None of these change the plan's direction — they tighten the evidence and close two real gaps (status tokens, specialized-button reconciliation) before Wave 1 executes. Nice work; this is a genuinely actionable audit.

---
_Generated by [Claude Code](https://claude.ai/code)_ _(Opus 4.8 Low)_

docs/planning/design-system-readiness.md
delano: `10+ SecretLinksTableRow*` variants — re-checking `src/apps/secret/components/`, there are 6 files matching that prefix: `SecretLinksTableRow.vue`, `SecretLinksTableRowActions.vue`, and four visual variants (`Console`, `Ledger`, `Timeline`, `SlotMachine`). That's 4 true variants plus a base + an actions sub-component, not "10+". Since Wave 2 sizes the DataTable-replaces-this-sprawl work off this number, might be worth citing the exact file list here so it's falsifiable.

Smaller note on the same cell: an independent recount of `brand-*` Tailwind-prefixed utilities (`bg-`/`text-`/`border-`/`ring-`/etc.) in `src/apps/secret` comes out to ~190 occurrences (~101 matching lines), not ~122. Not fatal, but worth stating whether the number is occurrences, matching lines, or matching files so the count is reproducible later.

---

docs/planning/design-system-readiness.md
delano: The hex `#d45a2a` in `AcceptInvite.vue:128` is real, but it undersells the actual bug: it's the fallback value for a `--brand-primary` CSS custom property that the component sets via `:style` at 7 separate call sites (lines 308, 338, 379, 409, 471, 553, 629) — and a repo-wide grep for `var(--brand-primary)` returns zero matches anywhere in `src/`. The whole mechanism is dead code, not just one hardcoded fallback bypassing tokens. That might be a stronger citation for this cell, and arguably deserves its own line item (dead theming variable) rather than being folded into "one hardcoded hex."

---

docs/planning/design-system-readiness.md
```diff
@@ -0,0 +1,148 @@
...
+| `secret` | **2.7/5** | Heaviest brand-token consumer (~122 `brand-*` uses); strong a11y (`aria-live`, `sr-only` status, focus rings); 82% of files have `dark:` coverage | `canonical/` vs `branded/` component forks are copy-and-diverge duplication; 10+ `SecretLinksTableRow*` variants; no variant API — raw `cornerClass` string props (`ConcealButton.vue`) |
+| `session` | **2.0/5** | Best a11y craft in the repo (~189 `aria-*` refs; full `aria-invalid`/`describedby` wiring; sr-only announcements); consistent dark mode | Zero abstraction: the same `bg-brand-600` submit-button block appears in 14 files; the ~8-line input class string is duplicated in every form (`SignInForm.vue:101-109`); one hardcoded hex `#d45a2a` bypasses tokens |
+| `workspace` | **2.0/5** | Heavy healthy reliance on shared logic (`OIcon` ×469, shared stores/composables ×324 imports); near 1:1 light/dark utility pairing | Worst duplication: 159 raw buttons, 51 raw inputs, 11 hand-rolled tables, 6 feature modals rebuilding dialog chrome; ~5:1 hardcoded-vs-brand color ratio; split folder roots (`billing/` vs `components/billing/`, three `account` locations); focus styles on <50% of interactive files |
```
delano: Two things worth re-checking on this row:

- **"11 hand-rolled tables"** — a grep for literal `<table` in `src/apps/workspace` finds exactly 3 files (`DomainsTable.vue`, `MembersTable.vue`, `billing/InvoiceList.vue`). Not sure where 11 comes from (maybe counting `role="list"` pseudo-tables or the dashboard cell-renderer helper components too?) — worth citing the file list so the Wave 2 "DataTable replaces N hand-rolled tables" claim is falsifiable.

- **"6 feature modals rebuilding dialog chrome"** — of the 7 `*Modal*.vue` files in workspace, 6 already use HeadlessUI's `Dialog` (only `AccountDeleteButtonWithModalForm.vue` doesn't). The files that actually hand-roll dialog chrome with no focus trap and no Escape handler are `MfaSettings.vue`, `RecoveryCodes.vue`, and `ActiveSessions.vue` — none of which have "Modal" in the name. That's a materially worse a11y gap than "6 HeadlessUI-based modals lacking one shared wrapper." Worth splitting into two distinct findings (with different severity/fixes) in Wave 1 item 4.

---

docs/planning/design-system-readiness.md
```diff
@@ -0,0 +1,148 @@
...
+
+### 2.2 Foundation
+
+| Area | Rating | Notes |
+|---|---|---|
+| Runtime domain branding | **5/5** | `src/shared/composables/useBrandTheme.ts` + `src/utils/brand-palette.ts`: 44-key CSS-var palette injection, memoization, WCAG `contrastRatio()`, clean fallback to compiled `@theme static` defaults; 3-step fallback chain in `src/shared/constants/brand.ts` |
+| Design tokens | **4/5** | Tailwind v4 CSS-first (`src/assets/style.css:38` `@theme static`): six 11-shade brand scales (`brand`, `branddim`, `brandcomp`, `brandcompdim`, `brand2`), `--font-brand-*` stacks, `--radius-brand`. **Gap: neutrals are not tokenized** — ~4,000 raw `gray-*`/`slate-*` utilities repo-wide, no semantic `surface`/`ink`/`muted` layer |
```
delano: "~4,000 raw `gray-*`/`slate-*` utilities repo-wide" looks understated — a grep for `(gray|slate)-[0-9]{2,3}` in `src/apps` alone (not counting `src/shared`) returns 6,233 occurrences; including `src/shared` it's 6,982. Doesn't change the conclusion (still a large gap either way), but if this number is used to size the Wave 2 codemod effort, the real scope looks ~50-75% bigger than what's stated here.

---

delano (commented):
Reviewed the audit and rolling-wave plan against the current `main` tree (I ran an independent readiness review of the same four apps a bit earlier, so cross-checked overlapping claims directly against source rather than just reading the prose).

**The diagnosis holds up well.** Most of the specific counts re-derive cleanly from source: workspace's 159 raw `<button>` / 51 raw `<input>`, admin's 96/24 and ~1,650 raw `gray-*`, `OIcon` ×469 in workspace, `pnpm-workspace.yaml` having no `packages:` field, and the `canonical/`/`branded/` `SecretDisplayCase.vue` fork (263 vs 195 lines, exact) all check out precisely. The core verdict — token/branding foundation is production-grade, the atomic primitive layer (Button/Input/Card/Dialog) is the real gap, `admin/kit` is the closest thing to a working seed — matches what I found independently.

Left four inline comments on §2's scorecard rows where a re-derived number came out different enough from what's stated that it's worth double-checking or citing methodology before other waves get sized against these figures as a baseline: the `SecretLinksTableRow*` "10+" count, the `--brand-primary` dead-code angle behind the `#d45a2a` hex, the workspace "11 hand-rolled tables" / "6 feature modals" figures, and the repo-wide `gray-*`/`slate-*` count. None of these change the overall verdict — if anything the neutral-utility undercount and the dialog a11y nuance suggest the problem in a couple of spots is somewhat worse/different-shaped than stated, not better.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CPLFcpu3yJfKTARngxpwdJ)_

---
_Generated by [Claude Code](https://claude.ai/code)_ _(Sonnet 5 Extra)_

delano:
## Review feedback dispositioned in `195c55aec`

Every re-derived count was re-verified against the live tree before editing (4 parallel read-only passes). Applied the accurate corrections; skipped the one that didn't hold up.

**Applied (12 threads):**
- describedby → `aria-describedby` typo fixed.
- `ui/` folder reworded (holds `EmptyState`/`ThemeToggle`/notifications, not only buttons).
- `SecretLinksTableRow*`: "10+" → **4 visual variants (6 files)**, file list cited.
- `#d45a2a`: recast as a **dead `--brand-primary` mechanism** (7 `:style` setters, 0 `var()` readers) and a third stray brand default alongside `#3b82f6`/`#dc4a22`.
- `rounded-brand` corrected to **already consumed** (SecretPreview:68, BaseUnknownSecret:29); only `brand2-*`/`brandbg`/`brandtext` unconsumed; citation fixed to the declarations (`style.css:112-129`).
- `BaseButton`: added a **reseat specialized buttons** sub-task and **semantic status tokens** (`danger`/`warning`/`success`) to Wave 1.
- Codemod order: reframed **admin-first** (golden-path template) → workspace; added **visual-diff/snapshot gating** to protect `dark:` coverage; added a retire-losing-A/B-variants decision.
- canonical/branded: added a **unified slot-API** reconciliation task to Wave 3 (incompatible slot vocabularies — an API-design problem, not just theming).
- Tables: "11 hand-rolled" → **3** (`DomainsTable`/`MembersTable`/`billing/InvoiceList`).
- Modals: recast as a **dialog-a11y gap** — 6 of 7 `*Modal*.vue` already use HeadlessUI `Dialog`; the real offenders are 4 hand-rolled dialogs (`MfaSettings`/`RecoveryCodes`/`ActiveSessions`/`AccountDeleteButtonWithModalForm`) with no focus-trap or Escape.
- gray/slate: "~4,000 repo-wide" → **~6,650 in `src/apps` (~7,540 incl. `src/shared`)** with the counting method stated.

**Two sub-points trimmed as inaccurate:** the exact gray/slate figures (6,233/6,982) didn't reproduce here (used verified numbers instead), and the "BaseInput vs PasswordStrengthChecker/DetailField" framing — neither renders an `<input>`, so only the button-reseat half was kept.

**Skipped — false positive (1 thread):** `.interface-design/system.md` is **not** gitignored and **is** tracked (`git check-ignore` clears it; `git ls-files` lists it), so the reference is actionable as written. No change.

PR #3862, opened 2026-07-23
[#3857] Make Rodauth reset-password-request path enumeration-safe

### LINKED ISSUE #3857: Rodauth reset_password request path is an account-existence oracle (sweep S3)
## Summary

The Rodauth `:reset_password` feature is enabled (`apps/web/auth/config/features/account_management.rb:12`, `auth.enable :reset_password`). The gem's password-reset **request** path raises `no_matching_login` when the submitted login has no account, which is an account-existence oracle. Meanwhile the app already ships an enumeration-safe custom reset-request endpoint at `apps/api/account/logic/authentication/reset_password_request.rb`, which issues identically whether or not the account exists.

Surfaced by the #3840 "locate-vs-bind" security sweep (#3838 item 6) as finding **S3**. Not a bind-invariant violation — tracked separately from #2729.

## Impact

If the gem's reset-password **request** route is reachable, an unauthenticated caller can enumerate accounts by observing the `no_matching_login` response versus a success/generic response. This contradicts the enumeration-safe posture the custom endpoint was written to enforce.

## Evidence

```
apps/web/auth/config/features/account_management.rb:12   auth.enable :reset_password
```

- Gem oracle: rodauth `reset_password` request handling raises `no_matching_login` for unknown logins.
- App's enumeration-safe alternative: `apps/api/account/logic/authentication/reset_password_request.rb` — issues uniformly regardless of account existence.

## Investigation + Fix

1. **Confirm which route is actually exposed** for password-reset *requests* — the gem's Rodauth route, the app's custom endpoint, or both. (The custom endpoint's enumeration-safety is only load-bearing if the gem route is not independently reachable.)
2. If the gem request route is reachable, either:
   - **Disable** the gem's reset-password *request* route (keep the reset *consume*/confirm path if it's the one in use), or
   - **Override** `no_matching_login` on the request path so it returns the same enumeration-safe response as the custom endpoint.

Note: the `change_password`/`reset_password` **consume** paths and their `password_match?` behavior (see `account_management.rb:112-113`) are a separate concern from the request-path existence leak addressed here.

## Acceptance Criteria

- [ ] The exposed password-reset *request* path does not distinguish "account exists" from "no account" to an unauthenticated caller.
- [ ] Regression test asserting the request path is enumeration-safe (identical response for existing vs non-existing login).

## Related

- Sweep finding S3 (#3838 item 6); sibling finding S2 filed separately.
- App's enumeration-safe custom endpoint: `apps/api/account/logic/authentication/reset_password_request.rb`.

delano:
## Summary

[#3857](https://github.com/onetimesecret/onetimesecret/issues/3857)

Closes an account-existence oracle (CWE-204) on the password-reset **request** path.

**Which route is actually exposed** (issue investigation step 1): in `full` auth mode the Rodauth app is mounted at `/auth`, and because `Rack::URLMap` dispatches to the longest matching prefix it handles *every* `/auth/*` route — including `POST /auth/reset-password-request`. That shadows the enumeration-safe custom endpoint in `apps/web/core` (mounted at `/`), so the **gem's** request route is the one unauthenticated callers reach. The custom endpoint's enumeration-safety is therefore not load-bearing here.

Stock Rodauth 2.44.0 distinguishes "account exists" from "no account" on that request path in three ways, each an existence oracle:

| Branch | Stock response |
| --- | --- |
| Unknown login | `no_matching_login` field-error |
| Existing but unverified/closed | `unverified_account` error |
| Existing, emailed moments ago | "reset password email recently sent" error |

A registered, verified account gets `200 { "success": "An email has been sent..." }`, so comparing responses reveals which addresses have accounts.

**Why override rather than disable the route:** in full mode the reset *consume* path (`/auth/reset-password`) is Rodauth's and expects a Rodauth-generated token, and the reset email links to it (`config/email/reset_password.rb`) — so the request path must stay Rodauth's to mint that token. Disabling it would 404 the flow (Rack::URLMap does not cascade to the core app). Instead, `config/overrides/reset_password_enumeration.rb` overrides the three oracle branches so every POST returns the same generic "email sent" response.

Key properties:
- Side effects (reset-key creation + email) still happen **only** for a valid, open, non-throttled account.
- Rodauth's resend **throttle is preserved** — a recently-emailed account returns the same generic success *without* resending, so closing the oracle does not open an email-bomb regression.
- The `no_matching_login` override is scoped to the reset-password-request route via `current_route`, leaving the shared `account_from_login` untouched on the login route and everywhere else.

## Related Issues

Fixes #3857

## Test Plan

Added `apps/web/auth/spec/integration/full/reset_password_request_enumeration_spec.rb` (full-stack Rack::Test through the mounted Rodauth app), asserting the exposed request path does not distinguish account existence:

- non-existent login returns a generic success (`200`, `success` key, **no** `field-error` / `no matching login`) rather than the stock error tuple;
- an existing **verified** login yields a response byte-identical to a non-existent login;
- an existing **unverified** account is likewise indistinguishable from a non-existent login;
- a **recently-emailed** account stays enumeration-safe *and* the throttle holds (no second email dispatched);
- **happy path:** a valid, verified account still enqueues exactly one reset email (the fix must not suppress real sends).

Delivery is stubbed for determinism; the enumeration property holds with or without a real send.

> **Verified** under Ruby 3.4.10 (rbenv): the enumeration spec passes **5/5** examples, and the full-auth integration directory `apps/web/auth/spec/integration/full/` runs **282 examples, 0 failures** (60 pending — OmniAuth/SSO specs that skip when OIDC routes aren't registered at boot). `rubocop` is clean on the override. Reproduce with `source .env.test && pnpm run test:rspec apps/web/auth/spec/integration/full/reset_password_request_enumeration_spec.rb` (requires Valkey/Redis on 2121 + `AUTH_DATABASE_URL`).

https://claude.ai/code/session_01BtXwmqCnodmPStTFp2UEQu

---
_Generated by [Claude Code](https://claude.ai/code/session_01BtXwmqCnodmPStTFp2UEQu)_

### DIFF
--- a/apps/web/auth/config.rb
+++ b/apps/web/auth/config.rb
@@ -89,6 +89,10 @@ module Auth
       # Method overrides (replace Rodauth methods, not before/after hooks)
       Overrides::PasswordMigration.configure(self)
       Overrides::ErrorHandling.configure(self)
+      # Enumeration safety for the reset-password-request path (issue #3857).
+      # Runs after AccountManagement enables :reset_password above, so the
+      # overridden methods exist.
+      Overrides::ResetPasswordEnumeration.configure(self)
       RodauthOverrides.configure(self)
 
       # Lockout: brute force protection
--- a/apps/web/auth/config/overrides.rb
+++ b/apps/web/auth/config/overrides.rb
@@ -18,4 +18,5 @@
 module Auth::Config::Overrides
   require_relative 'overrides/error_handling'
   require_relative 'overrides/password_migration'
+  require_relative 'overrides/reset_password_enumeration'
 end
--- a/apps/web/auth/config/overrides/reset_password_enumeration.rb
+++ b/apps/web/auth/config/overrides/reset_password_enumeration.rb
@@ -1,0 +1,121 @@
+# apps/web/auth/config/overrides/reset_password_enumeration.rb
+#
+# frozen_string_literal: true
+
+#
+# Reset-Password-Request Enumeration Safety (issue #3857)
+#
+# MECHANISM: method overrides via `auth_class_eval` — these REPLACE Rodauth
+# methods (overrides clobber like hooks do: last definition wins), which is why
+# this file lives in config/overrides/ rather than config/hooks/. `super` inside
+# each override invokes the stock Rodauth implementation (features are included
+# as modules, so the originals live in the class's ancestry).
+#
+# WHY THIS EXISTS
+# ---------------
+# The Rodauth `:reset_password` feature is enabled (config/features/
+# account_management.rb). In `full` auth mode the Rodauth app is mounted at
+# `/auth` and — because Rack::URLMap dispatches to the longest matching prefix —
+# handles EVERY `/auth/*` route, including `POST /auth/reset-password-request`.
+# That route is therefore the one actually exposed to unauthenticated callers;
+# it shadows the enumeration-safe custom endpoint in apps/web/core (mounted at
+# `/`).
+#
+# Stock Rodauth 2.44.0 (lib/rodauth/features/reset_password.rb, the
+# reset_password_request route) distinguishes "account exists" from "no account"
+# on that request path in THREE ways, each an account-existence oracle (CWE-204):
+#
+#   1. no matching login         -> throw_error_reason(:no_matching_login, ...)
+#   2. account not open/verified -> reset_password_request_for_unverified_account
+#                                   (throws :unverified_account)
+#   3. email recently sent       -> "reset_password_email_recently_sent" error
+#
+# A registered, verified account returns 200 { success: "An email has been
+# sent..." }, whereas a non-existent login returns an error tuple carrying a
+# `field-error`. Comparing the two lets an unauthenticated caller enumerate
+# accounts — exactly the posture the custom endpoint
+# (apps/api/account/logic/authentication/reset_password_request.rb) was written
+# to prevent.
+#
+# WHY OVERRIDE RATHER THAN DISABLE THE ROUTE
+# ------------------------------------------
+# We cannot simply disable the request route. In full mode the reset *consume*
+# path (`/auth/reset-password`) is Rodauth's and expects a Rodauth-generated
+# reset token, and the reset email links to Rodauth's reset URL
+# (config/email/reset_password.rb) — so the request path must remain Rodauth's to
+# mint that token. Instead we neutralise the three oracle branches so every POST
+# yields the SAME generic "email sent" response, regardless of whether the login
+# maps to an account, whether that account is open, and whether an email was
+# recently sent.
+#
+# Side effects (reset-key creation + email) still happen only for a valid, open,
+# non-throttled account, and Rodauth's resend throttle is PRESERVED: a
+# recently-emailed account returns the same generic success WITHOUT resending, so
+# closing the oracle does not open an email-bombing regression.
+#
+# See also: config/rodauth_overrides.rb (verify_account error-flash overrides)
+# and config/overrides/password_migration.rb (the `super` override idiom).
+#
+module Auth::Config::Overrides
+  module ResetPasswordEnumeration
+    def self.configure(auth)
+      auth_class = auth.instance_variable_get(:@auth)
+      # reset_password_email_sent_response and the two reset-request predicates
+      # only exist when the feature is enabled; nothing to harden otherwise.
+      return unless auth_class&.features&.include?(:reset_password)
+
+      # rubocop:disable Lint/NestedMethodDefinition -- Rodauth's auth_class_eval pattern
+      auth.auth_class_eval do
+        # ---- Oracle 1: no matching login ----------------------------------
+        # account_from_login is SHARED with the login route, so the
+        # enumeration-safe short-circuit is scoped to the reset-password-request
+        # route via current_route (set by Rodauth's route wrapper). On every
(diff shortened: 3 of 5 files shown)


delano:
Addressed the review feedback in fbcb5e29. Item-by-item:

**Timing side-channel (Greptile P1) — accepted and documented, not padded.** The content/status oracle is closed (identical response across all four cases). A *latency* residual remains because the miss path halts before the reset-key writes + email enqueue that a valid account performs. Tracing the hit path against Rodauth 2.44.0, that delta is ~4 single-row indexed statements on `account_password_reset_keys` inside a transaction, a random-key gen, and one email enqueue — an **async AMQP publish (~ms)** in production with background jobs enabled. So it's single-digit-millisecond server work, an order of magnitude below the network/Puma/GC jitter an off-host attacker sees; it degrades to a statistical-only, many-sample attack rather than a deterministic distinguisher.

I chose **not** to pad the miss path because every equalization is worse than the residual it removes: `account_password_reset_keys` is FK/PK-bound to `accounts`, so a non-existent login has no legal row to mirror — faking it needs junk accounts, insert-and-rollback theater, or dummy queue traffic (new side effects + an abuse surface on an unauthenticated route), while a fixed `sleep` pad is defeated by variance analysis and taxes every legitimate request. The full rationale is now in the override header under `RESIDUAL: TIMING SIDE-CHANNEL`. If you'd rather ship an explicit mitigation despite the tradeoffs, say so and I'll add one.

**FK-unsafe cleanup order (Copilot / Claude review) — fixed.** The `after` block now deletes `account_password_reset_keys` and `account_password_hashes` before `accounts`, so cleanup is real on PostgreSQL instead of a rescued no-op.

**Halt-contract dependency (Claude review) — documented.** Added a `HALT CONTRACT` note explaining the overrides rely on `reset_password_email_sent_response` halting, that a future Rodauth change there would fail *open* (not crash), and that the regression spec is the guard.

**Audit-log registration (Claude review) — not required.** The two new events go through `Auth::Logging.log_auth_event`, a plain SemanticLogger call with no registry/DB coupling to the Rodauth `audit_logging` DSL. Existing call sites (`:login_attempt`, `:around_rodauth`, even `:audit_log_skipped_orphaned_account` inside `audit_logging.rb` itself) emit ad-hoc event names without registration; the new ones follow that established pattern.

**`features` guard (Claude review) — documented** as Rodauth's public reader for enabled feature symbols.

**Minor:** moved `require 'argon2'` to the top of the spec; explained "shrimp" as this app's CSRF authenticity-token param; added the happy-path example above (valid verified account still enqueues exactly one reset email).

Verified under Ruby 3.4.10: enumeration spec **5/5**, full-auth integration directory **282 examples / 0 failures**, `rubocop` clean on the override.

---

delano:
Second review round addressed in 9ea9d82b — plus a **correction to my previous comment**.

**Correction — timing magnitude.** My earlier note called the reset-email dispatch an "async AMQP publish (~ms), an order of magnitude below jitter." That understated the common default. `config/email/delivery.rb` dispatches with `fallback: :sync` ("user is waiting for auth action"), and `jobs_enabled?` is `!$rmq_channel_pool.nil?` — false whenever RabbitMQ isn't wired up — in which case the publisher does a **blocking inline SMTP send** that can dominate the hit/miss delta. The header and changelog now state the dispatch cost is deployment-dependent and do **not** claim the residual is negligible under synchronous delivery.

**#1 Timing residual — still accept-and-document, now on the record.** Re-examined adversarially for a clean mitigation and found none worth shipping: a sleep pad is defeated by variance analysis *and* lets an unauthenticated caller pin Puma threads (a DoS amplifier); mirroring the writes needs junk rows on an FK/PK-bound table; matching only the key-gen is theater (µs vs. the DB+email ms); Rodauth 2.44.0 exposes no request-timing knob. The accepted residual is now documented in both the override header and the **changelog fragment** (so it ships in release notes). This remains a policy sign-off for a maintainer — suggested wording if you'd like it on the thread:
> Accepting the residual response-timing side-channel: the single-request CWE-204 oracle (content + status) is closed and regression-tested; the remaining latency delta has no side-effect-free mitigation, and closing it would add abuse surface on an unauthenticated route. — [maintainer]

**#2 Happy-path matcher — pinned.** `Mail::Message#to` always returns an array of bare addresses (even `"Name <addr>"` parses down to `["addr"]`), so I switched to `contain_exactly(verified)`, which fails loudly on a shape regression where `include()` would silently substring-match.

**Cleanup — removed rather than patched.** Adding a warn to the cleanup `rescue` (the silent-swallow note) immediately exposed an audit-log FK violation: the manual delete was both *incomplete* (an account is FK-referenced by ~20 tables, several `ON DELETE RESTRICT`) and *redundant* — `FullModeSuiteDatabase` already wipes every Rodauth table in an `after(:each)` hook for `:full_auth_mode` specs. So I removed the per-account cleanup entirely and documented why, instead of maintaining a fragile child-first delete list.

**Minor #3 / #4 / #5.** The literal early-return (#3) and defensive-`true` (#4) suggestions would each change the method's fall-through return value and weaken the documented fail-open posture (or trip `Style/RedundantReturn`), so I documented the intent at both sites instead. `auth_class` (#5) is kept — it's accurately named and matches the adjacent comment; inlining `instance_variable_get` into the guard reads worse.

Verified under Ruby 3.4.10: enumeration spec 5/5 (no cleanup warnings), full-auth integration directory 282 examples / 0 failures, rubocop clean on the override.

---

Issue #3877, opened 2026-07-23
[#3840] Phase 4: complete deferred SSO identity bind after MFA (link-sso interstitial)


delano:
## Context

Phase 3 of #3840 (the SSO sign-in interstitial, PR #3870) introduced `/auth/link-sso`: when an unauthenticated SSO sign-in resolves to an existing password-backed account, the user proves ownership with their existing password before the `(provider, issuer, uid)` identity is bound.

During PR #3870 review (greptile P1 "Identity commits before MFA"), a real MFA-bypass was found and fixed: for an **MFA-enabled** account, binding the SSO identity before the second factor completes is dangerous, because SSO logins are MFA-exempt — a just-bound identity would itself become an MFA-bypassing login path.

## What Phase 3 shipped (the safe stopgap)

`rodauth.login('password')` halts via Roda `:halt` and never returns, so we cannot inspect its result and *then* bind. Instead the bind is gated on a pre-bind MFA decision (`link_sso_second_factor_pending?`) that mirrors the `after_login` hook (`MfaStateChecker` + `DetectMfaRequirement.call(via_omniauth: false)`):

- **No second factor pending** → bind the identity, then log in. (Unchanged for default installs — `AUTH_MFA_ENABLED` is off by default.)
- **Second factor pending** → **skip the bind**, emit `mfa_required`, hand off to `/mfa-verify`.

This closes the vulnerability but leaves a gap.

## The gap this issue tracks

For an MFA-enabled account, the SSO identity is **never linked through the interstitial** — the bind is skipped and nothing completes it after the second factor succeeds. Consequences:

- The user re-hits the interstitial on their next SSO sign-in (or must link via the Phase-2 Connected Identities panel while authenticated).
- The feature is effectively inert for MFA accounts.

## Desired behavior

Complete the deferred `(provider, issuer, uid)` bind **after** the MFA second factor succeeds, so an MFA account that authenticates through the interstitial ends up linked exactly like a non-MFA account.

Sketch (needs design):
- Persist the pending bind (the challenge's `provider`/`issuer`/`uid` + `account_id`) into the partial MFA session state (`PrepareMfaSession`), and complete it in the MFA-success hook — reusing `bind_sso_identity` (which already returns `:ok`/`:conflict` with an ownership check).
- Preserve single-use / conflict semantics; the token is already consumed by the time MFA is reached.

## Test gap to close alongside

[...]

PR #3879, opened 2026-07-23
[#3840] Phase 4.A: complete deferred SSO identity bind after MFA (link-sso interstitial)

delano:
## Summary

[#3877](https://github.com/onetimesecret/onetimesecret/issues/3877)

The Phase 3 interstitial (PR #3870) correctly **skips** the `(provider, issuer, uid)` bind when the password login leaves a second factor pending — binding before 2FA would attach an MFA-exempt SSO login path to the account. But nothing ever completed the skipped bind, leaving the feature inert for MFA accounts: they re-hit the interstitial on every SSO sign-in. This PR completes the deferred bind after the second factor succeeds, so an MFA account that authenticates through the interstitial ends up linked exactly like a non-MFA account — and adds the `AUTH_MFA_ENABLED` integration harness #3877 called for, so the whole journey is covered end to end.

**How the deferred bind works:**

- **`Auth::Operations::DeferredSsoBind`** (new) owns the hand-off contract. `.defer` stashes the password-authorized bind tuple into the partial MFA session as a **string-keyed** payload (it must survive the JSON serialization the Redis session blob applies between the login request and the MFA-verify request). `.complete` consumes it **single-use** (deleted before the insert, so no outcome — including an exception — leaves a retryable stash), refuses to bind onto any account other than the one the password proved (`:mismatch`), and performs the insert via the shared `BindSsoIdentity` primitive (Phase 4.0). `:conflict`/`:mismatch` are audit-and-skip — MFA has already succeeded, so nothing may fail the login.
- **`routes/link_sso.rb`**: the deferred branch snapshots the challenge tuple and writes the stash inside the block passed to `rodauth.login('password')`. Rodauth yields that block **between `login_session` and `after_login`** — the only point where a session write both survives the login-time `clear_session` (the app's override destroys the previous session) and precedes the MFA hand-off. From there it rides the partial MFA session: `two_factor_authenticate` never clears the session, the same mechanism `awaiting_mfa` relies on.
- **`hooks/mfa.rb` `after_two_factor_authentication`**: completes the stash after OTP/recovery-code success, emitting `:sso_deferred_bind_completed` with the outcome. Wrapped in `ErrorHandler.safe_execute`; `:none` (one session-hash lookup, no DB access) for every login that didn't come through the interstitial's deferred branch.
- **`hooks/login.rb`** (non-MFA branch): self-heals a stash whose pre-login MFA prediction disagrees with the decision `after_login` just made (account factors changed between the two checks — otherwise the stash would never complete). The password verified moments ago and no factor is pending, which is exactly the authorization the direct-bind branch requires.

**Security invariants preserved:** the stash is not an authorization — it can only be written after the existing password verified and the single-use challenge token was consumed, and it lives in the server-side encrypted+HMAC'd session blob. No bind ever lands while a second factor is pending, so no MFA-exempt SSO path exists before the owner passes the second factor.

**The `AUTH_MFA_ENABLED` harness** (closing the test gap #3877 tracked):

- `Auth::Config` is one-shot per process (`auth-config-one-shot.md`), so the shared full-mode integration process — booted with MFA off — can never load the Rodauth OTP feature set. New rake lane **`spec:integration:full:mfa`** runs `apps/*/*/spec/integration/full_mfa` in a **fresh process** with `AUTH_MFA_ENABLED=true`, invoked automatically at the end of `spec:integration:full` so CI picks it up with no workflow changes (distinct `_mfa` results-file suffix so it never clobbers the parent lane's JSON output). Also exposed as `pnpm test:rspec:integration:full:mfa`.
- `integration/full_mfa/omniauth_signin_interstitial_mfa_spec.rb` drives the whole journey through real endpoints: TOTP provisioned via Rodauth's own two-phase JSON `otp-setup` (so the stored key shape — HMAC — matches production), then from a fresh unauthenticated session: SSO callback → interstitial → `POST /auth/link-sso` with the correct password → the same `mfa_required` body `POST /auth/login` emits, token consumed, **no identity row** (the PR #3870 Item 1 guard) → `POST /auth/otp-auth` with a valid code → the deferred bind lands (exactly one row, proven account, challenge's tuple, `:sso_deferred_bind_completed` with `:ok`). A second example locks in that a **failed** second factor binds nothing and does not lose the pending bind.
- `FullModeSuiteDatabase` now loads `date_arithmetic` on its SQLite branches too (previously PG-only) — Rodauth literalizes `Sequel.date_add` through the connection (the OTP last-use reuse guard), which raised on SQLite.
- The shared-harness interstitial spec's pending MFA example is removed; its comment points at the dedicated lane and the `DeferredSsoBind` unit spec. No pending examples remain for #3877's deferred-bind path.

## Related Issues

Closes the deferred-bind completion and its `AUTH_MFA_ENABLED` harness from #3877. Parent: #3840 (Phase 4), #3838.

## Test Plan

- **New MFA lane** `bundle exec rake spec:integration:full:mfa`: 2 examples, 0 failures — the full deferred-bind journey end to end (defer at password step → complete after OTP; failed OTP binds nothing and preserves the pending bind).
- **New unit spec** `apps/web/auth/spec/operations/deferred_sso_bind_spec.rb` (11 examples, 0 failures) against a real in-memory SQLite `account_identities` table with the real `(provider, issuer, uid)` unique index: string-key session contract via explicit JSON round-trip, single-use on every outcome (including when the insert raises), account binding, `:conflict` passthrough, idempotency, malformed-stash handling, `''` issuer sentinel.
- **`bind_sso_identity_spec.rb`** still green (11 examples, 0 failures).
- **Shared interstitial integration suite** (Valkey + full auth boot): 15 examples, 0 failures, 0 pending — the `rodauth.login('password') do ... end` block form leaves the non-MFA path (bind + login, single-use, rate limit, conflicts, tenant-surface refusal) unchanged.
- `rubocop --fail-level warning` clean on all changed files.

https://claude.ai/code/session_013h1Lzw1tAq66iSdoWZnFgm

### DIFF
--- a/apps/web/auth/config/hooks/login.rb
+++ b/apps/web/auth/config/hooks/login.rb
@@ -199,6 +199,35 @@ module Auth::Config::Hooks
           )
           session['awaiting_mfa'] = false
 
+          # Self-heal a DEFERRED SSO bind whose MFA prediction went stale
+          # (#3877). The link-sso interstitial stashes a deferred bind (via the
+          # rodauth.login block, i.e. BEFORE this hook) only when its pre-login
+          # MFA check said a second factor was pending. If the decision just
+          # made HERE disagrees (the account's factors changed in the window
+          # between the two checks), no second factor will ever fire to
+          # complete the stash — so complete it now: the password verified
+          # moments ago and no factor is pending, which is exactly the
+          # authorization the interstitial's direct-bind branch requires. For
+          # every other login the stash cannot exist (login_session destroyed
+          # the previous session before this hook) and this is a no-op (:none).
+          Onetime::ErrorHandler.safe_execute('complete_deferred_sso_bind', account_id: account_id) do
+            outcome = Auth::Operations::DeferredSsoBind.complete(
+              db: db,
+              session: session,
+              account_id: account_id,
+            )
+            unless outcome == :none
+              Auth::Logging.log_auth_event(
+                :sso_deferred_bind_completed,
+                level: outcome == :ok ? :info : :warn,
+                log_metric: true,
+                account_id: account_id,
+                outcome: outcome,
+                correlation_id: correlation_id,
+              )
+            end
+          end
+
           Onetime::ErrorHandler.safe_execute(
             'sync_session_after_login',
             account_id: account_id,
--- a/apps/web/auth/config/hooks/login.rb
+++ b/apps/web/auth/config/hooks/login.rb
@@ -215,7 +215,6 @@ module Auth::Config::Hooks
             outcome = Auth::Operations::DeferredSsoBind.complete(
               db: db,
               sid: session.id&.public_id,
-              session: session,
               account_id: account_id,
             )
             unless outcome == :none
--- a/apps/web/auth/config/hooks/login.rb
+++ b/apps/web/auth/config/hooks/login.rb
@@ -200,19 +200,21 @@ module Auth::Config::Hooks
           session['awaiting_mfa'] = false
 
           # Self-heal a DEFERRED SSO bind whose MFA prediction went stale
-          # (#3877). The link-sso interstitial stashes a deferred bind (via the
-          # rodauth.login block, i.e. BEFORE this hook) only when its pre-login
-          # MFA check said a second factor was pending. If the decision just
-          # made HERE disagrees (the account's factors changed in the window
-          # between the two checks), no second factor will ever fire to
-          # complete the stash — so complete it now: the password verified
-          # moments ago and no factor is pending, which is exactly the
-          # authorization the interstitial's direct-bind branch requires. For
-          # every other login the stash cannot exist (login_session destroyed
-          # the previous session before this hook) and this is a no-op (:none).
+          # (#3877). The link-sso interstitial stashes a deferred bind (a
+          # sid-bound SessionSidecar key written in the rodauth.login block,
+          # i.e. BEFORE this hook — #3858) only when its pre-login MFA check
+          # said a second factor was pending. If the decision just made HERE
+          # disagrees (the account's factors changed in the window between the
+          # two checks), no second factor will ever fire to complete the stash
+          # — so complete it now: the password verified moments ago and no
+          # factor is pending, which is exactly the authorization the
+          # interstitial's direct-bind branch requires. For every other login
+          # the stash cannot exist (login_session re-keyed the sid before this
+          # hook, and sidecar keys are sid-bound) and this is a no-op (:none).
           Onetime::ErrorHandler.safe_execute('complete_deferred_sso_bind', account_id: account_id) do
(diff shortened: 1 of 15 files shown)


delano:
Review feedback addressed in 23a27198 (plus ca1d38b6, which had already tightened the failed-OTP assertions). Disposition per point:

**Potential bugs**
1. *`full_mfa` tag derivation / `auth_db` availability* — verified: `auth_db` is defined in `apps/web/auth/spec/spec_helper.rb:533` and available to every auth spec; the explicit `:full_auth_mode` tag triggers `FullModeSuiteDatabase.setup!` via the config-level `before(:context, :full_auth_mode)` hook (tag-based, not path-based), confirmed by the lane running green. No change needed.
2. *`wrong_otp_code` termination* — adopted the counter-based construction: every candidate is six digits by construction and the loop is structurally bounded at `valid.size + 1` probes.
3. *`allow_immediate_otp_reuse!` subtlety* — documented: Rodauth stamps `last_use` only on **successful** validation (`otp_valid_code? && otp_update_last_use` short-circuits), so one rewind covers the failed-then-retry sequence.

**Code quality**
- *Rodauth login-block version sensitivity* — pinned in `link_sso.rb`: verified against rodauth 2.44.0 (`base.rb#login`), with a note that the `full_mfa` end-to-end example fails if an upgrade moves the block relative to `login_session`/`after_login` (the deferred bind would never land) — so the requested "test that would catch the block not firing" already exists; the comment now says so.
- *Double MFA-state read per interstitial POST* — documented as a deliberate tradeoff on `link_sso_second_factor_pending?`: the pre-login gate and `after_login`'s decision sit on opposite sides of Rodauth's hook boundary, the self-heal branch backstops the stale-prediction race, and caching the prediction would add coupling to save two indexed lookups on an interstitial-only path.

**Test coverage**
1. *String-typed `account_id`* — added a unit example locking the string-in/integer-authenticated direction of the `.to_s` coercion.
2. *Self-heal path* — documented as a known coverage gap in the unit spec header: the race cannot be opened deterministically from the integration lane (it would need `MfaStateChecker` to disagree with itself mid-request), so the branch is covered via the operation's contract examples plus the shared `safe_execute` wrapper. Also fixed the stale "harness pending" note there.
3. *Recovery-code path* — added end-to-end example (d): the deferred bind completes via `POST /auth/recovery-auth` too, locking in that `after_two_factor_authentication` is factor-agnostic. `provision_totp` now returns the auto-generated recovery codes the phase-2 setup response already surfaces.

**Security notes** — stash TTL on abandoned flows: correct as described (the stash dies with the partial-session TTL, no bind occurs, the user re-runs the SSO flow); moving the stash to a shorter, independent TTL is under separate discussion as a follow-up. The `ENV` load-time set and logger-fallback style points are accepted as-is per the review's own assessment.

Validation: MFA lane 3 examples / 0 failures, unit spec 12 examples / 0 failures, rubocop clean.

---

Issue #3889, opened 2026-07-24
[#3840] LinkSso: unmapped link_conflict/link_rate_limited codes render terminal failures as retryable password errors


delano:
Found during adversarial review of PR #3882 (Phase 4 mailbox-proof SSO linking). **Pre-existing in Phase 3 code** — not introduced by #3882, and deliberately left out of that PR to keep its scope tight.

## Problem

`src/shared/composables/useLinkSso.ts:79` (`resolveLinkErrorCode`) has no mapping for two error codes that `apps/web/auth/routes/link_sso.rb` actually emits:

- `link_conflict` (HTTP 409) — `link_sso.rb:183` and `:231`
- `link_rate_limited` (HTTP 429) — `link_sso.rb:264`

Neither status is in the resolver's `404/410` set nor its `401/403/422` set, so both fall through to `null`.

## Impact

`LinkSso.vue:143` branches on `errorCode !== 'invalid_token'`. With `errorCode === null`, a terminal failure takes the *wrong-password* path: the component clears the password field and refocuses the input (`LinkSso.vue:149-151`), inviting the user to retype a password against a challenge that can never succeed.

- **`link_conflict`**: raised when the account was re-emailed between challenge mint and POST, or when the `(provider, issuer, uid)` triple is already owned by another account. Unrecoverable by retry — but the UI presents it as retryable, under a generic message that does not say what went wrong.
- **`link_rate_limited`**: user gets a vague generic message instead of "too many attempts", is handed the focused password field back, and immediately re-triggers the throttle.

## Also stale

The composable's docstring at `useLinkSso.ts:33` still lists only `invalid_password` / `invalid_token` / `expired_token` / `link_expired` as the backend code set. That is now inaccurate against the route.

Note `expired_token` and `invalid_token` are frontend-side codes — grep confirms no Ruby in `apps/` or `lib/` ever emits either string.

## Suggested fix

Add `link_conflict` and `link_rate_limited` to `LinkSsoErrorCode` and to `resolveLinkErrorCode`, give each a distinct message, and make `LinkSso.vue` treat both as terminal (no password-field reset, no refocus) — mirroring how `useSsoLinkConfirm.ts` handles its terminal codes after #3882. Refresh the docstring while there.

## Reference

[...]

Issue #3886, opened 2026-07-24
Set-password affordance for passwordless accounts + password_auth_permitted policy flag


delano:
## Problem

Passwordless SSO accounts have **no way to set a local password**. `SecurityOverview.vue` hides the password/MFA/recovery cards entirely when `!has_password`, and `identities.rb` tells users to "Set a password first, then remove this identity" — but that instruction points at a screen that doesn't exist.

Separately, the gate needs a **policy axis**: today "no password" is the only reason to hide password management. SSO-*enforced* orgs/domains want it hidden even when a password could be set; self-serve consumer accounts want it offered. One signal can't express both.

## Current state (correct in shape)

Gating already keys off **credential-presence, not SSO membership** — this is right and matches Auth0/Clerk/WorkOS:

- `src/apps/workspace/account/settings/SecurityOverview.vue:138-147` — `PASSWORD_DEPENDENT_CARDS` filtered out when `!hasPw`
- `apps/web/auth/routes/identities.rb:105` — last-credential guard uses `rodauth.has_password?`
- `has_password` emitted by `apps/web/core/views/serializers/authentication_serializer.rb`, consumed via `hasPasswordOf` (`src/utils/features.ts:307`), contract `src/schemas/contracts/bootstrap.ts:567`

Gap: no "Set password" affordance for `!hasPw`, and no `passwordAuthPermitted` policy signal.

## Proposed logic (two independent axes)

```
showPassword = hasPassword                                  // change existing
   || (passwordAuthPermitted && !hasPassword)               // set new (recovery credential)
hidePasswordEntirely = !passwordAuthPermitted && !hasPassword   // SSO-enforced → "Managed by your IdP"
```

| State | Screen |
|---|---|
| has password | **Change password** (current pw required) |
| no password + `passwordAuthPermitted` | **Set password** (no current pw; mailbox-proof) |
| no password + `!passwordAuthPermitted` | hidden — existing `sso_managed_*` empty state |

Never gate on "is SSO." An account may hold a password **and** SSO identities (hybrid); both surfaces show.

## Changes

**Backend**
- [ ] Emit `password_auth_permitted` from `authentication_serializer.rb`. Sources: `authentication.mode == 'full'` AND (per-domain/org SSO-enforcement is OFF). Default `true` for consumer accounts.
[...]

Issue #3885, opened 2026-07-24
a11y sweep: auth surface repeats the live-region, focus-management and cursor defects fixed in the #3840 SSO views


delano:
## Context

While shipping #3840 Phase 4, two accessibility commits landed on the SSO linking views:

- `488c73a9f` — `SsoLinkConfirm.vue`: always-present live region, focus management on the terminal panel, `aria-disabled` → `aria-busy`, `role=group` + labelling, `cursor-pointer`, focus rings.
- (follow-up on the same branch) — the identical set ported to `LinkSso.vue`.

Both are in PR #3882. Auditing the rest of the session app afterwards showed these are **not** isolated to the SSO views — the same patterns are present across the auth surface, and two instances are more severe than the ones already fixed.

This issue collects all of it so it can be addressed in **one follow-up PR** rather than piecemeal.

Scope of the audit: all 11 views in `src/apps/session/views/` and all 14 components in `src/apps/session/components/`. Every line reference below was verified against the tree at `feature/3840-phase4`.

---

### 1. Focus is lost when a panel replaces the control that had it (P1)

**`views/AcceptInvite.vue`** — `handleAccept` / `handleDecline` set `actionResult`, which flips the `inviteState` computed. The panel holding the button the user just clicked is unmounted and replaced by the outcome panel at `AcceptInvite.vue:377`. There is no focus move, no `tabindex`, and no live region anywhere in the file. It then auto-redirects on a timer (`DIRECT_ACTION_REDIRECT_DELAY_MS`), so a keyboard or screen-reader user is navigated away without ever learning whether the invitation was accepted or declined.

This is the same defect fixed in `SsoLinkConfirm.vue` / `LinkSso.vue`, but worse: those dead-end panels are terminal and wait for the user; this one redirects on its own.

WCAG 2.4.3 (Focus Order) and 3.2.2 (On Input).

### 2. Live regions that can never announce (P1)

A live region must be in the DOM **before** its content changes. A region inserted with its text already present is not reliably announced.

**`views/MfaChallenge.vue:164` and `:211`** — each mode panel contains its own "mode active" announcement *inside* the `v-if` / `v-else` branch it describes:

```
149:  <div v-if="!useRecoveryMode" data-testid="mfa-otp-panel">
164:    <div aria-live="polite" class="sr-only">{{ t('web.auth.mfa.otp_mode_active') }}</div>
...
209:  <div v-else data-testid="mfa-recovery-panel">
[...]

Issue #3858, opened 2026-07-23
Store per-value session data as independent Redis keys with individual TTLs


delano:
### Summary

We'd like to explore storing individual session values as their own Redis keys with independent TTLs, rather than embedding all session state inside a single encrypted blob.

Currently the entire session is serialized, AES-256-GCM encrypted, HMAC-signed, and stored as one string under `session:<sid>` with a single global TTL (from `session_config['expire_after']`, default 24 hours). Every field — `authenticated`, `awaiting_mfa`, `locale`, `domain_context`, etc. — lives and expires together.

We want to evaluate whether specific short-lived or independently revocable values (e.g. an MFA-pending flag, a temporary domain context, an email-change token) should instead be stored as separate Redis keys with their own TTLs, decoupled from the main session blob.

---

### Motivation

- An MFA challenge should expire in minutes, not live for the full session lifetime.
- Flash values, pending-state flags, and transient UI context carry different risk profiles and lifetimes than the core auth claim.
- Fine-grained expiry enables targeted revocation of one capability without invalidating the entire session.

---

### What Already Exists (Proof of Concept)

Two patterns in the codebase already implement this approach:

1. **`session_metadata:<sid>`** — a `Familia::Horreum` sidecar written by `Onetime::Operations::Sessions::TrackMetadata` from inside `write_session`, with its own 30-day TTL, entirely outside the encrypted blob.
2. **Entitlement preview keys** (`SetEntitlementPreview`) — Redis SETs stored as siblings to the session blob (`session:<sid>:entitlement_preview_grants/revokes`, 24h TTL), with their key names stored as session fields that the middleware picks up.

Both are live and operational. The mechanism for the proposed feature is identical.

---

### Tasks

- [ ] Identify the candidate session values that would benefit from independent TTLs (e.g. `awaiting_mfa`, `authenticated_at` watermark, domain context, flash messages)
- [ ] Define the Redis key schema and naming convention for sidecar values
- [ ] Decide on encryption posture for each value (plaintext acceptable? or needs per-key cipher?)
- [ ] Update `delete_session` (and `delete_redis_sessions` in `ConfirmEmailChange`) to clean up sidecar keys alongside the main blob
- [ ] Update `find_session` or a dedicated middleware/hook to merge sidecar values into the session hash on read
- [ ] Validate orphan-TTL behaviour: sidecar TTL must be <= session blob TTL
- [ ] Add tests covering sidecar write, read, expiry, and delete

---

[...]

PR #3900, opened 2026-07-25
[#3840] SSO account-linking remediation: Phases 0–4 (integration)

### LINKED ISSUE #3859: [#3840] Connect-intent nonce not single-use: abandoned connect leaves live intent, next callback binds
## Summary

The SSO account-bound connect-intent nonce (`session[:sso_connect_intent]`), added in #3840 Phase 2 (commit `d8b724f3b`), was documented as a "single-use nonce" that "can never be replayed on a later callback." That is **false as implemented**. The nonce is cleared at exactly one site — the DELETE-on-read in `account_from_omniauth` — which only runs when a callback reaches that hook. An **abandoned** connect flow never reaches it, leaving the intent live for a subsequent, unrelated callback to consume and bind on.

The in-code comment has been corrected on `feature/3840-phase2-connect-sso-panel` to describe this gap honestly and link here. This issue tracks the behavioral fix + regression test.

## Location

`apps/web/auth/config/hooks/omniauth.rb`

- **Set** — request phase (`omniauth_request_validation_phase`, ~line 337): `session[:sso_connect_intent] = session_value`, only when `logged_in? && request.params['connect'] == '1'`.
- **Delete/consume** — callback (`account_from_omniauth`, ~line 99): `intent_account_id = session.delete(:sso_connect_intent)`.
- **No clearing** on `omniauth_on_failure`, on logout, or on a non-connect request phase.

Verified: `grep -rn "sso_connect_intent"` over `apps lib` returns only the set site, the delete site, and doc/spec comments — no other lifecycle handling.

## Failure / exploit scenario

1. Alice (logged in) opens the Connected Identities panel and initiates a connect → request phase sets `session[:sso_connect_intent] = <Alice account id>`.
2. Alice **abandons** at the IdP — cancels, the IdP errors, or she closes the tab. The connect's callback never reaches `account_from_omniauth`, so the DELETE never runs. Intent stays live.
3. On the SAME still-authenticated session, a **plain** SSO sign-in (`connect=0`) with a NEW, never-linked `(provider, issuer, uid)` reaches `account_from_omniauth`.
4. The callback reads `intent_account_id = session.delete(:sso_connect_intent)` = Alice's id; `logged_in?` is true; `intent_account_id == session_value` (still Alice) → `has_connect_intent = true`.
5. The arriving IdP identity **binds to Alice's account** — despite being a plain sign-in, not a connect.

This is exactly the shared-browser / second-tab bind the nonce is meant to prevent, gated behind the precondition "a dangling connect intent exists." Realistic trigger: a shared/public machine where Alice connect-then-abandons and walks away still logged in; Bob then does a plain SSO sign-in — Bob's identity is silently attached to Alice's account, and Bob subsequently signing in via SSO lands in Alice's account.

## Chosen fix: short-TTL sidecar key (built on #3858)

**Depends on #3858** (store per-value session data as independent Redis keys with individual TTLs). That mechanism gives this issue a strictly cleaner fix than any in-hook cleanup: store the connect-intent nonce as its own sidecar key with a **short TTL** (~5 min, one IdP round-trip) instead of a field inside the encrypted session blob.

Why the TTL sidecar is the right primitive here:

- **Self-correcting.** An abandoned connect (cancel / IdP-error / close-tab) needs *no* cleanup — the sidecar key just expires. It removes the dependency on a *subsequent* SSO flow passing through `omniauth_request_validation_phase` to sweep the stale nonce.
- **Genuinely single-use in time.** The TTL bounds the exposure window to one IdP round-trip unconditionally, whatever the next request is (or whether there is one). A request-phase clear only fires *if and when* another SSO flow happens to start.
- **No new branch in a security-sensitive hook.** Set-on-initiate + TTL, delete-on-consume. The lifecycle lives in the key, not in cleanup logic threaded across hooks.

### Plan (once #3858 lands)

1. Write `sso_connect_intent` as a sidecar key (e.g. `session:<sid>:sso_connect_intent`, `EX ~300`), value = the session account id — set in `omniauth_request_validation_phase` when `logged_in? && connect=1`.
2. Callback consumes it (GET + DEL) and binds only on session-account match, exactly as today. The one behavioral change: a **miss now means "expired"**, which is the correct default-deny — an abandoned intent can no longer be consumed by a later callback because it is gone by then.
3. **Encryption posture:** the value is an account id bound to the session — capability-adjacent. Default to encrypted per #3858's per-field sensitivity assessment; plaintext acceptable only if #3858 rules a sid-scoped key acceptable (a bare account id is low-value to an attacker with Redis access, since the bind still requires the matching authenticated session, but default to encrypted).
4. **Cleanup:** register the new prefix in #3858's `delete_session` / `delete_redis_sessions` sweep so logout doesn't orphan it. Sidecar TTL (~5m) is well under the blob TTL (24h), so #3858's orphan-TTL invariant holds trivially.
5. **Test:** abandoned-intent (now: TTL-expired) → plain sign-in → no bind — see Test gap below.

### Fallback: request-phase clear (if #3858 slips)

If #3858 is not available in time, clear stale intent at the **next request phase**, which every subsequent callback must pass through (it is where OmniAuth mints the `state` the callback validates against):

```ruby
# omniauth_request_validation_phase
if logged_in? && request.params['connect'].to_s == '1'
  session[:sso_connect_intent] = session_value
else
  session.delete(:sso_connect_intent)  # kill any dangling intent from an abandoned connect
end
```

This deterministically invalidates a stale nonce before a plain sign-in's callback can consume it — covering cancel, IdP-error, **and** close-tab uniformly, without coupling to OmniAuth internals. It closes the same hole as the TTL sidecar, but only reactively (on the next SSO flow) rather than unconditionally on a timer. It is also what the current in-code `TODO(#3859)` points at. Prefer the TTL sidecar when #3858 is available; a failure-path-only delete on `omniauth_on_failure` is the weakest option (covers only the error-callback case, not close-tab) and is not the plan of record.

## Test gap

No test covers abandoned-intent → plain-sign-in → no-bind. Add a regression test to `apps/web/auth/spec/integration/full/omniauth_connect_link_spec.rb`:

- Log in as Alice; run the request phase with `connect=1` (sets intent); do NOT complete that callback.
- Simulate abandonment: with the TTL sidecar, let the key expire (or delete it directly to model expiry); with the fallback, run a non-connect request phase to trigger the sweep.
- Run a plain sign-in callback (`connect=0`) with a new `(provider, issuer, uid)`.
- Assert the new identity is NOT bound to Alice (falls through to the email / H-3 branch), and the intent is gone.

### LINKED ISSUE #3877: [#3840] Phase 4: complete deferred SSO identity bind after MFA (link-sso interstitial)

delano:
Integration branch for the #3840 SSO account-linking remediation. Phases 0 through 4, each already reviewed and merged individually into this branch.

## How to review this

**Don't re-read the 15k-line diff.** Every line here except the `main` merge already went through its own PR:

| PR | Phase | What it did |
|---|---|---|
| #3845 | 0 | Issuer-scoped SSO identity key — closes the cross-tenant takeover (#3838 item 5) |
| #3844 | 1 | Trusted-IdP email-linking flag + boot guard (unblocks #3836) |
| #3851 | 2 | Connected Identities: connect-provider panel + session-bound SSO linking |
| #3874 | 2 | Connect-intent nonce made single-use via short-TTL sidecar key (#3859) |
| #3870 | 3 | SSO sign-in interstitial (LinkSso challenge flow) |
| #3878 | 4.0 | Extract the `BindSsoIdentity` shared bind primitive |
| #3879 | 4.A | Complete the deferred SSO identity bind after MFA |
| #3882 | 4 | Mailbox-proof SSO linking for passwordless accounts |

**The only content not previously reviewed in a PR is the `origin/main` merge** (`bc4b999b3`). `main` had moved 55 commits ahead (admin-ops expansion #3897, compose pin #3894). Exactly one conflict, in `apps/web/auth/config/hooks/mfa.rb`, and it was **comment-only** — both sides independently arrived at `session['awaiting_mfa'] = false` with the string key, so no code was chosen or discarded. The two rationales were merged into one comment:

- **String key (#3854, from main):** `PrepareMfaSession` writes the string key and `BaseSessionAuthStrategy` reads it — a symbol `:awaiting_mfa` would silently never match. `SyncSession` deletes both key forms but runs inside `safe_execute`, so on a sync failure this line is the only remaining clear.
- **`false` rather than a delete (#3858, from this branch):** the sidecar commit treats a falsy `awaiting_mfa` as a DELETE (`absent_when_falsy`), so on success the field converges to absent everywhere. If the sidecar commit fails, `write_session`'s rescue keeps this `false` in the blob, where blob-wins outranks the stale `true` and the next healthy commit heals it. A delete could not win that conflict — the blob would carry nothing, the stale `true` would re-merge with a fresh TTL every request, and the session would be locked out of every gated route indefinitely.

That one file is the place to focus a re-read.

## The security property this establishes

Email may **locate** an account; only a credential may **bind** one. An IdP that emits a victim's address cannot reach the victim's account:

- **Authenticated connect** routes by session, never by email — the session is the authorization. An IdP asserting a different account's email still binds only to the actor's own session account.
- **Unauthenticated first contact** with an existing account hits the H-3 refusal and is routed to an explicit, credentialed link flow (interstitial, or mailbox proof for passwordless accounts) rather than auto-linked.
- **The one sanctioned exception** is the opt-in, per-provider, boot-guarded trusted-IdP flag from Phase 1, for single-tenant operators who control both OTS and the IdP.
- **Identities are issuer-scoped**, so `(provider, uid)` collisions across tenants can no longer cross-bind.

## Testing

CI lanes on this PR are the gate. Local auth-integration specs were not run for the merge commit — the test datastores (Valkey 2121, PG 2132) were down, and the conflict resolution is comment-only. `ruby -c` and RuboCop pass.

The suite added across these phases covers issuer-scoped identity, connect/link binding, the no-intent and expired-intent refusals, surface isolation, the sign-in interstitial (including an MFA lane), and mailbox-proof confirm. Notably `omniauth_connect_link_spec.rb` scenario 1b pins that a connect succeeds when the IdP asserts **no email claim at all**.

## Issues

Closes #3877
Closes #3859

Parent #3840 is left open deliberately — its explicitly deferred children (#3849, #3856, #3885, #3886) are out of scope here. #3836 is unblocked by Phase 1 but not auto-closed; its Bug A shipped separately as #3848.

Known follow-ups, none blocking:

- **#3889** — real user-facing defect: the frontend renders `link_conflict` / `link_rate_limited` as retryable password errors.
- **Unfiled** — per-address issuance throttle on the mailbox-proof email. Phase 3 has `check_login_rate_limit!` in `link_sso.rb`; Phase 4 issuance has none. #3882's own body flags this as worth a rate-limit follow-up.
- **#2729** — still owes sweep S1 (fold-pair matrix) and F6 (DisableMfa guard), plus the 4 pending OIDC-callback E2E specs from #3882 that need a live-discovery boot harness.
- **#3499** — first-contact / JIT-create for email-omitting IdPs remains schema-blocked (`login_column :email`, `accounts.email NOT NULL` + `valid_email` CHECK, no `resolve_omniauth_email`). The connect path already works without an email claim; see the state-of-play comment on that issue.


delano:
@greptile-apps pls review despite the many files

delano:
## Review feedback resolutions (a7f0f30..fe07f6d)

All outstanding items from the Copilot, Claude, and greptile reviews are addressed in five commits:

| Item | Resolution | Commit |
|---|---|---|
| **greptile P1** — issuerless identities cross tenant boundaries (`features/omniauth.rb:146`) | **Confirmed.** Tenant surfaces can configure GitHub/Google (`SsoConfig::PROVIDER_TYPES`), both resolve to the shared `''` issuer sentinel, and the step-1 exact lookup has no `platform_path` gate — a tenant callback could authenticate a platform/other-tenant row and additively org-join it. Fixed fail-closed: `refuse_issuerless_on_tenant?` rejects issuerless callbacks on tenant surfaces before any lookup (blocks the match *and* JIT-creation of colliding `''` rows). Platform issuerless SSO and tenant OIDC/Entra unaffected. **Consequence: tenant GitHub/Google SSO is intentionally disabled** — no per-tenant issuer exists to isolate on. Follow-up to drop them from the tenant provider options: #3902. Regression specs pin the collision and the refusal (48 callback-suite examples green). | 1de51b46d6 |
| **Rate-limit gap** on `sso-link-confirm` POST (acknowledged follow-up in PR body) | Canonical `LoginRateLimiter` gated **before** token consumption, keyed `sso-link-confirm:<client-ip>` (a guessed token resolves to no record, so there's no email to key on; per-IP subject avoids a global-bucket DoS). Only `:link_expired` counts as a guess; success clears. 429 `confirm_rate_limited` + `Retry-After`, wired through `useSsoLinkConfirm` with wait-not-dead copy. | 7c2fa69a0e |
| **`json_body_params` falsy loss** (Copilot r3649712258, Claude-confirmed) | `Hash#key?` presence check — explicit JSON `false`/`0`/`null` now wins over form params; current routes behaviorally unchanged. Unit spec pins precedence + rewind. | fe07f6d722 |
| **`last_credential` 409 guidance** (Copilot r3649817302) | "Set a password first" pointed at UI passwordless accounts never see. Reworded (backend + locale) to the two paths that actually exist: connect another provider (same panel, lifts the guard) or the signed-out Forgot-password flow (verified: Rodauth `set_password` inserts a hash when none exists; SSO-created accounts are open). Spec pins the copy. | 4444b4ec6e |
| **Locale `content_hash` contradiction** (Copilot r3649712270, Claude-confirmed) | Stale "intentionally omitted" sentence dropped from the context; the hash it contradicted is the script-generated one. | 7de07d2696 |

Notes for merge: two locale keys are intentionally bare / hash-dropped (`web.sso_link_confirm.errors.confirm_rate_limited` new, `web.auth.connections.errors.last_credential` edited) — `locales:hashes` should run as its own pass since it rewrites hashes tree-wide. The `link_sso_second_factor_pending?` double-read note from Claude's review is an acknowledged tradeoff, no change. Combined verification: 35 integration + 19 unit examples green locally across all touched suites.

PR #3933, opened 2026-07-28
Add colonel per-domain config management

delano:
Adds colonel-side management for the seven per-domain config kinds. `CustomDomain::ConfigRegistry` becomes the single catalog of kinds and their field specs (specs now live on the config models themselves), backed by new upsert/delete/ensure operations and internal API endpoints.

Admin UI: a domain-configs section on `AdminDomainDetail`, split into row / modal / dialog / field-grid child components, driven by new `useAdminDomains` store actions and Zod response schemas.

Also includes migration `20260727_01_backfill_signin_config`, which restores sign-in on custom domains created before v0.26 — `SigninConfig` started defaulting sign-in OFF with no backfill, silently disabling it fleet-wide.

Coverage: tryouts for the new endpoints plus a registry drift guard that fails when model field specs and the registry diverge; vitest for the store and the section components. Colonel session-admin specs reorganized under `docs/specs/colonel-ui/`.

Note: `locales/content/en/admin-domains.json` is new and its `content_hash` values are not populated — `locales:hashes` rewrites ~2000 hashes tree-wide, so that belongs in its own commit rather than bundled here.


delano:
Feedback pushed to feature/colonel-orgs

PR #3957, opened 2026-08-01
feat(branding): extract BrandMark and resolve install logo on disabled homepage

delano:
Stacked on #3953 (`feature/brand-logo-dark-url`) — review that one first. This PR
adds only the two commits above the fork point at `f0038a892`.

## What this does

Extends the operator install logo (`BRAND_LOGO_DARK_URL`, added in #3953) to the
disabled-homepage variants, and extracts the light/dark rendering into a shared
component instead of a third copy of the same `` pair.

### `BrandMark.vue` (new)

MastHead and both disabled-homepage variants each needed the same logic: render
the light logo, swap to the dark one via class-based dark mode, fall back through
alt-text sources, recover on load error. That's now one component rather than
three near-identical template blocks.

### Logo resolution in `useDisabledConfig`

Resolves through tenant logo → install logo → branded monogram → neutral keyhole
mark, with one guard worth calling out: **the install logo is never surfaced on a
custom domain.** A tenant's branded page showing the operator's mark would leak
one customer's identity onto another's page. The install-only companions
(`logoDarkUri`, `logoAlt`) are likewise suppressed when a tenant logo is active,
so they can't attach themselves to the wrong image.

## Tests

145 passing across `useDisabledConfig`, `DisabledVariantsLogo`, `BrandMark`, and
`MastHead`. The 8 new `useDisabledConfig` cases cover tenant-over-install
precedence, the custom-domain identity-leak guard, operator alt text, and
dark-logo pairing (a dark logo with no light logo to pair with resolves to null
rather than rendering alone).

## Merge order

Depends on #3953. If that merges to `main` as a squash, the shared commits leave
`main`'s history and this branch needs a rebase onto `main` rather than a base
retarget.

### DIFF
--- a/src/apps/secret/views/disabled/useDisabledConfig.ts
+++ b/src/apps/secret/views/disabled/useDisabledConfig.ts
@@ -82,8 +82,29 @@ export interface DisabledHomepageProps {
    * radii instead of identityStore's DEFAULT_CORNER_CLASS fallback.
    */
   cornerClass: string | null;
-  /** Tenant-uploaded logo URL; falls back to monogram when null. */
+  /**
+   * Logo URL: the tenant-uploaded logo, falling back to the operator's
+   * install-wide logo (brand.logo_url) on the canonical site. Deliberately
+   * NOT identityStore's logoSource — that ends in the DEFAULT_LOGO_COMPONENT
+   * sentinel, while variants want URL-or-null so their own monogram/keyhole
+   * fallbacks still apply. Null when neither is configured.
+   */
   logoUri: string | null;
+  /**
+   * Dark-theme variant of the install logo (brand.logo_dark_url). Non-null
+   * only while the install logo is the asset being rendered (the store nulls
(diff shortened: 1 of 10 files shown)


delano:
Addressed in `762a518` — `BrandMark.vue` now tracks dark-logo load errors in their own `logoDarkError` ref, resets it when `logoDarkUri` changes, and gates both the dark `` (`v-if="hasUsableDarkLogo"`) and the light img's `dark:hidden` class on that flag. A dark asset that 404s now degrades to the working light logo in dark mode instead of rendering nothing.

Coverage: 7 new cases in `BrandMark.spec.ts` (19 passing, was 12) plus one MastHead integration case. Full run: 5 files / 204 tests passing.

One scenario left as-is deliberately: light errors while dark is fine still falls through to the fallback slot, since both imgs sit inside `<template v-if="hasUsableLogo">`. That is the designed degradation rather than broken branding; pinned with a test.

Resolving the two review threads — both were filed against `6a59563`, before the fix.

PR #3959, opened 2026-08-01
feat(billing): admin-initiated checkout links + unified Stripe automatic-tax policy

delano:
Adds admin-initiated Stripe checkout links: a `CreateCheckoutLink` operation with colonel API, CLI (`checkout-links create`), and a modal on the colonel user detail page.

Consolidates Checkout::Session param assembly into a single builder (`CreateCheckoutLink.build_session_params`) that the self-serve controller now delegates to, eliminating the duplicated param construction in `billing.rb`.

Introduces deployment-level `STRIPE_AUTOMATIC_TAX` config (following the `STRIPE_CHECKOUT_HOST` precedent): a shared `apply_tax_policy!` drives `automatic_tax`, `tax_id_collection`, `billing_address_collection`, and `customer_update` identically across all three checkout-creation paths (admin op, `billing.rb`, `plans.rb`).

Scope of that unification, to be precise: `plans.rb#checkout_redirect` shares the **tax block only**. It still assembles its own `session_params` and keeps this surface's historical subscription metadata shape (`debug_info` JSON + `customer_extid`, no `orgid`), which is webhook-visible and deliberately unchanged here; `checkout_completed#find_target_organization` falls back to `stripe_customer_id` and then the default org when `orgid` is absent. Only `billing.rb` and the admin op share `build_session_params` and therefore the metadata shape.

The per-link `--tax`/`enable_tax` flag was removed — tax treatment is deployment policy, not operator choice.

Deploy note: `plans.rb#checkout_redirect` previously hardcoded `automatic_tax: enabled` — production must set `STRIPE_AUTOMATIC_TAX=true` to retain tax collection on `/plans` checkouts.

## Review round (2026-07-31)

- **Duplicate-subscription guard now covers admin links.** The predicate moved out of `Billing::Controllers::Base` into `Billing::SubscriptionGuard`, and `CreateCheckoutLink` applies it before plan resolution and before the dry-run branch. A support-issued link for an org with a live, non-canceling subscription would otherwise have created a second chargeable subscription and let the webhook overwrite `org.stripe_subscription_id`. No operator override: the legitimate second-checkout cases (`cancel_at_period_end`, pending currency migration) are already exempt inside the predicate.
- **`details.region` is never null.** `BillingConfig#region` is nil on deployments that are not region-scoped; the colonel logic now maps that to `UNSCOPED_REGION` (`'global'`) at the API boundary. The Zod schema stays `z.string()` so a regression fails loudly.
- **`plans.rb` requires the op explicitly** instead of relying on `controllers.rb` load order.


delano:
Addressed across `46b51a9`, `298e23a8`, `a860094d`. Two findings were real, two did not hold up as reported — detail below.

**Active-subscription guard (P1, real).** Both self-serve paths called `org_has_blocking_active_subscription?`; the operation did not. The predicate and its Stripe-failure fail-safe now live in `Billing::SubscriptionGuard` (`apps/web/billing/lib/subscription_guard.rb`), since the operation has no request and cannot include `Controllers::Base`. `Base#org_has_blocking_active_subscription?` is a two-line delegator; the operation runs the guard before plan resolution and before the dry-run branch, so `--dry-run` reports the block too.

No operator override, deliberately. The two cases where a second checkout is legitimate — `cancel_at_period_end` and a pending currency migration — are already exempt inside the predicate. Anything an override would additionally unlock is exactly the double-charge state, since there is one `org.stripe_subscription_id` and the webhook overwrites it. Live-subscription plan changes belong in the plan-change flow.

**`details.region` nil (P1, real).** Fixed Ruby-side with `UNSCOPED_REGION = 'global'` rather than relaxing the schema to `.nullable()`. The modal renders `region` into a labelled `<dd>`, so null would show an empty cell under a "Region" label; keeping `z.string()` strict means a genuine backend regression still fails loudly.

**Missing require — no runtime bug.** `controllers.rb` requires `controllers/plans` then `controllers/billing`, and `controllers/billing.rb:13` requires the operation, so the constant is always defined before any request reaches `checkout_redirect`. There is no autoloader. Added the explicit `require_relative` anyway to remove the implicit load-order dependency.

**Metadata unification — claim was accurate as written.** The PR body said `apply_tax_policy!` drives `automatic_tax`, `tax_id_collection`, `billing_address_collection` and `customer_update` identically across all three paths, which holds. The underlying observation about `plans.rb` metadata differing is correct, but that metadata is webhook-visible and `checkout_completed#find_target_organization` already falls back through `stripe_customer_id`; unifying it was judged riskier than the drift. Added a scope note instead.

Tests: 21 operation examples (+6 guard regressions), 9 colonel examples (+1 nil-region regression), 305 controller examples, 14 vitest — all green. The one failure in the wider billing lane (`catalog_validate_command_spec.rb:36`) is pre-existing and unrelated: it asserts a generated schema file that `generated/.gitignore` excludes until `schemas:json:generate` runs.

PR #3999, opened 2026-08-05
feat(cli): add customers role reconcile to repair role-index drift

### LINKED ISSUE #3974: customers role list omits accounts whose role field is colonel
## Problem Statement

`bin/ots customers role list` can omit accounts whose `role` field is actually `colonel`, under-reporting who holds site-admin privileges.

Observed on a fresh database with two colonel accounts: `role list` printed only one and reported `Total: 1`, while `bin/ots customers show <other-account>` reported `role: colonel` for the account that was missing from the listing.

## Early Diagnosis

Authorization and this listing read from two different sources, and they have diverged.

- Authorization reads the field directly: `has_system_role?` checks `cust.role == 'colonel'` (`authorization_policies.rb:60`).
- The CLI listing reads the index: `Customer.find_all_by_role(target_role)` (`role_command.rb:141`), backed by the Familia `multi_index` set `customer:role_index:colonel` (`role_index.rb:27`).

The listing applies no `verified` filter — it prints verification state as a label but does not exclude on it (`role_command.rb:143-147`) — so filtering is not the explanation. That points at the index genuinely missing an entry.

`RoleIndex::InstanceMethods#save` (`role_index.rb:57`) is what maintains the set, and its own comment notes `multi_index` auto-adds on save but does not auto-remove from the old role's index, so removal is handled explicitly. Suspect some write path sets `role` without going through that override — a direct field write or a fast-writer that bypasses `save`. Not yet root-caused; I have not traced which path was responsible for the account observed here.

Impact is on trustworthiness of an audit command rather than on access control — the index does not grant anything, so a stale entry cannot escalate. But "list all site admins" silently returning an incomplete answer is a bad property for a command an operator would reach for during a security review. `colonel_count` (`role_index.rb:48`) reads the same set and would under-report identically.

## Affected Areas

- [ ] `lib/onetime/models/customer/features/role_index.rb` — `save` override that maintains `customer:role_index:*`; `find_all_by_role`, `colonel_count`, `list_colonels`
- [ ] `lib/onetime/cli/customers/role_command.rb` — `list_customers_by_role` reads the index
- [ ] `apps/web/auth/operations/customers/set_role.rb` — canonical role-write path, to confirm it goes through `save`
- [ ] `lib/onetime/application/authorization_policies.rb` — reads the field, establishing the field as the authority

## Notes

Worth deciding explicitly which store is authoritative. The field is what authorization honors, so the index should be treated as a derived cache — which argues for a reconcile/rebuild path so a drifted index can be repaired rather than silently trusted.

mahdi13830510:
## Summary
- Root cause confirmed: `customer:role_index:*` (Familia `multi_index`) is only kept in sync automatically on the full `Familia#save` path (`auto_update_class_indexes`, called from `persist_to_storage`). Targeted writers — `save_fields`, `multi_field_update`, `multi_field_fast_write` — persist scalar fields directly and never call it, so a customer's `role` field and its index membership can drift apart silently. Reproduced locally: a customer set to `role = 'colonel'` via `save_fields(:role)` never appears in `find_all_by_role('colonel')`, even though `customer.role` correctly reads `'colonel'` (and authorization, which reads the field directly, is unaffected).
- Adds `bin/ots customers role reconcile`, which calls Familia's existing (already-generated but previously unexposed) `Customer.rebuild_role_index` to rebuild the index from the authoritative `role` field on every customer record.
- This directly addresses the issue's own conclusion: the index is a derived cache and needs "a reconcile/rebuild path so a drifted index can be repaired rather than silently trusted."
- Does not affect authorization (`has_system_role?` reads the field directly, never the index) — this only restores trust in `role list` / `colonel_count` as an audit tool.

Closes #3974

## Test plan
- [x] New tryout section in `try/unit/cli/customers/role_command_try.rb` reproduces the drift via `save_fields(:role)` and asserts `rebuild_role_index` repairs it
- [x] `bundle exec try try/unit/cli/customers/role_command_try.rb try/unit/models/v2/customer_role_index_try.rb` — 78 passed
- [x] `bundle exec rake try:unit` (full unit tryout suite) — 7364 passed, 0 failed
- [x] `bundle exec rubocop` on changed files — no new offenses
- [x] Manually ran `bin/ots customers role reconcile` end-to-end

### DIFF
--- a/lib/onetime/cli/customers/role_command.rb
+++ b/lib/onetime/cli/customers/role_command.rb
@@ -11,6 +11,7 @@
 #   bin/ots customers role demote user@example.com               # Demote to customer
 #   bin/ots customers role list                                  # List all colonels
 #   bin/ots customers role list --role admin                     # List users with specific role
+#   bin/ots customers role reconcile                              # Rebuild the role index from source
 #
 # The actual role mutation + admin audit event is performed by the shared
 # Auth::Operations::Customers::SetRole op (single implementation); this command
@@ -27,7 +28,7 @@ module Onetime
       argument :action,
         type: :string,
         required: true,
-        desc: 'Action to perform: promote, demote, or list'
+        desc: 'Action to perform: promote, demote, list, or reconcile'
 
       argument :email,
         type: :string,
@@ -59,9 +60,11 @@ module Onetime
           demote_customer(email, force)
         when 'list'
           list_customers_by_role(role)
+        when 'reconcile'
+          reconcile_role_index
         else
           puts "Unknown action: #{action}"
-          puts 'Valid actions: promote, demote, list'
+          puts 'Valid actions: promote, demote, list, reconcile'
           exit 1
         end
       end
@@ -148,6 +151,20 @@ module Onetime
 
         puts '-' * 40
         puts "Total: #{customers.size}"
+      end
+
+      # Rebuilds customer:role_index:* from the authoritative `role` field on
+      # every customer record. The index is a derived cache -- targeted
+      # writers (save_fields, multi_field_update, multi_field_fast_write)
+      # persist `role` without going through the full Familia#save path that
+      # maintains the index, so a customer's role and its index membership
+      # can drift apart silently (#3974). This does not affect authorization
+      # (authorization_policies.rb reads the field directly), only the
+      # trustworthiness of `role list` / `colonel_count` as an audit tool.
+      def reconcile_role_index
+        puts 'Rebuilding customer role index from source (customer.role field)...'
+        processed = Onetime::Customer.rebuild_role_index
+        puts "Done: #{processed} customers scanned."
       end
 
       def validate_email_provided!(email, action)
--- a/try/unit/cli/customers/role_command_try.rb
+++ b/try/unit/cli/customers/role_command_try.rb
@@ -369,6 +369,36 @@ demote_customer(@colonel_email)
 reset_role(@colonel_email, 'colonel')
 #=> "colonel"
 
+# -------------------------------------------------------------------
+# Reconcile: rebuild_role_index repairs drift between the role field
+# and the index (#3974). Targeted writers (save_fields et al.) persist
+# `role` without going through the full Familia#save path that
+# maintains customer:role_index:*, so the two can drift apart silently.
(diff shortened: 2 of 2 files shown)


delano:
@mahdi13830510 Sending this one back. Main bumped familia to 2.12 (`f580fa7`), and under 2.12 `save_fields` on a class-indexed field routes through `add_to_class_*`. So the drift your tryout asserts (`#=> false`) no longer occurs, and the code comment explaining the mechanism is false on current main. Please rebase on main and re-verify the symptom in #3974 can still be reproduced. 

Two more things: the rebuild is destructive (it DELs the index before repopulating), so the command should default to dry-run and report what would change; only mutate when passed `--run`. And the tryouts test Familia's rebuild_role_index directly rather than the CLI command this PR adds; the command itself has no coverage.

delano:
Thanks @mahdi13830510 — closing in favor of #4175, which carries this forward with you credited as co-author.

Two findings from re-verifying this against current `main`, one in your favor:

**Your premise was right.** The review concern — that familia 2.12's routing of `save_fields` through `add_to_class_*` made the drift obsolete — turns out to hold only for `unique_index`. For `multi_index` (which `role_index` is), `add_to_class_role_index` is add-only by design: familia's own source says prior buckets are "intentionally retained on a value change." Verified empirically against 2.12.0 with a live Redis: after `save_fields(:role)`, the customer sits in both the old and new buckets. There's also a second drift source nobody had named: RTBF's 365-day TTL expires the customer hash while its index member persists forever, permanently inflating `colonel_count`. So #3974 stays real on current `main`.

**But this branch couldn't be rebased.** It's 557 commits behind, and its key tryout assertion encodes the 2.11.2 behavior (`save_fields` not touching the index at all), which now fails under 2.12 — the mechanism comment would also have baked the wrong explanation into the tree. Beyond staleness, the replacement addresses the standing review requests: dry-run default with `--apply`/`--force`/`--json` (matching `org reconcile`), logic in a shared op rather than the command, and incremental SADD/SREM instead of `rebuild_role_index`'s DEL-then-repopulate — so a mid-run crash can never leave the index worse than it started.

The reconcile idea was the right call; #4175 exists because of this PR.

---

Issue #3478, opened 2026-06-15
Unable to login with SSO/EntraID when the user doesn't have email address


nv-pipo:
## Summary

When using SSO/EntraID, login hangs when if user doesn't have an email address in their profile. The UI just freezes on a loading screen, which is a poor user experience and makes it look like the app has crashed.

### Impact

Users without emails can't use the product at all, and the "frozen" UI looks like a crash. Onetimesecret should either show an error message or successfully log in without an email, rather than freezing.

## Steps

1. Login via SSO/EntraID with an account that has no email set.
2. The auth flow gets stuck.

### Expected vs Actual

* **Expected:** The app should either show an error message indicating that an email is required, or it should allow the user to log in without an email (falling back to a UPN or other identifier).
* **Actual:** Browser just hangs indefinitely.

### Additional context

<img width="1926" height="1204" alt="Image" src="https://github.com/user-attachments/assets/4df72eb3-1e2d-426b-a68d-b0855c5d4c73" />

delano:
@nv-pipo Thanks. I have a pretty good idea what's happening. I want to make sure I can replicate it reliably and then I'll include a fix for v0.25.11.

delano:
@nv-pipo v0.25.11 has the interim fix which should address the confusing UX with an appropriate error message. Please let me know if that works as intended. 

Addressing the underlying issue has a number of implications that will require a full release cycle. Added #3499 to document it.

delano:
@nv-pipo Haven't forgot about this. We made a bunch of progress with SSO and should be able to tie this one up soon. I don't think it'll make it into v0.26.3 but most likely in v0.26.4.

PR #4000, opened 2026-08-05
fix(auth): mirror accounts.valid_email CHECK in the SSO email guard

### LINKED ISSUE #3971: SSO callback 500s on email claims the accounts.valid_email CHECK rejects but split('@') accepts
## Problem

`before_omniauth_create_account` (`apps/web/auth/config/hooks/omniauth.rb`) validates the IdP email claim structurally — `split('@')` must yield exactly two non-empty parts — but the `accounts.valid_email` CHECK it is protecting is stricter:

```ruby
# apps/web/auth/migrations/001_initial.rb:27
constraint :valid_email, email: /^[^,;@ \r\n]+@[^,@; \r\n]+\.[^,@; \r\n]+$/
```

Every shape the CHECK rejects but the guard accepts reaches the INSERT and raises `Sequel::CheckConstraintViolation` → **500 on the SSO callback**, which is the frozen-loading-screen failure #3478 exists to prevent.

Probed against `main` (PostgreSQL), all 500:

| claim | why the CHECK rejects it |
|---|---|
| `al ice@contoso.com` | internal space |
| `al,ice@contoso.com` | comma in local part |
| `alice@cont;oso.com` | semicolon in domain |
| `alice@contoso` | dotless domain — plausible from internal AD FS / OIDC issuers |

## Context

Found while resolving Copilot feedback on #3965, which closed one instance of this class: a padded-whitespace claim (`"  alice@contoso.com  "`) 500'd the same way. That was fixed by trimming in the `omniauth_email` accessor (ca5ae1fcb). The remaining shapes were left out of that PR deliberately — tightening the guard changes signup validation semantics for a class of addresses, so it wants its own change and its own review. A comment at the guard records the gap.

## Suggested fix

Make the guard mirror the CHECK rather than merely counting `@`, and redirect to `/signin?auth_error=invalid_email` as the existing structural branch does.

Two things to get right:

- **Duplicate the pattern in Ruby, don't lean on the database.** The CHECK is created only on PostgreSQL (`001_initial.rb:29` stores a plain `String` otherwise), so a guard that relied on the DB raising would silently stop guarding on SQLite installs — and the SQLite CI leg would stop catching regressions of this kind.
- **Anchor with `\A`/`\z`, not `^`/`$`.** Ruby's `^`/`$` match at line boundaries, so `"ok@example.com\nbad"` would pass a `^`/`$` guard here and still be rejected by PG (whose `^`/`$` are string anchors), reintroducing the 500.

## Test note

Assert on `accounts.email` directly, not via `Onetime::Customer.email_exists?` — the Customer index is written by `CreateCustomer`, which normalizes the address first, so it holds the clean form whether or not the account row does. And note this file's primary CI leg is SQLite (`lib/tasks/spec.rake:129-140`), which has no CHECK, so a test that relies on the constraint firing will pass there with the bug present.

mahdi13830510:
## Summary
- `before_omniauth_create_account` (`apps/web/auth/config/hooks/omniauth.rb`) only validated the IdP email claim by counting `@` characters, but the `accounts.valid_email` CHECK it's protecting (`apps/web/auth/migrations/001_initial.rb:27`) is stricter. Every shape the CHECK rejects but the old guard accepted reached the INSERT and raised `Sequel::CheckConstraintViolation` → 500 on the SSO callback — the same frozen-loading-screen failure class #3478 exists to prevent.
- Adds `Onetime::SignupValidation.structurally_valid_email?`, which duplicates the CHECK's regex in Ruby rather than relying on the database to raise: the CHECK is PostgreSQL-only (SQLite stores email as a plain String per that same migration), and SQLite is this file's primary CI leg, so a guard that only counted `@` would leave that leg blind to a regression of this kind.
- Anchored with `\A`/`\z`, not `^`/`$` — Ruby's `^`/`$` match at line boundaries, so `"ok@example.com\nbad"` would slip past a `^`/`$`-anchored guard and still be rejected by PG (whose `^`/`$` are string anchors), reintroducing the 500.
- Fixes the four repro shapes from the issue: internal space in local part, comma in local part, semicolon in domain, dotless domain (plausible from internal AD FS / OIDC issuers).

Closes #3971

## Test plan
- [x] New cases in `try/unit/signup_validation_try.rb` cover all four repro shapes plus empty/no-`@`/blank-local-part/embedded-newline edge cases
- [x] `bundle exec try try/unit/signup_validation_try.rb` — 24 passed
- [x] `bundle exec rspec apps/web/auth/spec/config/hooks/omniauth_spec.rb apps/web/auth/spec/unit/omniauth_domain_validation_spec.rb` — 126 examples, 0 failures (both re-derive their own test-local email-parsing helpers, unaffected by the guard swap, confirmed still green)
- [x] `bundle exec rake try:unit` (full unit tryout suite) — 7369 passed, 0 failed
- [x] `bundle exec rubocop` on changed files — no new offenses

### DIFF
--- a/apps/web/auth/config/hooks/omniauth.rb
+++ b/apps/web/auth/config/hooks/omniauth.rb
@@ -631,19 +631,20 @@ module Auth::Config::Hooks
       # Global: Set via ALLOWED_SIGNUP_DOMAIN environment variable (comma-separated)
       #
       auth.before_omniauth_create_account do
-        email       = omniauth_email.to_s.strip.downcase
-        email_parts = email.split('@')
+        email = omniauth_email.to_s.strip.downcase
 
         # Reject unusable emails from IdP (distinct from policy rejection): a
-        # missing/empty claim, or one without both a local part and a domain.
+        # missing/empty claim, or any shape the accounts.valid_email CHECK
+        # constraint would reject (internal spaces, comma/semicolon in either
+        # part, a dotless domain, etc. — see SignupValidation::VALID_EMAIL_PATTERN).
         # Redirect with a stable error code so Login.vue can show a localized
         # message — matches the email_auth/omniauth_on_failure convention.
         # Inline JSON via throw_error_status was clobbered by omniauth_on_failure,
         # collapsing the specific code into the generic sso_failed. Failing here
-        # (rather than letting a blank local part like "@example.com" fall
-        # through to account creation, which 500s on the PG valid_email CHECK)
-        # keeps the user on a localized error instead of a frozen screen (#3478).
-        if email_parts.length != 2 || email_parts.first.to_s.empty? || email_parts.last.to_s.empty?
+        # (rather than letting a claim the CHECK rejects fall through to account
+        # creation, which 500s as Sequel::CheckConstraintViolation) keeps the
+        # user on a localized error instead of a frozen screen (#3478, #3971).
+        unless Onetime::SignupValidation.structurally_valid_email?(email)
           Auth::Logging.log_auth_event(
             :omniauth_invalid_email,
             level: :warn,
@@ -662,7 +663,7 @@ module Auth::Config::Hooks
             :omniauth_domain_rejected,
             level: :warn,
             email: OT::Utils.obscure_email(email),
-            domain: email_parts.last,
+            domain: email.split('@').last,
             display_domain: display_domain,
             provider: omniauth_provider,
           )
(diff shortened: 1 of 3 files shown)


delano:
@mahdi13830510 The regex is a character-exact mirror of the DB CHECK with correct `\A`/`\z` anchors. Looks good 👍 

I'll merge once the checks are done.

PR #4001, opened 2026-08-05
fix(auth): set verified on SSO-provisioned customers, reconcile drifted ones

### LINKED ISSUE #3973: SSO accounts never set customer.verified, blocking all system roles
## Problem Statement

An account provisioned via SSO/OmniAuth ends up with `verified = false` on its Redis `Customer` record, even though its Rodauth `accounts` row has `status = Verified`. Because `has_system_role?` gates on `cust.verified?` *before* it checks the role field, such an account cannot exercise any system role — promoting it with `bin/ots customers role promote` sets `role = colonel` but every colonel-gated endpoint still returns 403 `Requires site admin`.

Observed on a fresh database: the Colonel Admin UI produced a 403 on every panel it loaded, because each panel issues its own request and each one fails the same check.

```
Rodauth accounts row:  status = Verified
Redis Customer:        role = colonel, verified = false
```

## Early Diagnosis

Two customer-creation paths set `verified` differently, and nothing reconciles them afterward.

1. `Auth::Operations::CreateCustomer` hardcodes `verified: false` with the comment "needs to be updated in after_verify_account" (`create_customer.rb:65`). `after_omniauth_create_account` calls this operation directly (`omniauth.rb:712`), so an SSO signup lands here.
2. `Auth::Operations::SyncSession#create_customer` does the right thing — `customer.verified!(rodauth_status_verified?)` (`sync_session.rb:244`), deriving the flag from `status_id == 2`.

The catch is that path 2 only runs when `SyncSession` has to create the customer itself. In the normal SSO flow the customer already exists (created by path 1), so `find_existing_customer` (`sync_session.rb:216`) returns it untouched and the stale `verified: false` is never corrected.

`after_verify_account` — the hook that would otherwise flip the flag (`account.rb:342`, `verified_by: 'email'`) — is Rodauth's email-verification callback and never fires for SSO, since the IdP already asserted the email and Rodauth creates the account already-verified. The codebase notes the consequence at `account.rb:262`: "the Redis `customer.verified` flag never becomes true without `verify_account`".

Suspect this was simply missed rather than deliberate: the paths that do set the flag record provenance (`verified_by = 'invite_token'` at `account.rb:218`, `verified_by = 'email'` at `account.rb:343`) and there is no `'sso'` equivalent.

This fails closed — it denies access rather than granting it — so it is not a privilege-escalation risk. But it silently blocks SSO-provisioned operators from admin surfaces, and the failure mode is confusing because `customers show` reports `role: colonel`.

Likely fix: set `verified` in `after_omniauth_create_account` with a distinct `verified_by` provenance value for the federated assertion. Keeping it distinguishable from email-proved ownership matters — `CreateDefaultWorkspace`'s pending-federation claim treats verification as proof of email ownership, and an IdP assertion is a different kind of proof. Worth checking whether the provider's `email_verified` claim should be consulted before trusting it, since not every IdP asserts one.

## Affected Areas

- [ ] `apps/web/auth/config/hooks/omniauth.rb` — `after_omniauth_create_account` never sets `verified`
- [ ] `apps/web/auth/operations/create_customer.rb` — hardcodes `verified: false` for every caller
- [ ] `apps/web/auth/operations/sync_session.rb` — reconciles on create only, not on find
- [ ] `apps/web/auth/config/hooks/account.rb` — `after_verify_account`, the only flag-flipping path for password signups; `verified_by` provenance convention
- [ ] `lib/onetime/application/authorization_policies.rb` — `has_system_role?` verified-before-role gate
- [ ] `apps/web/auth/operations/create_default_workspace.rb` — consumes `verified?` as proof of email ownership for federated claims

## Reproduction

1. Fresh database, sign in via SSO to provision an account.
2. `bin/ots customers role promote <sso-user>`
3. `bin/ots customers show <sso-user>` → `role: colonel`, `verified: false`
4. Load Colonel Admin → 403 `Requires site admin` on every panel.

Workaround: `bin/ots customers verify <sso-user>`, then re-login.

mahdi13830510:
## Summary
- `Auth::Operations::CreateCustomer` hardcoded `verified: false` for every new customer, including SSO/OmniAuth JIT signups. Rodauth creates an SSO account already `status=Verified` (the IdP asserted the email) and never fires `after_verify_account` — the hook that flips this flag for password signups. Since `has_system_role?` gates on `verified?` *before* it checks the role field, an SSO-provisioned account could never exercise any system role, even after `bin/ots customers role promote`.
- `CreateCustomer` now accepts `verified:`/`verified_by:`, defaulting to the prior behavior (`false`/`nil`) so the password-signup path is unchanged. The OmniAuth `after_omniauth_create_account` hook passes `verified: account[:status_id] == 2, verified_by: 'sso'` — mirroring the logic `Auth::Operations::SyncSession#create_customer` already got right.
- `SyncSession` reconciles `verified` on every login now, not just when it has to create the customer itself (previously the only branch that set it correctly) — so an account already stuck unverified from before this fix self-heals on next login instead of requiring the `bin/ots customers verify` workaround. Safe for password accounts: Rodauth only reaches `status=Verified` for those via `after_verify_account`, which already sets this flag in lockstep, so the reconcile is a no-op there.
- `'sso'` is a new distinct `verified_by` provenance value, alongside the existing `'email'`/`'invite_token'`/`'stripe_payment'` — a federated IdP assertion is a different kind of proof than email confirmation.

Closes #3973

## Test plan
- [x] New `apps/web/auth/spec/operations/sync_session_verification_reconcile_spec.rb` (same doubles-based pattern as `sync_session_colonel_signin_spec.rb`) covers: reconciles an unverified customer when Rodauth says verified; is a no-op when already verified (no redundant writes on every login); is a no-op when Rodauth itself is not verified (password path unaffected)
- [x] `bundle exec rake spec:apps:web_auth` — 1090 examples (1086 + 4 new), 0 failures
- [x] `bundle exec rake try:unit` (full unit tryout suite) — 0 failed
- [x] `bundle exec rubocop` on changed files — no new offenses beyond the same pre-existing classes (`RSpec/VerifiedDoubles`, `RSpec/MultipleExpectations`) already present in the sibling spec this one mirrors

### DIFF
--- a/apps/web/auth/config/hooks/omniauth.rb
+++ b/apps/web/auth/config/hooks/omniauth.rb
@@ -704,6 +704,16 @@ module Auth::Config::Hooks
         signup_domain_id = custom_domain&.identifier
 
         # Create Customer record (same as regular signup)
+        #
+        # verified: true when the Rodauth account is already status=Verified
+        # (status_id == 2) — the IdP asserted the email, so Rodauth creates
+        # the account pre-verified and never fires after_verify_account (the
+        # hook that flips this flag for password signups). Hardcoding false
+        # unconditionally here left every SSO-provisioned customer unable to
+        # exercise any system role, since has_system_role? gates on
+        # verified? before it checks the role field (#3973). 'sso' keeps the
+        # provenance distinguishable from 'email'/'invite_token' — a
+        # federated assertion is a different kind of proof.
         customer = Onetime::ErrorHandler.safe_execute(
           'create_customer_omniauth',
           account_id: account_id,
@@ -715,6 +725,8 @@ module Auth::Config::Hooks
             db: db,
             provisioning_origin: 'sso_jit',
             signup_domain_id: signup_domain_id,
+            verified: account[:status_id] == 2,
+            verified_by: 'sso',
           ).call
         end
 
--- a/apps/web/auth/operations/create_customer.rb
+++ b/apps/web/auth/operations/create_customer.rb
@@ -22,12 +22,26 @@ module Auth
       #   Set on newly-created Customer records. Ignored for existing customers to
       #   preserve original signup context (same "don't rewrite history" rule as
       #   provisioning_origin).
-      def initialize(account_id:, account:, db: nil, provisioning_origin: nil, signup_domain_id: nil)
+      # @param verified [Boolean] Verification state for a newly-created Customer.
+      #   Defaults to false (the password-signup path: verification happens later,
+      #   via after_verify_account). SSO/OmniAuth JIT provisioning passes true here
+      #   when the Rodauth account is already status=Verified (the IdP asserted the
+      #   email) — hardcoding false unconditionally left every SSO-provisioned
+      #   customer unable to exercise any system role (#3973), since
+      #   has_system_role? gates on verified? before it checks the role field.
+      #   Ignored when the customer already exists (we don't rewrite history).
+      # @param verified_by [String, nil] Provenance for `verified`, distinguishing
+      #   a federated IdP assertion ('sso') from email confirmation ('email') or
+      #   an invite token ('invite_token'). Ignored when the customer already exists.
+      def initialize(account_id:, account:, db: nil, provisioning_origin: nil, signup_domain_id: nil,
+                     verified: false, verified_by: nil)
         @account_id          = account_id
         @account             = account
         @db                  = db || Auth::Database.connection
         @provisioning_origin = provisioning_origin
         @signup_domain_id    = signup_domain_id
+        @verified            = verified
+        @verified_by         = verified_by
       end
 
       # Executes the customer creation/loading operation
@@ -62,7 +76,8 @@ module Auth
           customer = Onetime::Customer.create!(
             email: email,
             role: 'customer',
-            verified: false, # needs to be updated in after_verify_account
+            verified: @verified,
+            verified_by: @verified ? @verified_by : nil,
             provisioning_origin: @provisioning_origin,
             signup_domain_id: @signup_domain_id,
           )
--- a/apps/web/auth/operations/sync_session.rb
+++ b/apps/web/auth/operations/sync_session.rb
@@ -207,8 +207,41 @@ module Auth
           end
           retried_customer || raise(OT::Problem, "Customer index sync failed for #{@account[:email]}")
         end
+        reconcile_verification_state(customer)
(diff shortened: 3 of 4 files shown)


delano:
@mahdi13830510 This will need some revisions. Three things:

1. The `return unless rodauth_status_verified?` guard is vacuous at the real call site — in a deployment with `verify_account` disabled, it fires for every password customer on first login, which amounts to "verify anyone who logs in unverified."
2. `verified_by = 'sso'` is stamped unconditionally on reconcile, including for accounts verified via other paths that drifted (the exact window documented in `set_customer_verification.rb:36-40`). That falsifies provenance in a PR whose purpose is preserving it.
3. #3973 asked two questions this skips: whether to consult the IdP's `email_verified` claim (right now `status_id == 2` trusts every IdP unconditionally), and the `CreateDefaultWorkspace` pending-federation interaction.

The headline change (passing `verified:` through `CreateCustomer`) will need test coverage as well. Happy to review a revision that addresses these.

delano:
Thanks @mahdi13830510 — the diagnosis here was exactly right, and it lives on in #4177 with you credited as co-author. Re-verified on current `main` before acting: `create_customer.rb` still hard-codes `verified: false`, the omniauth hook passes no verification state, and `has_system_role?` short-circuits on `!verified?` before reading the role — the lockout you identified is real.

Closing this PR because the review objections to the reconcile half were confirmed and stood unaddressed:

1. The login-path guard (`status_id == 2`) verifies ordinary password accounts too when `verify_account` is disabled — #4177 therefore has **no login-path reconcile at all**; drifted records are healed by an operator-invoked `customers doctor --repair` check gated on both `provisioning_origin == 'sso_jit'` (real provenance) and the accounts row at VERIFIED.
2. Stamping `verified_by = 'sso'` on any drifted record falsified provenance for email/stripe-verified customers — the doctor repair preserves existing `verified_by` and only stamps `'sso'` where there is none.
3. The hand-rolled `verified=`/`save_fields` path — replaced with the canonical `SetCustomerVerification` op (`rodauth_already_synced: true`, Redis-only).

#4177 also swaps the literal `2` for `AccountStatuses::VERIFIED`, registers `'sso'` in `VALID_VERIFIED_BY`, adds an explicit-`email_verified: false` IdP veto on the forward path, and covers the headline change with integration specs driving the real Rodauth callback — the coverage gap flagged in review here.

The hard part of this fix was seeing the bug's shape, and that was your work.

---

PR #4002, opened 2026-08-05
fix(security): resolve admin allowlist IP before privacy masking

### LINKED ISSUE #3912: AdminNetworkIsolation: allowed_cidrs finer than IP-privacy mask granularity silently never match (or over-match)
# AdminNetworkIsolation: allowlist CIDRs finer than the IP-privacy mask granularity silently never match (or over-match)

## Summary

`site.admin.allowed_cidrs` entries narrower than the IP-privacy masking granularity — IPv4 finer than `/24`, IPv6 finer than `/48` — are accepted at boot, logged as active, and then can never match a real client. An operator who allowlists their exact admin host (`203.0.113.7/32`, or a Tailscale node's IPv6 `/128`) is silently locked out of `/colonel` and `/api/colonel` with the intentional indistinguishable-from-absent 404.

The inverse also holds: an entry whose host bits are zero (`203.0.113.0/32`) matches **every** client in `203.0.113.0/24`, because all of them mask to `.0`. That widens the network gate silently (the two app-layer auth layers still enforce beneath it, so this is defense-in-depth erosion, not a bypass).

## Mechanism

This is not a parsing bug — `IPAddr.new("203.0.113.7/32").include?(IPAddr.new("203.0.113.7"))` is `true` (verified Ruby 3.4.10). It is a middleware-order interaction:

1. `Otto::Security::Middleware::IPPrivacyMiddleware` is mounted early in the universal stack (`middleware_stack.rb:296`). Privacy is enabled by default, and `ip_privacy_security_config` (`middleware_stack.rb:212`) sets `mask_private_ips = true` unconditionally — so **every** request gets a masked canonical IP.
2. Masking zeroes the last octet for IPv4 (`203.0.113.7` → `203.0.113.0`) and the last 80 bits for IPv6 (effective `/48`) — `otto/lib/otto/privacy/ip_privacy.rb` (`mask_ipv4`, `mask_ipv6`). The masked value is written to `env['otto.client_ip']` (`ip_privacy_middleware.rb:166`).
3. `AdminNetworkIsolation` is mounted **after** masking (`middleware_stack.rb:322`) and compares the allowlist against that masked IP (`admin_network_isolation.rb:109-116`, reading `env['otto.client_ip']` at line 121).

So the comparison operand is always on a `/24` (IPv4) or `/48` (IPv6) boundary. Entries coarser than that work — which is why the docstring's examples (`100.64.0.0/10`, RFC1918 ranges) behave and the bug stayed latent. Entries finer than that either never match or match the whole masked block.

The denial log prints the masked IP (`ip: 203.0.113.0`), which does not obviously connect to the configured `203.0.113.7/32` — the operator's most likely conclusion is "the middleware is broken," followed by disabling it.

Existing coverage doesn't catch this: `try/unit/middleware/admin_network_isolation_try.rb` stubs `env['otto.client_ip']` directly with unmasked IPs, so the masking interaction is never exercised.

## Related edge (same `allowed?` method)

An IPv4-mapped IPv6 client (`::ffff:203.0.113.7`, common on dual-stack sockets) never matches any IPv4 CIDR — `IPAddr#include?` returns `false` across families (verified). Normalizing with `IPAddr#native` in `allowed?` fixes this independently of the masking issue.

## Proposed fix

Check the allowlist against the **real** client IP, before masking — the same pattern IPPrivacyMiddleware itself uses for `otto.via_trusted_proxy` (`ip_privacy_middleware.rb:59`, "record the trust decision BEFORE any masking"):

1. Move `AdminNetworkIsolation` above `IPPrivacyMiddleware` in the universal stack. Construct it with the same `Otto::Security::Config` from `ip_privacy_security_config` and resolve the client IP itself via `Otto::Utils.resolve_client_ip(env, config)` — the code path it already has as a fallback (`admin_network_isolation.rb:124`). The real IP is used for the containment check only and never written to env, preserving the privacy contract. `/32` and `/128` entries then work as configured.
2. Normalize `addr = addr.native` in `allowed?` so IPv4-mapped IPv6 clients match IPv4 entries.
3. Keep fail-closed behavior for unresolvable IPs, unchanged.

Alternative considered and rejected: boot-time rejection of entries finer than mask granularity. It's honest but forces operators to widen their allowlist to `/24`/`/48` — worse posture than making exact-host entries actually work.

## Acceptance criteria

- [ ] With IP privacy masking active (the always-on default), a client whose real IP is `203.0.113.7` passes a `203.0.113.7/32` allowlist and reaches `/colonel` and `/api/colonel`.
- [ ] A client at `203.0.113.9` is denied (404) by that same `/32` — no over-match via the masked `.0`.
- [ ] IPv6: a `/128` entry admits exactly that host under masking.
- [ ] `::ffff:203.0.113.7` (IPv4-mapped) passes a `203.0.113.7/32` entry.
- [ ] `env['otto.client_ip']` observed by downstream middleware and apps remains the masked value — no real-IP leakage into env, logs, or session metadata.
- [ ] Unresolvable client IP with a configured allowlist still denies (fail closed).
- [ ] Tryouts exercise the real mount order (IPPrivacyMiddleware then/before AdminNetworkIsolation as fixed) rather than stubbing `otto.client_ip` with unmasked values.

## Testing

Extend `try/unit/middleware/admin_network_isolation_try.rb` plus an integration tryout that builds the two middlewares in stack order and drives Rack env through both, covering the acceptance matrix above.

## Affected files

- `lib/onetime/middleware/admin_network_isolation.rb` (`allowed?`, `resolve_client_ip`, construction)
- `lib/onetime/application/middleware_stack.rb` (mount order, config handoff)
- `try/unit/middleware/admin_network_isolation_try.rb`

## Severity / impact

- **Availability**: silent admin-surface lockout for exact-host allowlists; failure mode (404) is indistinguishable from success-at-hiding, by design.
- **Security**: `.0`-anchored fine entries silently widen the gate to the masked block (/24 IPv4, /48 IPv6). App-layer colonel auth still applies underneath.
- No data exposure. No action needed for deployments using `/24`-or-coarser IPv4 or `/48`-or-coarser IPv6 entries.

mahdi13830510:
## Summary
- `AdminNetworkIsolation` (the `site.admin.allowed_cidrs` gate for `/colonel` + `/api/colonel`) compared the allowlist against `env['otto.client_ip']` — but `IPPrivacyMiddleware`, mounted earlier in the universal stack, had already masked that value to a `/24` (IPv4) or `/48` (IPv6) boundary before this middleware ever saw it. Verified: `IPAddr.new("203.0.113.7/32").include?(IPAddr.new("203.0.113.0"))` is `false` — a `/32` (or `/128`) allowlist entry for an operator's exact admin host could never match the masked value, silently locking them out with the intentional indistinguishable-from-absent 404. Conversely, an entry whose host bits are zero (`203.0.113.0/32`) matched **every** client in that masked `/24`, silently widening the network gate (defense-in-depth erosion, not a bypass — the two app-layer auth checks still enforce beneath it).
- Fix mirrors the issue's own proposed direction: mount `AdminNetworkIsolation` **before** `IPPrivacyMiddleware`, sharing the same `Otto::Security::Config` (`ip_privacy_security_config`), and have it resolve the client IP itself via `Otto::Utils.resolve_client_ip` for the containment check only — the real IP is never written to `env`, so nothing downstream (logging, sessions, the privacy fingerprint) ever observes it. The privacy contract is unaffected; this middleware just gets to look at the real value before it's overwritten.
- Also normalizes with `IPAddr#native` in `allowed?` so an IPv4-mapped IPv6 client (`::ffff:203.0.113.7`, common on dual-stack sockets) matches an IPv4 CIDR entry — previously `IPAddr#include?` never matches across families.

Closes #3912

## Test plan
- [x] Rewrote `try/unit/middleware/admin_network_isolation_try.rb` to stop stubbing `env['otto.client_ip']` directly (the issue calls out this exact gap: "the masking interaction is never exercised") — now drives `REMOTE_ADDR`/`X-Forwarded-For` through the real `Otto::Utils.resolve_client_ip` with real `Otto::Security::Config` objects (direct-connect, trusted-proxy chain, and masking-enabled configs)
- [x] New coverage matches the acceptance criteria: a `/32` entry admits the exact real IP under masking; a neighbor on the same masked `/24` is still denied (no over-match); an IPv4-mapped IPv6 client passes an IPv4 `/32` entry; a spoofed `X-Forwarded-For` from an untrusted direct peer cannot bypass; a real trusted-proxy chain resolves and gates on the forwarded client correctly in both directions
- [x] `bundle exec try try/unit/middleware/admin_network_isolation_try.rb` — 34 passed
- [x] `bundle exec rake try:unit` (full unit tryout suite) — 7366 passed, 0 failed
- [x] `bundle exec rubocop` on the two production files — no offenses (the tryout file's offenses are the same pre-existing classes the original file already had, from the tryouts `#=>` assertion idiom rubocop doesn't understand)
- [x] Full app boot (`Onetime.boot! :app`) succeeds with the reordered middleware stack

### DIFF
--- a/lib/onetime/application/middleware_stack.rb
+++ b/lib/onetime/application/middleware_stack.rb
@@ -278,6 +278,28 @@ module Onetime
           # off, so native Rack X-Forwarded-Proto handling is unaffected.
           builder.use Onetime::Middleware::AssumeHttps
 
+          # Built once, shared by both AdminNetworkIsolation (below) and
+          # IPPrivacyMiddleware, so their trusted-proxy resolution can never
+          # drift apart. See ip_privacy_security_config.
+          ip_privacy_config = ip_privacy_security_config
+
+          # Admin network isolation - optional CIDR allowlist for the Colonel
+          # surfaces (/colonel + /api/colonel). No-op unless
+          # site.admin.allowed_cidrs is configured; then a request from
+          # outside the allowlist gets a 404 (indistinguishable-from-absent).
+          #
+          # Mounted BEFORE IPPrivacyMiddleware and given the same security
+          # config so it resolves the REAL client IP itself, rather than the
+          # masked value IPPrivacyMiddleware would otherwise have already
+          # written to REMOTE_ADDR / X-Forwarded-For by the time this ran. An
+          # allowlist entry finer than the masking granularity (a /32 or /128
+          # host entry) would otherwise silently never match against the
+          # masked value, or silently over-match the masked block (#3912).
+          # The real IP is used for the containment check only and never
+          # written to env, so the privacy contract is unaffected.
+          logger.debug 'Setting up Admin Network Isolation middleware'
+          builder.use Onetime::Middleware::AdminNetworkIsolation, ip_privacy_config
+
           # IP Privacy - masks public IPs before logging/monitoring
           # Private/localhost IPs are automatically exempted for development
           # Uses Otto's privacy middleware as a standalone Rack component.
@@ -288,7 +310,6 @@ module Onetime
           # visitor IP from every downstream consumer (ban checks, sessions,
           # identity resolution, the Colonel "current IP" panel). See
           # ip_privacy_security_config.
-          ip_privacy_config = ip_privacy_security_config
           logger.debug 'Setting up IP Privacy middleware',
             {
               note: 'masks public and private IPs',
@@ -313,14 +334,6 @@ module Onetime
           # Health endpoint access control - restrict to localhost/private networks
           logger.debug 'Setting up Health Access Control middleware'
           builder.use Onetime::Middleware::HealthAccessControl
-
-          # Admin network isolation - optional CIDR allowlist for the Colonel
-          # surfaces (/colonel + /api/colonel). No-op unless
-          # site.admin.allowed_cidrs is configured; then a request from outside
-          # the allowlist gets a 404 (indistinguishable-from-absent). Runs after
-          # IP privacy so it can use the trusted-proxy-resolved client IP.
-          logger.debug 'Setting up Admin Network Isolation middleware'
-          builder.use Onetime::Middleware::AdminNetworkIsolation
 
           builder.use Rack::ContentLength
           builder.use Onetime::Middleware::StartupReadiness
--- a/lib/onetime/middleware/admin_network_isolation.rb
+++ b/lib/onetime/middleware/admin_network_isolation.rb
@@ -30,14 +30,24 @@ module Onetime
     #
     # The allowlist check MUST use the trusted-proxy-resolved client IP, never a
     # raw forwarding header, or the allowlist is trivially spoofable by sending
-    # `X-Forwarded-For: <an-allowed-ip>`. We resolve from the canonical
-    # env['otto.client_ip'] set once by the universal IPPrivacyMiddleware mount
-    # (configured from site.network.trusted_proxy via
-    # MiddlewareStack.ip_privacy_security_config), with the same
-    # Otto::Utils.resolve_client_ip fallback the auth strategies use. This is the
-    # identical resolution the rest of the stack relies on, so the network gate
-    # agrees with ban checks, sessions, and audit attribution. When the IP cannot
-    # be resolved and an allowlist is configured, the request is denied (404) —
+    # `X-Forwarded-For: <an-allowed-ip>`. It must ALSO use the real client IP,
+    # never the IP-privacy-MASKED one — an allowlist entry finer than the
+    # privacy masking granularity (IPv4 finer than /24, IPv6 finer than /48)
+    # would otherwise silently never match (a /32 or /128 host entry) or
+    # silently over-match (any entry whose host bits are zero matches the
+    # entire masked block) (#3912).
+    #
+    # So this middleware is mounted BEFORE the universal IPPrivacyMiddleware
+    # (MiddlewareStack.configure) and resolves the client IP itself, via the
+    # same Otto::Utils.resolve_client_ip the auth strategies and
+    # IPPrivacyMiddleware itself use, fed the identical
+    # Otto::Security::Config (MiddlewareStack.ip_privacy_security_config) —
+    # so this gate's trusted-proxy trust decision cannot drift from the rest
+    # of the stack's. The resolved real IP is used for the containment check
+    # ONLY; it is never written to env, so no downstream consumer (logging,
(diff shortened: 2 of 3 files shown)


delano:
The mount-order fix and keeping the real IP out of `env` are right. 

One regression: the denial log at `admin_network_isolation.rb:94` now emits the unmasked IP where it previously logged the masked value. The spec is explicit that blocked-request logging uses masked `otto.client_ip` (`domain-ip-allowlist.md:245`). Please fix that, add a test for the mount-order change (reverting `middleware_stack.rb` alone currently breaks nothing), and address the issue's IPv6 `/128` items.

The ticket is a pretty gnarly change but happy to go back and forth until we get there. Or it's also okay if you want to skip it for now. Whichever you prefer.

delano:
Thanks for this, and apologies it sat while the surrounding code moved.

Closing as superseded: the defect you identified was real, but it was fixed on `main` in 144256a (2026-08-10) via a different mechanism. That commit judges allowlist membership through `env['otto.ip_match']` — the verdict-only closure `IPPrivacyMiddleware` installs over the resolved pre-mask address — rather than relocating the middleware. The effect is the same `/32`-`/128` precision you were after, but the unmasked address never lands in `env`, a log, or the middleware itself, and the denial `WARN` still reports only the masked value.

Two reasons a rebase isn't the right path here rather than just merge friction:

1. **The mount-order move is no longer safe.** `main` mounts `AdminNetworkIsolation` at `middleware_stack.rb:567`, deliberately below both `Rack::DetectHost` and `IPPrivacyMiddleware`. Since this PR opened, #4062 added a host-allowlist gate that depends on that ordering — `host_provenance_trusted?` and `detected_host` read state `DetectHost` sets. Mounting above it, as this PR does, leaves that state unset and would misjudge the host gate on every proxied deployment.
2. **Rebasing would discard both production hunks**, leaving only tests that duplicate coverage `main` already has through `ip_match_env`. That's a rewrite, not a rebase.

What you raised that is **still genuinely unfixed**, and worth a small separate PR if you'd like to take it:

- `allowed?` still has a bare `IPAddr.new(client_ip)` on the fallback path with no `.native` normalization, so IPv4-mapped IPv6 (`::ffff:`) addresses aren't normalized there.
- No tryout example covers `::ffff:` or IPv6 `/128`.

#3912 stays open for exactly those items. Your `IPAddr#native` change and the mapped-address tryout case would land cleanly as a focused follow-up against current `main`.

---

PR #4010, opened 2026-08-05
fix(workspace): keep recipient field in generate-password mode

### LINKED ISSUE #4009: Generated password is not emailed unless the recipient was filled in first
Picking Generate Password removes the recipient field, so the password comes back as a link only and no mail goes out. Fill in a recipient *before* switching and the same flow does send it, so the result depends on the order you touch the controls.

Steps:
1. Open the homepage, leave the recipient field empty.
2. Open the dropdown next to Create Link and pick Generate Password.
3. The message field and the recipient field are both replaced by the password generator panel.
4. Submit. The password is created but nobody is notified, the link is the only way to share it.

Expected: if a recipient is set, the generated password is mailed, same as the create-link flow.

Suggestion: let the generator panel replace only the message field and leave the recipient field where it is. One delivery path, and the click order stops mattering.

Version: v0.26.3, self-hosted (Docker image `onetimesecret/onetimesecret:v0.26.3`).

syphernl:
Fixes #4009.

Picking Generate Password from the split button unmounts the recipient input, so the generated password never gets mailed and the link is the only way to share it. Fill in a recipient before switching and the same submit does send the mail, so the result depends on the order you click things.

`createPayload` in `useSecretConcealer.ts` already sends `recipient` for both `conceal` and `generate`, and the generate path reaches `send_email_to_recipient` on the server, so dropping the `selectedAction === 'create-link'` term from `showRecipient` is the whole fix.

The second commit moves the field below the generator panel. The textarea is hidden in that mode, so the input was otherwise the first thing in the card, sitting above the box that explains a password will be generated.

The new case in `WorkspaceSecretFormCapabilities.spec.ts` fails on the old code and passes on the new one. Reported on v0.26.3, same code path on main.

### DIFF
--- a/src/apps/workspace/components/forms/WorkspaceSecretForm.vue
+++ b/src/apps/workspace/components/forms/WorkspaceSecretForm.vue
@@ -200,12 +200,12 @@
   // Track selected action from SplitButton
   const selectedAction = ref<'create-link' | 'generate-password'>('create-link');
 
-  // Recipient field shows in create-link mode unless ui.capabilities.recipient
-  // is explicitly disabled. An unset flag is treated as enabled (config default).
+  // Recipient field shows in both modes unless ui.capabilities.recipient is
+  // explicitly disabled. An unset flag is treated as enabled (config default).
+  // Generated passwords are mailed too, so hiding the field in generate mode
+  // silently dropped the only way to reach the recipient.
   const { uiCapabilities } = storeToRefs(bootstrapStore);
-  const showRecipient = computed(
-    () => selectedAction.value === 'create-link' && uiCapabilities.value?.recipient !== false
-  );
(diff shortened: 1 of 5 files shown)


delano:
Nice job. The logic checks out, matches your #4009 ticket exactly, and the new test reproduces the regression correctly.

One pre-existing codesmell: `submit()` in `useSecretConcealer.ts` skipped *all* client-side validation for `type === 'generate'` (since the full schema requires `secret`, which generate mode doesn't fill). Now that it's visible, a malformed email typed there skipped validation entirely and went straight to the server's `invalid_email` check. 

I pushed a follow-up commit (`a0c5b155b2`) that adds a recipient-only validator (`validation.validateRecipient()` in `useSecretForm.ts`) and runs it for generate submissions.

PR #4021, opened 2026-08-06
*fix(middleware): respect otto.via_trusted_proxy in DetectHost

delano:
> [!IMPORTANT]
> **Deploy checklist** — no new config vars, but the fix only takes effect when Otto's existing proxy-trust settings cover the ingress:
> - [ ] `TRUSTED_PROXY_ENABLED=true` and `TRUSTED_PROXY_MODE=filter` (depth mode gets no forwarded-host trust — known gap, otto #4024)
> - [ ] Current Approximated egress ranges present in `TRUSTED_PROXY_CIDRS` — unlisted egress IPs get `via_trusted_proxy=false` and custom domains still fall to canonical
> - [ ] Edge proxy sets or strips `X-Forwarded-Host`/`Forwarded` rather than passing them through from clients (see `TRUSTED_PROXY_CIDRS` note in `.env.reference`)
> - [ ] Post-deploy: watch for the `Apx-Incoming-Host … 2026-08-05 incident signature` WARN — silence plus resolving custom domains confirms the trust path

When `TRUSTED_PROXY_ENABLED` is on (filter mode), Otto's IPPrivacyMiddleware rewrites `REMOTE_ADDR` to the resolved public visitor IP before DetectHost runs, so DetectHost's `private_ip?(REMOTE_ADDR)` trust check failed and forwarded host headers (`Apx-Incoming-Host`, `X-Forwarded-Host`, …) were discarded — every custom domain fell back to canonical (2026-08-05 incident).

**Fix — grant-only trust semantics:** `otto.via_trusted_proxy == true` (recorded by otto from the original peer, pre-rewrite) grants forwarded-host trust; a false/absent/non-boolean key falls through to the legacy private-IP heuristic rather than suppressing it. IPPrivacyMiddleware is mounted unconditionally, so with proxy trust unconfigured the key is false on every request — suppressing the heuristic would strip forwarded-host trust from default-config self-hosted installs behind a local reverse proxy. The security invariant is unchanged: direct public clients get Host-only, and the OR is a strict superset of pre-incident behavior adding only the otto-verified case.

Also: key literal centralized in `VIA_TRUSTED_PROXY_KEY` (contract-pinned against `Otto::EnvKeys::VIA_TRUSTED_PROXY`), the untrusted-path log now names the actual trust inputs, and discarding a header set that includes `Apx-Incoming-Host` logs at WARN — the incident's exact signature, per the incident-report alerting suggestion.

**Tests:** unit tryouts pin grant-only semantics, non-boolean graceful fallthrough, and the WARN escalation; a new cross-gem stack tryout mounts the real IPPrivacyMiddleware in production order via the production `ip_privacy_security_config` builder (catches config-construction drift), reproduces the incident shape, and pins the depth-mode gap as a KNOWN LIMITATION (depth mode has an empty trust list, so the key is always false — forwarded-host trust unavailable; pre-existing, needs an upstream otto change).

**Lane runner:** env scrub consolidated into one owned block (family globs + explicit singletons), extended with `ENABLE_ORGS`, `TRUSTED_PROXY_*`, `APPROXIMATED_*`, mail/identity vars and friends; generator script's embedded runner synced byte-identical.

Verified: `try/integration/middleware/` 98/98; `tests/lanes/run unit` green (try:unit 7367/0, spec:fast 3307/0 incl. `ip_privacy_parity_spec`); `tests/lanes/run disabled` 838/0; rubocop clean.

### DIFF
--- a/.env.reference
+++ b/.env.reference
@@ -1160,6 +1160,15 @@ TRUSTED_PROXY_HEADER=X-Forwarded-For
 # filter mode only: additional CIDR ranges to trust as proxies beyond the
 # RFC1918 defaults (e.g. a CDN/proxy with public IPs). Comma-separated.
 # Example: 203.0.113.0/24,2001:db8::/32
+#
+# SECURITY: listing a range here is a full infrastructure-trust grant, not
+# just client-IP trust. Requests arriving from these ranges also have their
+# forwarded HOST headers honored (X-Forwarded-Host, X-Original-Host,
+# Apx-Incoming-Host, Forwarded) for custom-domain detection. The proxy MUST
+# therefore set or strip those headers itself rather than pass them through
+# from clients — a pass-through proxy lets a client pick the tenant domain
+# the app renders. This is the same requirement stated above for
+# X-Forwarded-For, extended to the host headers.
 #TRUSTED_PROXY_CIDRS=
 
 # depth mode only: number of proxy hops to skip from the right. Default: 1.
--- a/apps/api/organizations/spec/logic/organizations/list_secret_activity_spec.rb
+++ b/apps/api/organizations/spec/logic/organizations/list_secret_activity_spec.rb
@@ -186,7 +186,9 @@ RSpec.describe OrganizationAPI::Logic::Organizations::ListSecretActivity do
 
     # The denial is an operator-visible security event: extid/actor must land
     # in OT.info's **payload (structured, filterable) rather than being
-    # concatenated into the message string.
+    # concatenated into the message string. Values are pinned, not `anything` —
+    # `actor: anything` matches nil, so a regression that dropped the actor
+    # would still pass while leaving the alert unattributable.
     it 'logs the denial with a structured payload' do
       allow(OT).to receive(:conf)
         .and_return(conf_with_audit_logs_flag('audit_logs_enabled' => false))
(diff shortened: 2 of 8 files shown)


delano:
**Re: forwarded-host trust now coupled to the trusted-proxy CIDR list**

This coupling is intentional — it is the fix, not a side effect:

1. The old behavior (categorical refusal of public peers) *was* the 2026-08-05 incident. Approximated's ingress egresses from public IPs; refusing them forwarded-host trust sent every custom domain to canonical. Any fix must grant host trust to some public peers.
2. The trusted-proxy list already means "this peer is our infrastructure." A hop trusted to assert the client IP via `X-Forwarded-For` terminates the connection and can equally assert any host header — splitting the two grants doesn't reduce what a trusted hop can do.
3. The residual risk is narrower than stated: a trusted proxy that *passes through* a client-supplied `X-Forwarded-Host` instead of setting/stripping it. That hole is identical under the pre-existing private-IP heuristic (an nginx on 10.x that doesn't strip the header), so no new vulnerability class is introduced — the existing operator responsibility now extends to the explicitly configured CIDR list.
4. Impact is bounded downstream: `DomainStrategy` sanitizes `display_domain` via `basically_valid?` with canonical fallback, and a spoofed hostname only selects a strategy — `:custom` still requires a registered `CustomDomain` to change behavior.

Addressed in 093578d7f1 by documenting the grant explicitly in `.env.reference` under `TRUSTED_PROXY_CIDRS`: listing a range is a full infrastructure-trust grant, and the proxy must set or strip the forwarded host headers rather than pass them through.

If a real deployment surfaces that needs client-IP trust from a hop *without* host trust (a CDN that passes headers through unsanitized), the follow-up is a separate `HOST_TRUSTED_PROXIES` override defaulting to the existing list — not building it speculatively.

PR #4027, opened 2026-08-07
fix: stop leaking server-side domains config through the bootstrap serializer

delano:
The bootstrap window payload serialized the raw `:domains` config section, which includes server-side fields (notably the cluster `api_key`) that must never reach the browser. This enforces an output boundary in two layers:

- **Backend**: `ConfigSerializer` now emits an explicit allowlist for the domains section, and the serializer registry strips any output key not declared by a registered serializer, so future config additions default to private.
- **Frontend**: the Zod domains/features config schemas drop the server-side fields so the contract matches what the backend now sends.

Note: the exposed cluster `api_key` should be rotated independently of this fix.


delano:
Redundant — this fix already landed on main as ab2ba9289 (shipped in v0.26.x, keys rotated). Branch was a stale duplicate.

Issue #4048, opened 2026-08-08
Billing currency display: locale-blind formatting, mixed free/paid symbols, null-currency crash


delano:
Three defects found while tracing `BILLING_CURRENCY` -> UI symbol rendering. Line numbers verified against main at 4aa35f1ea.

## 1. Locale-blind price formatting

`src/types/billing.ts:123-128`:

```ts
export function formatCurrency(amount: number, currency: string = 'USD'): string {
  return new Intl.NumberFormat(undefined, {
    style: 'currency',
    currency,
  }).format(amount / 100); // Assuming amount is in cents
}
```

`undefined` means the **browser** locale, not the app's i18n locale. The same deployment renders `CA$29.00` (en-US browser), `$29.00` (en-CA), `29,00 CA$` (de-DE) — and the app language switcher has no effect on prices.

Affects every caller: `PlanCard.vue`, `Pricing.vue`, `InvoiceList.vue`, `PlanChangeModal.vue`.

## 2. Mixed currency symbols on one pricing grid

Config-only plans (free tier) are stamped with `OT.billing_config.currency` in `apps/web/billing/operations/catalog/config_loader.rb` (`upsert_config_only_plans`, ~lines 38-77; currency stamping at ~line 171 `family_currency = OT.billing_config.currency`), while Stripe-sourced paid plans carry `price.currency` from Stripe (`data_extractor.rb:46,139`).

After a `BILLING_CURRENCY` change these two sources diverge and a single pricing grid can mix `$` and `€`. `planSelectorLogic.ts` exempts the free tier from mismatch checks, but only for button enablement — not for display.

## 3. `null` currency crashes the formatter

`formatCurrency`'s `currency: string = 'USD'` default only catches `undefined`. But `plan_page_record` (`apps/web/billing/controllers/billing.rb:1062-1087`) computes:

```ruby
currency = (price_data ? price_data['currency'] : nil) || plan.currency
```

which can emit `null` (price-less plan whose `plan.currency` is unset). `Intl.NumberFormat` with `currency: null` throws `RangeError`, blanking the plan card. The TS contract claims `currency: string` (`billing.service.ts:124`), so nothing downstream guards against it.

## Origin

Found during #4025/#4033 currency-migration follow-up trace.

PR #4049, opened 2026-08-08
fix(billing): stop formatCurrency crashing on a null currency

eeshsaxena:
### The bug

`formatCurrency` in `src/types/billing.ts` declared `currency: string = 'USD'`:

```ts
export function formatCurrency(amount: number, currency: string = 'USD'): string {
  return new Intl.NumberFormat(undefined, {
    style: 'currency',
    currency,
  }).format(amount / 100);
}
```

A default parameter value only applies when the argument is `undefined`. A `null` currency passes straight through, and `Intl.NumberFormat` throws a `RangeError` on a null currency code, which takes the price display down.

`null` really can reach here. `plan_page_record` in `apps/web/billing/controllers/billing.rb` computes:

```ruby
currency = (price_data ? price_data['currency'] : nil) || plan.currency
```

so when the Stripe price carries no currency it falls back to `plan.currency`, which can itself be nil. After serialization the frontend calls `formatCurrency(amount, null)` and crashes.

This is defect 3 in #4048.

### The fix

Normalise a null or empty currency to the `USD` default before handing it to `Intl.NumberFormat`:

```ts
const resolvedCurrency = currency || 'USD';
```

I also widened the parameter type to `string | null` so it matches what actually arrives at runtime. Calling with no argument still defaults to USD, so existing callers are unaffected.

### Tests

Added two cases to `src/tests/types/billing-helpers.spec.ts`: a null currency and an empty-string currency both fall back to USD instead of throwing. Verified against the same Intl engine that `currency: null` raises `RangeError` today and that `currency || 'USD'` formats cleanly.

The other two defects in #4048 (locale-blind formatting and mixed free/paid symbols across a pricing grid) are separate concerns and left for their own changes.

### DIFF
--- a/src/tests/types/billing-helpers.spec.ts
+++ b/src/tests/types/billing-helpers.spec.ts
@@ -49,6 +49,18 @@ describe('formatCurrency', () => {
     expect(result).toContain('50');
   });
 
+  it('falls back to USD when currency is null instead of throwing', () => {
+    // A plan can reach the UI with a null currency (no Stripe price currency
+    // and no plan currency), which used to crash Intl.NumberFormat.
+    expect(() => formatCurrency(5000, null)).not.toThrow();
+    expect(formatCurrency(5000, null)).toContain('50');
+  });
+
+  it('falls back to USD when currency is an empty string', () => {
+    expect(() => formatCurrency(5000, '')).not.toThrow();
+    expect(formatCurrency(5000, '')).toContain('50');
+  });
+
   it('formats EUR currency', () => {
     const result = formatCurrency(1999, 'EUR');
     expect(result).toContain('19.99');
--- a/src/types/billing.ts
+++ b/src/types/billing.ts
@@ -120,9 +120,15 @@ export function getInvoiceStatusLabel(
   return t(`web.billing.invoices.${status}`);
 }
 
-export function formatCurrency(amount: number, currency: string = 'USD'): string {
+export function formatCurrency(amount: number, currency?: string | null): string {
+  // A default parameter value only fills in `undefined`, so a `null` currency
+  // slipped straight through to Intl.NumberFormat, which throws a RangeError on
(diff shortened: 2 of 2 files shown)


delano (commented):
Reviewed and verified locally against the PR head (a4941d3). This looks correct and ready to merge once CI is approved to run.

**Verified:**
- The root-cause claim checks out: `plan_page_record` in `apps/web/billing/controllers/billing.rb:1065` computes `currency = (price_data ? price_data['currency'] : nil) || plan.currency`, and `plan.currency` can be nil (elsewhere the same controller guards it with `plan&.currency || 'unknown'`). So `null` genuinely reaches `formatCurrency`, and a default parameter only fills in `undefined` — `Intl.NumberFormat` then throws `RangeError`.
- `currency || 'USD'` handles `null`, `undefined`, and `''`; calling with no argument still yields USD, so existing callers are unaffected. Widening the signature to `string | null` matches what actually arrives after JSON serialization.
- Full spec file passes locally: 26/26 in `src/tests/types/billing-helpers.spec.ts`, including the two new cases. `pnpm run type-check` is clean with the widened signature.

**Two minor notes, neither blocking:**
1. The hardcoded `'USD'` fallback could show the wrong symbol on a deployment configured for another currency (`BILLING_CURRENCY=EUR`). As a crash guard it's the right trade-off — matching the existing default — but if defects 1–2 in #4048 lead to plumbing the configured currency into the frontend, this fallback should pick that up too.
2. This guards the display layer; the backend still emits `currency: null` in that edge case. Worth a small server-side fallback in `plan_page_record` eventually so the API contract (`currency: string`) holds — but that's separate from this fix and fine as a follow-up.

---

delano:
Looks good. Thank you. 

The visual tests passed. The failure is just workflow infrastructure, not related to the change.

PR #4053, opened 2026-08-08
Fix RFC 7239 Forwarded host detection

### LINKED ISSUE #4040: DetectHost lists RFC 7239 `Forwarded` but never parses its host= parameter
## Problem

`Rack::DetectHost` includes `'Forwarded'` in `FORWARDED_HEADERS` (`lib/middleware/detect_host.rb`), but `normalize_host` treats the raw header value as a plain hostname: it takes `split(',').first` and hands it to `DomainParser.extract_hostname`. An RFC 7239 value looks like `for=203.0.113.7;host=example.com;proto=https` — the `host=` parameter is never extracted, so the header can never contribute a valid host. It silently does nothing while the constant and the class doc comment imply support.

Flagged independently by two reviewers on #4028 and #4029 (both non-blocking, pre-existing — the tri-state work didn't change this behavior, so it was deliberately left out of that stack).

## Options

1. **Parse it**: extract the `host=` parameter (first element of a comma-separated list, quoted-string aware per RFC 7239 §4) before handing off to `DomainParser`. Honors the documented precedence.
2. **Remove it**: drop `'Forwarded'` from `FORWARDED_HEADERS` and the class doc. Smaller, honest about actual behavior; RFC 7239 emission is still uncommon (HAProxy/Apache opt-in).

Either resolves the misleading listing; a fix should also update the header-precedence documentation in the class comment.

## References

- #4028 review thread: https://github.com/onetimesecret/onetimesecret/pull/4028#discussion_r3739368865 (follow-up note 1 in the claude-review summary)
- #4029 review: https://github.com/onetimesecret/onetimesecret/pull/4029#issuecomment-5223196136 (follow-up suggestion 2)
- Related: #4024 (tri-state trust arc, where this surfaced)

XX-Q:
## Summary

[#4040](https://github.com/onetimesecret/onetimesecret/issues/4040)

- extract the `host=` parameter from the first RFC 7239 `Forwarded` element
- support quoted host values, escaped quoted strings, ports, and quoted delimiters
- fail closed on malformed values and continue to the next host header
- add regression coverage and a changelog fragment

## Related Issues

Fixes #4040

## Test Plan

- `tests/lanes/run unit` — 7,415 Tryouts passed; all RSpec unit groups passed with zero failures
- `bundle exec rubocop lib/middleware/detect_host.rb try/integration/middleware/detect_host_try.rb`
- exercised unquoted, quoted-with-port, multi-element, missing-host, and malformed-quote cases

### DIFF
--- a/changelog.d/20260808_170500_codex_4040_forwarded_host.rst
+++ b/changelog.d/20260808_170500_codex_4040_forwarded_host.rst
@@ -1,0 +1,14 @@
+.. A new scriv changelog fragment.
+
+Fixed
+-----
+
+- ``Rack::DetectHost`` now extracts the RFC 7239 ``host=`` parameter from
+  trusted ``Forwarded`` headers, including quoted host values and ports,
+  instead of treating the complete parameter list as a hostname. (#4040)
+
+AI Assistance
+-------------
+
+- Codex assisted with the implementation and regression tests for the
+  ``Forwarded`` host parsing fix. (#4040)
--- a/lib/middleware/detect_host.rb
+++ b/lib/middleware/detect_host.rb
@@ -228,7 +228,7 @@ module Rack
       # Try headers in order of precedence
       headers_to_check.each do |header|
         header_key = "HTTP_#{header.tr('-', '_').upcase}"
-        host       = self.class.normalize_host(env[header_key])
+        host       = self.class.normalize_host(env[header_key], forwarded: header == 'Forwarded')
         next if host.nil?
 
         if self.class.valid_domain_name?(host)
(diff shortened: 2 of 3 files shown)


delano (commented):
Reviewed the parser implementation, ran the full DetectHost tryout suite locally on this branch (24/24 pass, including the 5 new cases), and probed the new `forwarded_host` path with adversarial inputs. Overall this is a solid, correctly-scoped implementation of option 1 from #4040. One item from the issue is missing; the rest are observations, none blocking.

**Verified behavior (beyond the PR's own tests)**

- Case-insensitive `HOST=` matching, whitespace around `;`/`=`, quoted values containing `;`/`,`/`=`, escaped quotes, and multiple `host=` params (first wins) all behave per RFC 7239 §4.
- Empty `host=` / `host=""`, trailing backslash inside a quote, and unterminated quotes all fail closed to the next header, as documented.
- A non-RFC bare hostname in `Forwarded` (e.g. `Forwarded: legacy.example.com`) now falls through to `Host` instead of being accepted. That's a behavior change from the old code path, but the old acceptance was accidental and the value isn't valid RFC 7239 syntax — the new behavior is correct. Worth being aware of in case anything in the wild sends that shape.
- Decoded quoted values can carry characters like `"` or `;` into `DomainParser` (e.g. `host="semi;colon.example.com"` → detected host `semi;colon.example.com`). I checked the same values through the untouched `X-Forwarded-Host` path and the leniency is identical, so this PR adds no new validation surface — that's a pre-existing `DomainParser` characteristic, out of scope here.

**Missing from the issue's ask**

#4040 says "a fix should also update the header-precedence documentation in the class comment." The class doc (lines 32–37) still lists the old precedence — it omits `Apx-Incoming-Host` and orders `Forwarded` before `Host` differently than `HEADER_PRECEDENCE` actually does — and still doesn't mention that the `host=` parameter is now parsed. A small doc update there would close out the issue fully.

**Follow-up candidate (pre-existing, not this PR)**

RFC 7239 requires IPv6 hosts to be bracketed and quoted, and `host="[2001:db8::1]:8080"` currently yields a detected host of `[2001` (DomainParser splits on the first `:`; the result then passes `valid_domain_name?` since `[2001` isn't parseable as an IP). The same happens via `X-Forwarded-Host: [2001:db8::1]:8080` today, so it's not introduced here — but the Forwarded syntax makes bracketed IPv6 more likely to show up legitimately. Might deserve its own issue.

**Nits**

- `split_quoted_header` fails the *entire* header when a later element has unbalanced quotes (e.g. `host=good.example.com, for="unterminated`), even though the first element is well-formed. Conservative and fine for a security-sensitive path — just noting it's stricter than element-local failure, in case a comment there would help future readers.
- The changelog fragment's "AI Assistance" section matches existing repo convention 👍

---

delano:
Nice work — the quote-aware scanner held up against every edge case we threw at it, and the fail-closed instinct was exactly right. #4054 carries this commit forward (authorship preserved) and swaps the scanner for `Rack::Utils.forwarded_values`: Rack already ships a hardened RFC 7239 parser with DoS bounds, and delegating beats maintaining our own. Thanks for the contribution!

---

PR #4122, opened 2026-08-11
chore(ci): give the hygiene job a Node env for Prettier

delano:
Reduced from #4119 

Prepares the `T1 · Hygiene (pre-commit)` job to run a `prettier-staged` hook: adds
the `setup-node-env` action (the hook is `language: system`, so it needs `pnpm` on
PATH and `prettier-plugin-vue` / `prettier-plugin-tailwindcss` resolvable from
`node_modules` — `dlx` won't do), and raises the timeout 3 → 6 min to cover the
install. The comment explains why Prettier belongs here rather than in T1 ·
TypeScript Lint: it also owns md/yaml/json/css, which the `typescript` path filter
doesn't match, so a docs-only PR would skip formatting entirely.

Also marks three shebang-carrying scripts 100755 (mode-only; siblings already were).

Note: the `prettier-staged` hook itself is not in `.pre-commit-config.yaml` yet —
this lands the CI infrastructure ahead of it, so the job is a no-op change today.


delano:
Not worth merging since `prettier-staged` isn't included and it's not included b/c it's way too greedy for files to update. Same reason the PR #4119 this came from was closed.

Issue #4024, opened 2026-08-07
Depth-mode trusted proxy cannot grant forwarded-host trust (otto.via_trusted_proxy always false)


delano:
Follow-up from PR #4021 / the 2026-08-05 trusted-proxy custom-domain incident.

**Problem:** In `TRUSTED_PROXY_MODE=depth`, `middleware_stack.rb` sets only `trusted_proxy_depth` (otto raises if CIDRs are mixed in), leaving otto's trusted-proxy matcher list empty. `Otto::Security::Config#trusted_proxy?` short-circuits to false on an empty list, so `IPPrivacyMiddleware` records `otto.via_trusted_proxy = false` on every request — and it also rewrites `REMOTE_ADDR` to the resolved public client IP. Result: in depth mode, `Rack::DetectHost` has no trust signal at all (key false, REMOTE_ADDR public), so forwarded host headers (`Apx-Incoming-Host`, `X-Forwarded-Host`) are discarded and all custom domains classify as canonical.

This is pre-existing (not a regression from #4021 — filter mode was the incident; depth mode was equally broken before) and is pinned as a KNOWN LIMITATION case in `try/integration/middleware/detect_host_ip_privacy_stack_try.rb`, which will deliberately fail when the fix lands.

**Fix direction (upstream in otto):** depth mode should record a peer-trust verdict in `otto.via_trusted_proxy` — e.g. treat the immediate connecting peer as trusted whenever `trusted_proxy_depth > 0` is configured, or expose the pre-rewrite peer address in a dedicated env key so downstream middleware can make its own decision. Note `otto.original_ip` does not help: it is only set when privacy is disabled and holds the resolved client IP, not the connecting peer.

**Related:** the depth-value off-by-one (`ots_depth + 1` remap vs documented values) makes depth mode spoofable in the documented single-proxy setup — operators should stay on filter mode until both issues are reconciled. That caveat is documented in the DetectHost comment block.

delano:
Upstream otto issue filed: delano/otto#226 (depth mode never records peer trust in `otto.via_trusted_proxy`). The KNOWN LIMITATION tryout case in `detect_host_ip_privacy_stack_try.rb` will fail when that fix lands, prompting cleanup here.

Issue #3987, opened 2026-08-04
Encryption rotation, write side: per-deployment personalization (#3630 phase 2)


delano:
Second half of the encryption rotation started in v0.26.4 (#3630, `feature/3630-familia-212-reader-prep`).

**Prerequisite:** every process reading the datastore is on ≥ v0.26.4 (familia ≥ 2.12).

Rotate `encryption_personalization` off the library default (`'FamilialMatters'`) to a per-deployment value derived from `site.secret` (KeyDerivation, dedicated purpose). Must be two-phase within a datastore — v0.26.4 is necessary but not sufficient, since v0.26.4 readers have no history entry for the new value:

1. **Deploy A:** derive the target value and add it to `encryption_personalization_history`; `current` stays `'FamilialMatters'`. Readers now accept both; writers unchanged.
2. **Deploy B:** flip `encryption_personalization` to the derived value. Legacy envelopes keep decrypting automatically — familia 2.12 always appends `'FamilialMatters'` as the last decrypt candidate, so no history entry is needed for it.
3. Pin `algorithm: 'xchacha20poly1305'` explicitly on encrypted fields (already the de-facto default with rbnacl bundled; pin for determinism) and update the crypto-config regression spec in `spec/integration/all/initializers/boot_part2_spec.rb`, which intentionally fails on this change.
4. Fixture-envelope regression: decrypt a pre-rotation ciphertext under the rotated config.
5. Upgrade notes: self-hosters get the same A-then-B order; both steps are config-only after v0.26.4.

---

## Decision framework: verification-on-read (#3630)

**Current state:** `Familia::VerifiableIdentifier.verified_identifier?` is never called — identifiers are opaque lookup strings. Signing-key changes are safe today.

**Constraint:** Re-minting identifiers is not an option — existing links must keep resolving indefinitely.

**Required path:** Key-history fallback mirroring `encryption_personalization_history`. Familia must ship a secret-history mechanism (`verifiable_identifier_secret_history` or equivalent) so identifiers minted under prior keys remain verifiable after a signing-key rotation.

**Prerequisites before enabling verification-on-read:**
- Upstream familia ships secret-history for `VerifiableIdentifier`
- OTS populates history with any prior keys (committed fallback, empty string if used, rotated values)
- Decide rollback story (verification off → on is one-way once enforcement begins)

**Out of scope for v0.27:** no upstream mechanism exists. #3630 tracks this work.

Issue #3630, opened 2026-07-04
Derive IDENTIFIER_SECRET fallback at boot so familia 2.11+ upgrade can't strand or silently weaken identifier signing


delano:
## Status

> **⚠️ This issue stays OPEN by design — it is the sentinel for the familia 2.12 follow-up integration.**
> The boot-time mitigation below is **implemented and pushed** (branch `claude/xchach20-poly1305-upgrade-xqekdg`, commit `dcc0e28`), so the familia 2.11 upgrade no longer crashes or silently weakens identifier signing. But this ticket must **not** be closed on that fix alone: it tracks the eventual switch to verification-on-read once the familia 2.12 work lands. Close only when that integration is done (or explicitly de-scoped). See "Remaining follow-up" at the bottom.

## Problem

Secret and Receipt objids are minted via `Familia::VerifiableIdentifier` (`lib/onetime/models/secret.rb:15`, `lib/onetime/models/receipt.rb:15`), which HMAC-signs identifiers with `VERIFIABLE_ID_HMAC_SECRET`. Two upgrade hazards land when we move to familia >= 2.11 (needed for the XChaCha20 upgrade + the #310 security fixes):

1. **Unset var → crash after upgrade.** `IDENTIFIER_SECRET` is optional today (commented out in `.env.example`; only `init.rake` ever writes it). On familia <= 2.10 a missing secret silently used a committed fallback key (familia #310 S1 — a live weakness). familia 2.11 removed that fallback, so an install without the var now **raises `KeyError` at first secret creation**.

2. **Empty var → silent weak key.** The compose files pass `IDENTIFIER_SECRET=${IDENTIFIER_SECRET:-}` (`docker/compose/docker-compose.full.yml:71`, `docker-compose.simple.yml:32`), injecting an **empty string** when unset. The old passthrough bridge propagated it to `VERIFIABLE_ID_HMAC_SECRET`, and familia's `ENV.fetch` accepts a present-but-empty value — so identifiers got HMAC'd with an empty key: no crash, no forgery protection. (Upstream companion fix: delano/familia#335.)

## Fix (implemented — commit `dcc0e28`)

`ConfigureFamilia#execute` now derives `VERIFIABLE_ID_HMAC_SECRET` from `site.secret` via the existing `KeyDerivation` infrastructure whenever `IDENTIFIER_SECRET` is unset **or blank**, and uses an explicitly-set value verbatim:

```ruby
[...]

PR #4125, opened 2026-08-11
chore: record ancestry for two already-shipped branches (#3630, #3959/#3961)

delano:
Closes out two stale branches by recording the ancestry link that their
original landings never created. **Zero net diff against `main`** — this is a
topology change, not a content change.

## Why the graph showed them unmerged

Both branches' content is already on `main`, but by commits with no ancestral
relationship to the branch tips, so `git log --graph` renders them as dangling
stubs.

**`feature/billing-checkout-link`** — the bulk merged normally via #3959 at
`a860094d7`. Afterwards one more commit (`af3688684`, hide the checkout-link
button when billing is disabled) was added on top of that same tip. Its content
was re-authored on `fix/admin-checkout-link-billing-guard` and landed as
`bb36606e1` via #3961, so the branch tip itself never merged — the lane
converges into `main` and then continues one node past the junction and stops.

**`feature/3630-familia-212-reader-prep`** — never PR'd at all. The familia 2.12
reader bump was hand-reapplied and landed on `main` as `f580fa7333` (same commit
message minus the `[#3630]` prefix), an independent commit with no shared
ancestry.

## Conflict resolution

One conflict, in `lib/onetime/initializers/configure_familia.rb`, purely in the
`encryption_personalization` comment block. Resolved to `main`'s text, which is
a strict superset of the branch's: same rotation semantics, plus the
cross-reference to #3987 and the `'FamilialMatters'` compatibility note.

`Gemfile` and `Gemfile.lock` auto-merged cleanly to `main`'s pins. Worth stating
explicitly because a branch-vs-`main` *diff* looks alarming here: the branch
predates #4024 and shows `otto ~> 2.5` against `main`'s `~> 2.8`. It never
touched the otto line, so that side is unchanged from the merge base and
`main`'s 2.8 wins uncontested. Post-merge: `otto ~> 2.8`, `familia ~> 2.12`.

## Ticket status

- #3630 stays **open** — it tracks the `IDENTIFIER_SECRET` fallback derivation,
  not the reader bump, and the branch had nothing left to contribute to it.
- #3987 (write-side rotation, #3630 phase 2) is unaffected. Its precondition is
  unchanged: every process reading the datastore must be on the 2.12 reader
  floor before the write-side flip.
- #3840 is already closed; its branch `feature/3840-phase4-shared-primitive` is
  a true ancestor of `main` (the ref sits on merge commit `d95a2d583` from
  #3878) and needs no merge — just deletion.

## Merge method

**Merge commit, not squash.** Squashing discards the second-parent links, which
are the entire point of this PR, and would reduce it to an empty commit.

## Testing

None run. The merge resolves to a byte-identical tree with `main`
(`git diff origin/main HEAD` is empty), so there is no behavior to exercise.

## After merge

All three branches are then safe to delete:

```
git push origin --delete \
  feature/3630-familia-212-reader-prep \
  feature/billing-checkout-link \
  feature/3840-phase4-shared-primitive
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -74,15 +74,20 @@ gem 'truemail', '~> 3.3'
 # ORMs and database drivers
 # NOTE: We install both db drivers for the OCI images so that users can choose
 # which database to use at runtime via environment variable without rebuilding.
-# familia 2.11.2 floor: rejects a blank VERIFIABLE_ID_HMAC_SECRET at the library
-# layer (delano/familia#335); the 2.11 line decouples the AES-256-GCM HKDF salt
-# from the XChaCha20 personalization -- which activates the salt/personalization/
-# history pinning in ConfigureFamilia; and 2.11.2 stops persisting nil declared
-# fields as the JSON string "null" (HDEL on clear), restoring HSETNX/HEXISTS
-# atomic-claim semantics (no migration -- stale "null" decodes to nil on read).
-# Do NOT relax to 2.12: it lands breaking encryption personalization/salt-history
-# changes (delano/familia#333, #334) that need the migration tracked in issue #3630.
-gem 'familia', '~> 2.11.2'
+# familia 2.12 floor: read-side prep for the encryption rotation tracked in
+# issue #3630. 2.12.0 adds encryption_personalization_history (delano/familia#333)
+# so XChaCha20 decrypt walks current -> history -> the library default
+# 'FamilialMatters', making personalization rotatable without stranding old
+# envelopes; per-field `algorithm:` pinning (delano/familia#334, shipped 2.11.1)
+# plus envelope-driven provider dispatch means readers accept both aes-256-gcm
+# and xchacha20poly1305 ciphertext with zero configuration. This release changes
+# nothing on the write side -- ConfigureFamilia still pins the 2.11-era values
+# byte-for-byte; the write-side rotation flip lands in v0.27.0 and requires
+# every process reading the datastore to be on >= this release first.
+# (2.11.2 remains the behavior floor for blank VERIFIABLE_ID_HMAC_SECRET
+# rejection, delano/familia#335, and for nil declared fields persisting as HDEL
+# instead of the JSON string "null".)
+gem 'familia', '~> 2.12'
 gem 'pg', '~> 1.6'
 gem 'sequel', '~> 5.0'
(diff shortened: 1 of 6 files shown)


delano:
Closing unmerged. The only thing this PR would add to `main` is two
second-parent pointers — the tree is byte-identical either way — and that
isn't worth a merge commit plus the CI awkwardness of a zero-diff branch.

Leaving the findings here so the next person looking at the graph doesn't
have to re-derive them.

**All three branches are fully superseded and safe to delete.** The repo uses
merge commits rather than squash, so the usual "squash broke the link"
explanation doesn't apply — each branch breaks graph connectivity a different
way:

| Branch | Ancestor of `main`? | Where the content actually is |
| --- | --- | --- |
| `feature/3840-phase4-shared-primitive` | **yes** | ref sits on merge commit `d95a2d583` (#3878); reads as unmerged only because the branch was fast-forwarded past its own merge, so there's no topic lane left to draw |
| `feature/billing-checkout-link` | no | bulk merged at `a860094d7` via #3959; the one later commit `af3688684` was re-authored elsewhere and landed as `bb36606e1` via #3961, leaving the label one node past the junction |
| `feature/3630-familia-212-reader-prep` | no | never PR'd; hand-reapplied and landed as `f580fa7333` — independent commit, no shared ancestry |

Ticket status is unchanged by closing this:

- #3630 stays open on its actual subject, the `IDENTIFIER_SECRET` fallback
  derivation. The reader bump it was named after already shipped.
- #3987 (write-side rotation, #3630 phase 2) is unaffected — `main` carries the
  familia `~> 2.12` reader floor its precondition depends on.

One thing worth carrying forward from the merge attempt: the branch-vs-`main`
diff on `feature/3630-familia-212-reader-prep` shows `otto ~> 2.5` against
`main`'s `~> 2.8`, which looks like a regression waiting to happen. It isn't.
The branch predates #4024 and never touched the otto line, so in a three-way
merge `main`'s side wins uncontested. Confirmed by actually running it here —
the merge resolved to `otto ~> 2.8`, `familia ~> 2.12`, zero net diff.

Cleanup:

```
git push origin --delete \
  feature/3630-familia-212-reader-prep \
  feature/billing-checkout-link \
  feature/3840-phase4-shared-primitive
```

Issue #4137, opened 2026-08-11
webauthn: support passkeys on custom domains via per-domain credential scoping


delano:
## Summary

Passkeys do not work on custom domains today, and `restrict_to='webauthn'` is rejected for domain signin configs. Per the ADR-024 amendment (2026-08-11), this is **pending support, not a ban**: there is no security, privacy, or UX objection to custom-domain passkeys — the constraint is that our credential store is not domain-scoped. This issue is the scoped work that lifts the guards.

## Mechanism

WebAuthn credentials are origin-bound: an assertion succeeds only when the request host falls within the credential's `rp_id` scope (registrable-suffix match, enforced by the browser — this is the protocol's phishing resistance, a feature to build with, not a bug to work around).

Current state:

- `webauthn_rp_id` is already dynamic — `request.host` (`apps/web/auth/config/features/webauthn.rb`). Ceremony-time rp_id selection is not the gap.
- All real-world registrations happen on the canonical host (passkey management lives in account settings), so every stored credential carries the canonical rp_id.
- `account_webauthn_keys` / `account_webauthn_user_ids` (Rodauth defaults) store **no rp_id/origin column**. Credentials cannot be filtered by host, so on a custom domain the allow-list would offer credentials that can never assert there, and nothing can be registered *for* that host through any existing surface.

## Design direction: per-domain credential scoping

1. Schema: add an rp_id (or domain reference) column to `account_webauthn_keys`; backfill existing rows with the canonical host.
2. Filter challenge/allow-list generation to credentials whose rp_id matches the current request host, in both the login ceremony and autofill.
3. Registration UX: an explicit "add a passkey for this domain" surface, since a canonical-host passkey cannot be reused. Where this lives depends on the account↔domain identity binding work (see below).
4. Empty-state handling: a domain with `restrict_to='webauthn'` and zero credentials registered for that host must render fail-closed per the ADR-024 amendment degradation policy, not a dead ceremony.

Interaction: this dovetails with account↔domain identity binding (companion issue, filed alongside this one). Domain-scoped accounts make per-domain credentials the natural shape — an account that exists only within an org's domain registers its passkeys under that domain's rp_id and the scoping falls out. Sequencing the two together will avoid building registration UX twice.

## Rejected alternative: Related Origin Requests

[...]

PR #4155, opened 2026-08-13
Enforce restrict_to as access control across every sign-in surface

### LINKED ISSUE #4139: Enforce restrict_to as access control per ADR-024 A1-A4
## Problem Statement

PR #4130 shipped the display half of domain `restrict_to`: the server renders a single sign-in method for a restricted host, but accepts crafted POSTs for every other method — there is no server-side enforcement and no server-side re-validation. Those gaps were recorded deliberately, pending decisions; the decisions are now ratified as ADR-024 amendments A1–A4. Nothing tracks the implementation work — this issue does. Until it ships, domain `restrict_to` must not be documented as an access control.

## Decided scope (per ADR-024 amendments, 2026-08-11)

**A1 — enforcement.** When resolution yields a single method for a request host, the server MUST reject submission of every other method on that host, crafted POSTs included. Corollary: a disabled auth method must never function even when fully configured — configuration presence is never availability.

**A2 — model-owned resolver.** `restrict_to` resolution moves behind `SigninConfig.resolve_restrict_to(global, config)`, consumed by all three gates: display (`ConfigSerializer`), runtime (`Core::Controllers::Base`, once A1 ships), and settings-API `details` (A4). No caller re-derives. `ConfigSerializer#resolve_restrict_to` (today an inline display-only implementation) becomes a consumer.

**A3, domain half — fail-closed degradation.** A domain restriction whose backing method is unavailable resolves to *sign-in unavailable*, never widening to standard mode. Display side shipped in #4123/#4130; runtime side lands with A1.

**A3, global half — boot validation, fail closed (ratified 2026-08-11).** `AuthConfig#restrict_to` currently returns nil silently when the named method's prerequisites fail (`lib/onetime/auth_config.rb`, `#restrict_to`), and every caller reads nil as "unrestricted" — the install widens to standard mode with no signal, re-exposing the methods the operator restricted away. New behavior:

- Unavailability determinable at boot (e.g. `restrict_to: sso` with SSO disabled or zero providers) is a **fatal boot error** — fail loud, fail closed. A boot failure surfaces at deploy time to the operator who holds the config file; the trap to avoid is the silent runtime widen, not a refused boot (#4062's lockout-trap argument was considered and overruled for this path).
- Unavailability only discoverable after boot degrades fail-closed, same as the domain path.
- The drop-to-standard fallback is retired entirely.

**A4 — serialize resolved truth.** Settings-API `details` gains `effective_restrict_to` (A2 resolver output) alongside `global_restrict_to`; the frontend fallback shipped in #4130 repoints at it and client-side re-derivation is removed.

## Current state

- The only server-side teeth anywhere is *global* `restrict_to='sso'` via `SsoOnlyGating` (account-management operations, keyed to the global value).
- No boot-time validation or warning exists for global `restrict_to`; the nil fallback is fully silent.

## Affected Areas

- [ ] `lib/onetime/auth_config.rb` — `#restrict_to` silent-nil fallback retired; boot-time validation added
- [ ] `lib/onetime/models/custom_domain/signin_config.rb` — home of the A2 resolver `resolve_restrict_to(global, config)`
- [ ] `apps/web/core/views/serializers/config_serializer.rb` — inline `#resolve_restrict_to` becomes a resolver consumer
- [ ] `apps/web/core/controllers/base.rb` — runtime enforcement gate (A1)
- [ ] `apps/api/domains/logic/signin_config/base.rb` / `get_signin_config.rb` / `put_signin_config.rb` — `details` gains `effective_restrict_to` (A4)
- [ ] `src/shared/composables/useSigninConfig.ts`, `src/schemas/api/domains/responses/signin-config.ts` — repoint fallback at `effective_restrict_to`, delete client re-derivation (A4)

## Out of scope

- WebAuthn on custom domains (per-domain credential scoping): #4137
- Account↔domain identity binding (`SsoOnlyGating` re-keying to owning-org policy): #4138

## References

- ADR-024 amendments A1–A6: `docs/architecture/decision-records/adr-024-custom-domain-auth-override-resolution.md`
- PR #4130 (display half + known-gaps record)

delano:
Closes #4139.

Implements ADR-024 A1–A4: `restrict_to` becomes access control rather than a
display preference, and the runtime gate, the sign-in page, the invite page and
the domain settings API all answer the same question the same way.

## What was wrong

`restrict_to` narrowed which sign-in methods a page *offered*, but the routes
behind the hidden methods stayed reachable — a restricted host could still be
signed into by a method it did not offer. Closing that surfaced a second class
of bug: availability was only ever asked at global scope, so a custom host
inheriting a restriction to a method it cannot actually run (password and
email-auth need the domain's sign-in opt-in, SSO needs tenant credentials or the
platform fallback, WebAuthn is not scoped per custom host yet — #4137) resolved
`:restricted` and advertised a sign-in path that could not complete.

## What changed

**Enforcement (A1/A7).** Every pre-auth Rodauth route is classified by the
sign-in method it mints a session with, and a route whose method the host
restricts away 404s — indistinguishable from a route that does not exist.
Includes the account-verification autologin, so a key minted on an unrestricted
host cannot replay on a restricted one. `create_account` / `verify_account` /
`verify_account_resend` are classified as WebAuthn routes when Rodauth's
`webauthn_verify_account` feature is loaded, since that feature changes the
ceremony without changing the route names.

**One resolver, one availability gatherer (A2).** `SigninConfig.resolve_restrict_to`
owns precedence; `SigninConfig.restriction_available_for_request?` owns the
post-boot availability rule, including narrowing an inherited global restriction
through the custom host's own capabilities. Five consumers ask it — the full-mode
route gate, the simple-mode controller gate, the display serializer, the invite
logic and the domain settings API — so none can drift. A parity spec asserts
three-way `to_wire` equality across those consumers.

**Fail closed on unavailable, without manufacturing restrictions (A3).** A
restriction whose backing method cannot run resolves `:unavailable` and offers
nothing. Datastore failures on the custom-host hot path degrade on the *global*
value (in-memory config, so it cannot itself have failed): fail closed when a
restriction is in force, stay unrestricted when the install restricts nothing —
otherwise a transient blip would take every gated route on every custom domain
to 404. The degrade path logs its own `restrict_to_domain_lookup_degraded` event
because it enforces nothing and is otherwise silent.

**Display parity (A4).** `features.restrict_to` is a string-or-null scalar and
could not express `:unavailable`, which serialized as null — standard mode — so
the page rendered every method on a host whose gate rejects all of them. The
bootstrap now carries `effective_restrict_to` (`{ state, restrict_to, source }`)
alongside the legacy scalar, and `AuthMethodSelector` fails closed on the state.

**Invite page.** `ShowInvite` advertised tenant SSO from the stored
`sso_config.enabled?` flag, which outlives the route. It now asks the same
runtime ladder the sign-in route asks and falls back to the platform providers
when the host permits SSO but the tenant arm does not, so an sso-restricted
domain cannot hand an invitee a single dead method.

**Domain settings form.** The SSO-lockout confirmation fired on
`sso_not_permitted` caused solely by the current config's `sso_enabled=false` —
a state that the same atomic PUT fixes. Warning on it trained operators to click
through the one prompt that matters; it now guards only causes the PUT will not
resolve.

## Known gap (recorded, not fixed)

The settings API does not apply the `'sso'` host pin that the display serializer
and the route gate both apply, so it reports `unrestricted` for an SSO-only
tenant where the two request-time consumers report `restricted/sso`. Same drift
shape, different input (the `global` value rather than `available:`), and closing
it changes what the settings page shows for every SSO-only tenant — out of scope
here. Pinned as a pending example in the parity spec so it reds the moment
someone fixes it.

## Tests

| Suite | Result |
|---|---|
| `try/unit/models/custom_domain_auth_killswitch_try.rb` | 93 passed, 0 failed |
| `try/unit/models/custom_domain_auth_default_off_try.rb` | 123 passed, 0 failed |
| `apps/web/core/spec` | 392 examples, 0 failures, 1 pending |
| `apps/web/auth/spec` | 1624 examples, 0 failures, 91 pending |
| `spec/unit` | 3927 examples, 0 failures, 22 pending |
| `apps/api/domains/spec` + `apps/api/invite/spec` | 556 examples, 0 failures, 11 pending |
| `spec/integration/simple/restrict_to_enforcement_spec.rb` | 20 examples, 0 failures |
| `apps/web/auth/spec/integration/full/restrict_to_enforcement_spec.rb` | 18 examples, 0 failures, 4 pending |

Frontend: `AuthMethodSelector`, `AcceptInvite`, `DomainSigninConfigForm` and the
bootstrap contract specs updated alongside their components.


delano:
re: "[CI / T1 · i18n Validation (pull_request)](https://github.com/onetimesecret/onetimesecret/actions/runs/31742792074/job/94590096879?pr=4155) Failing after 10s", we'll update those in a follow-up.

delano:
The visual failure is a non-issue. A slight change in vertical empty space when "Show More" appears. Happens intermittently and has nothing to do with branch changes.

Screenshot 2026-08-13 at 14 31 36

PR #4159, opened 2026-08-13
Decompose ADR-024, adopt ADR-036 living-doc format

### LINKED ISSUE #4158: Decompose ADR-024 to remove cruft and capture valuable docs
A substantial amount can leave this ADR. It currently combines four things:

1. the original UI/storage decision;
2. later `restrict_to` policy decisions;
3. implementation notes and debugging history;
4. ticket/status tracking.

Only the first two belong in durable decision records.

### Transient implementation status

Remove statements describing what had or had not shipped at a particular moment:

- `A1` lines 177–181: “Current state, for the record: display-only…”
- “once A1 ships” in `A2`
- PR sequencing such as “runtime side lands with A1”
- “frontend fallback introduced in #4130”
- “when A1 lands”
- guards that “#4137 retires”
- similar “follow-up work” language

Tickets and repository history are better sources for rollout status.

### Investigation chronology

Remove phrases such as:

- “Found while implementing…”
- “Settled 2026-08-11”
- “ratified 2026-08-11”
- “Recorded for posterity”
- “The tell was…”
- “written confidently and was still short four entries”

These explain how the team discovered a decision, not the decision or its rationale.

### Low-level implementation discoveries

Most of `A7 implementation findings` at lines 312–330 should move to code comments, tests, or implementation documentation:

- frozen `route_hash`
- `post_configure`
- `@current_route`
- middleware placement
- exact callback hook names
- simple-mode routing mechanics

The ADR only needs the resulting constraint:

> Enforcement occurs per request and must cover Rodauth, OmniAuth, and simple-mode entry points.

Likewise, most of `A10` lines 419–438 is an implementation coverage checklist. Preserve it executably in the coverage spec, not as prose that can drift.

### Ticket-specific implementation prescriptions

Remove details such as:

- exact model columns proposed by `#4137`
- exact hook or source file names
- precise line-of-code placement such as “first line of `raise_concerns`, before the rate limiter”
- internal identifiers such as “H-3 refusal”
- lengthy walkthroughs of particular call paths

These belong in ticket acceptance criteria or code documentation.

### The editorial paragraph in A9

Lines 388–393 can go. The normative rule is already clear:

> Invalid or conflicting access-control configuration must fail at boot rather than silently become unrestricted.

The “three independent silent fail-opens” retrospective adds emphasis but no new decision.

### “Custom-domain default-OFF is implemented twice”

Lines 118–129 are important but too implementation-specific for the original ADR.

Keep only:

- custom domains default off;
- tenant SSO and platform-fallback SSO intentionally have different visibility rules;
- `AUTH_ENABLED` disables both;
- `AUTH_SIGNIN` disables password/email sign-in but not the SSO carve-out.

Remove the current method names, hook flow, route-registration discussion, and issue chronology.

Better still, this may deserve a separate ADR because it is an SSO surface policy, not a single-control settings-UI decision.

### A3

Keep:

- unavailable restricted methods fail closed;
- no degradation to unrestricted/standard mode;
- boot-detectable invalid global restrictions cause boot failure;
- runtime unavailability remains unavailable.

Remove:

- the historical webauthn population argument;
- references to releases and PUT carry-over behavior;
- the long rebuttal of `#4062`;
- descriptions of the old implementation returning `nil`.

### A5

Keep only the policy:

> Custom-domain WebAuthn is supported in principle but unavailable until credentials can be scoped to the applicable RP ID. Until then, attempts to configure or use it fail closed.

The browser-limit research and proposed schema belong in `#4137`.

### A6

A6 is a distinct architectural decision about identity tenancy. It should become its own ADR rather than an amendment here.

If retained temporarily, reduce it to:

- identities are scoped to their owning organization/domain;
- canonical identities cannot authenticate on arbitrary customer domains;
- account-management policy follows identity ownership, not the current request host.

The Keycloak/Slack/Notion/Google comparison is optional rationale and can be removed.

### A7 and A10

Merge these into one concise enforcement-scope section:

- restricted pre-auth methods behave as unavailable and return 404;
- all pre-auth entry points and ceremony phases must be covered;
- authenticated account-management and second-factor completion are account-scoped and exempt;
- coverage is enforced through an exhaustive route/surface contract test.

Remove the endpoint-by-endpoint discovery narrative. The test is the durable enumeration.

### A11

Keep the resulting rule:

> Invite signup is a pre-auth password account-creation surface and is subject to `restrict_to`; invitation display and authenticated acceptance are not.

Remove the detailed orphan-account and SSO-callback narrative unless that behavior is otherwise undocumented. It is implementation rationale, not a new architectural principle.

## Move to separate ADRs

The amendment has grown beyond the original title. I would split out at least:

1. **Authentication method restriction resolution and fail-closed enforcement**
   - A1–A4
   - A7–A11
   - operator/domain intersection
   - boot-error policy
   - pre-auth versus account-scoped boundary

2. **Tenant identity and authentication-policy scope**
   - A6

3. Optionally, **Custom-domain WebAuthn RP-ID scoping**
   - A5, if `#4137` will make a durable schema/security decision

Then ADR-024 can return to its original subject:

- two-flag storage model;
- `enabled` as the explicit-configuration pin;
- model-owned effective resolution;
- API-provided effective state;
- single-control UI;
- reset semantics.

## Possible removal from the original decision

Even lines 102 and 131–155 are implementation inventory rather than architecture:

- exact composable and component names;
- exhaustive source-file list;
- exhaustive test-file list.

A short “Implementation references” link to the principal resolver and contract test is enough. Git search and ownership tools are better than an ADR for maintaining a file inventory.

The JSON response example can also be shortened to the stable API semantics without comments or every current field.

## Recommended final structure

ADR-024 could be reduced to roughly:

1. **Context**
   - two stored flags caused contradictory UI and ineffective writes;
   - explicit configuration must remain distinguishable from inheritance.

2. **Decision**
   - `enabled` is an internal pin;
   - one visible control;
   - writes pin and DELETE unpins;
   - model resolver owns effective state;
   - API returns raw and effective state;
   - UI displays effective state.

3. **Invariants**
   - global kill switch wins;
   - unconfigured inherits;
   - configured values remain pinned;
   - all gates use the same resolver.

4. **Consequences**
   - legacy `enabled=false` behaves as unconfigured;
   - GET returns `record: null`;
   - API compatibility impact.

Everything from `## Amendments` onward should either move to the new restriction-enforcement ADR or be deleted as implementation history. That is the clearest boundary.

delano:
Splits the sprawling ADR-024 (custom-domain auth override resolution) into
focused, current-state documents, moves ADR authoring to the ADR-036
living-doc format, and re-homes all ADRs from `docs/decision-records` to
`docs/adr`. Adds `scripts/adr-lint.rb` (wired into pre-commit) to enforce
the new format and catch stale cross-references.

Also includes a couple of small fixes uncovered along the way: stale
auth-policy tryout expectations, a locale hash refresh for #4155, and an
import-order cleanup in AcceptInvite.vue from eslint --fix.

Resolves #4158


delano:
> * 2. **`restrict_to_allows?` in `Core::Controllers::Base:236–244` — double raise on a Redis error during the availability probe for an operator host.** `signin_policy_read_failed!(ex)` returns `nil` for operator hosts, then `resolve_lookup_failure` calls `Onetime.auth_config.restrict_to_available?`. That method is in-memory on production (`AuthConfig` reads from the parsed YAML) so a second Redis raise is not reachable today. If a future implementation makes `restrict_to_available?` do a live read, the resulting `Redis::BaseError` would escape the rescue block as an unhandled 500 on the canonical sign-in page rather than as a `SigninPolicyUnavailable` 503. Not a current defect — flagging only because the comment at line 196–205 explicitly says "an in-memory config that cannot have failed", and that invariant is currently held by convention not enforcement.

base.rb:236-244 — Confirmed correct as written. restrict_to_available? → restrict_to_unmet_prerequisite (auth_config.rb:515-530) checks only sso_enabled?, sso_providers, email_auth_enabled?, webauthn_enabled? — all in-memory config reads, no Redis calls anywhere in that path today. So the second raise Redis::BaseError the comment worries about is genuinely unreachable right now, and the finding correctly flags it as a convention-enforced invariant rather than a current bug.

Issue #4157, opened 2026-08-13
A misclassified custom domain must not inherit operator treatment


delano:
## Background 

Based on ADR-024. Found while implementing A3's fail-closed rescue (2026-08-13, #4139).

## Problem

`DomainStrategy` answers `:invalid` in two unrelated situations, and the
distinction is invisible at every consumer:

1. the host is genuinely unplaceable (unparseable, or matching nothing);
2. `known_custom_domain?` — a datastore read — **raised**, and
   `choose_strategy`'s `rescue StandardError` returned nil, which `call`
   resolves to `:invalid`.

In case 2 the request is a real customer domain wearing the classification
of an unrecognised one. Roughly ten consumers test `== :custom` and take
their else branch for it. Most of those rows are cosmetic (branding,
favicon). Two are not: `signin_enabled?` and `signup_enabled?` branch on
`custom_domain_request?`, and the two branches have **opposite defaults** —
custom domains are default-OFF and must opt in, canonical follows the
global default. So for the duration of a datastore blip, a custom domain
that never opted into sign-up follows the operator's global default
instead. That is the A3 failure mode (a widen the domain owner did not
choose) re-entered through classification rather than through resolution.

**Normative.** A check whose false branch grants operator treatment to a
host that could be a customer's must be a POSITIVE test against the operator
classifications (`:canonical`, `:subdomain`), never `!= :custom`.
`SigninConfig::OPERATOR_HOST_STRATEGIES` / `operator_host?` is that test and
is the only place it is written.

The carve-out rests on a **one-directional** invariant, and the obvious
stronger version of it is false. `:canonical` is genuinely read-free — the
exact/`www.` arm runs before `known_custom_domain?`. `:subdomain` is not:
its sweeps run *after* that read, and the rescue wraps the whole if/elsif
chain, so a raise aborts the method rather than falling through, and a
subdomain host classifies `:invalid` during a blip. The practical
consequence is that a datastore outage takes subdomain hosts down with the
custom domains while canonical sign-in stays up.

What holds in both cases is that a failure can never *manufacture*
`:canonical` or `:subdomain`. That is the only direction the carve-out
needs: `operator_host?` may be over-strict during an outage, never
over-permissive. A new datastore read placed ahead of the exact canonical
arm in `choose_strategy` would break it silently and take canonical sign-in
dark as well.

**Scope of the #4139 fix, and what remains open.** A3's rescue closes this
[...]

PR #4165, opened 2026-08-14
fix(auth): close the settings-API restrict_to fail-open and repoint stale docs

### LINKED ISSUE #4140: AuthConfig#restrict_to fails open silently when method unavailable
## Problem Statement

An operator who sets global `full.restrict_to` to a method whose prerequisites are not met — `sso` with SSO disabled or zero providers, `email_auth` with email auth disabled, `webauthn` with webauthn disabled — gets no restriction and no signal. `AuthConfig#restrict_to` returns nil, every caller reads nil as "unrestricted", and the install widens to standard mode: the sign-in page and runtime offer **all** enabled methods, re-exposing exactly the ones the operator restricted away. There is no boot error, no boot warning, no log line. The operator believes the restriction is active; it is silently inert.

ADR-024's amendment text originally described this path as "fails loud (boot WARN)" — that warning does not exist and never did (corrected in ef64bc6c7's follow-up, f1c063871f).

## Early Diagnosis

The guard block in `AuthConfig#restrict_to` (`lib/onetime/auth_config.rb`, `#restrict_to`) checks each value's prerequisites and returns nil on failure. It was written as graceful degradation; in effect it is a silent fail-open for an operator-facing security control. Every derived predicate (`sso_only_enabled?`, `password_only_enabled?`, `email_auth_only_enabled?`, `webauthn_only_enabled?`) and every consumer (`SsoOnlyGating`, `ConfigSerializer`) inherits the silent widen. The legacy `sso.sso_only` fallback flows through the same guards, so legacy configs are equally affected.

Decided fix (ADR-024 A3, global half, ratified 2026-08-11): unavailability determinable at boot is a **fatal boot error**; unavailability only discoverable after boot degrades **fail-closed**; the drop-to-standard fallback is retired. Implementation is scoped into #4139 (A2 resolver + A1 runtime gate + boot validation) — this issue documents the live defect and should close with #4139's A3-global slice, or earlier if the boot validation lands independently.

## Affected Areas

- [ ] `lib/onetime/auth_config.rb` — `#restrict_to` prerequisite guards return silent nil; boot-time validation missing
- [ ] `lib/onetime/logic/sso_only_gating.rb` — keys off `sso_only_enabled?`; account-management gating silently disengages with the widen
- [ ] `apps/web/core/views/serializers/config_serializer.rb` — serializes the widened (unrestricted) view to the frontend

## References

- #4139 — enforcement tracking issue carrying the decided fix
- ADR-024 amendment A3 (global half): `docs/architecture/decision-records/adr-024-custom-domain-auth-override-resolution.md`

delano:
Closes #4140.

## Context

Most of #4140 was already fixed on main by 6ac13e848 (#4139 / PR #4155): `AuthConfig#restrict_to` no longer returns nil to mean "could not be honored", `validate_restrict_to!` makes boot-determinable unavailability a fatal boot error, and `restrict_to_available?` carries the separate availability question. What remained was one live fail-open path, plus documentation across the tree still describing the retired behaviour.

## 1. The residual fail-open (the substantive change)

`apps/web/core/spec/views/serializers/restrict_to_parity_spec.rb` carried a pending example recording the gap: for an SSO-only tenant the settings API reported

```
{restrict_to: nil, source: "global", state: "unrestricted"}
```

while the request-time gate was enforcing

```
{restrict_to: "sso", source: "global", state: "restricted"}
```

The client was told sign-in was unrestricted for methods that 404 in practice — the fail-open direction `ADR-034#degradation-is-fail-closed` forbids.

Root cause: the **SSO host pin** — a custom domain with no enabled `SigninConfig`, reachable only via SSO, inherits `restrict_to='sso'` because password and email default OFF there — existed as two private copies (`ConfigSerializer#effective_global_restrict_to`, `Auth::RestrictTo.global_restrict_to`) and one omission. The settings API had neither copy and passed `auth_config.restrict_to` verbatim as both the `global` value and the value keyed for `restriction_available_for_request?`. Two copies plus one omission is the same shape as the four #4139 defects, one input over.

The fix consolidates instead of adding a third copy: `SigninConfig.inherited_restrict_to(config, domain_id:, custom_host:)` owns the policy; the gate, the display serializer, and the settings API all call it. `restrict_to.rb` sheds 26 lines of policy. Intersection semantics are preserved — an enabled config speaks for itself, since an unconditional pin would resolve `:conflict` and lock the host out.

Direction check: nothing was relaxed. The gate is behaviourally unchanged; the settings API moved from `unrestricted` toward the enforced answer.

**One deliberate asymmetry:** simple-mode `Core::Controllers::Base#restrict_to_allows?` keeps no pin. Simple mode serves no SSO route, so pinning `'sso'` there would darken the only working method. Flagging it because it means "one resolver, all consumers" now has a documented exception.

## 2. Operator-facing documentation drift

`.env.reference` promised that conflicting `AUTH_*_ONLY` flags were "ignored (safe fallback to showing all enabled methods)". They are now a fatal boot error. An operator reading the old text expects a misconfiguration to be tolerated when the install will refuse to start — so `.env.reference` and `etc/defaults/auth.defaults.yaml` lead this commit. Also reframes `restrict_to` as an access control rather than a display preference, adds the fail-closed mode to the cannot-log-in runbook, and corrects two frontend comments still describing the backend "nilling" an unhonorable restriction.

## 3. Dead ADR citations

`c8c9f61ca` decomposed ADR-024's A1–A12 amendments into ADR-034/ADR-035, leaving ~55 citations naming labels that resolve to nothing. This completes the pass 4cccdb35e started for A12. Mapping was derived from the decomposition commit and current ADR text; where an A-number's content could not be located in a live ADR, the citation was left alone rather than given an invented anchor. Three citations were substantively wrong rather than stale-labelled — detailed in the commit message, including a spec citing "ADR-024 invariant 5" when that list stops at 4.

`scripts/adr-lint.rb`: **930 references across 38 ADRs, all resolve** (was 861).

## Verification

Branch-touched tests, green: parity spec 26/0 (was 25 + 1 pending), `signin_config_restrict_to_spec` 23/0, `config_serializer_spec` 127/0, `restrict_to_gate_spec` 42/0, `auth_config_spec` 122/0, killswitch tryouts 93/0, `AcceptInvite.spec.ts` 35/0. RuboCop clean.

**The full `tests/lanes/run unit` is red, and this is not attributable to the branch.** Pristine `origin/main` in a throwaway worktree reproduces the same failure signatures (billing plan-cache miss, `nil.objid`, org-membership drift) with a *different set every run*: 81/110, 89/53, 20/25 errors across three runs. Cause is a shared test datastore — one valkey container on `127.0.0.1:2163`, DB 0, used by every worktree in the forest with no per-worktree isolation; connection count was observed spiking from ~3 to 649 mid-run while a sibling worktree ran `bin/ots billing diagnose`. Running the identical `--only` subset in both trees gives branch 311 examples/43 failures vs baseline main 303/43 — the same 43, all a pre-existing `NameError: uninitialized constant Auth::Config` in invite specs whose stubs this branch never touches.

Worth a separate issue: the lane has no per-worktree DB isolation, and the boot guard regex at `lib/onetime/boot.rb:180` already permits a non-zero DB index, so a lane overlay would fix it. Not done here.

### DIFF
--- a/.env.reference
+++ b/.env.reference
@@ -441,11 +441,17 @@ AUTH_PASSWORD_REQUIREMENTS_ENABLED=true
 AUTH_ACTIVE_SESSIONS_ENABLED=true
 AUTH_VERIFY_ACCOUNT_ENABLED=true
 
-# Restrict the login page to a single authentication method.
-# Set exactly ONE of these to 'true'. The result maps to
+# Restrict sign-in to a single authentication method. This is an access
+# control, not a display preference: every other method's routes 404 on
+# the restricted host, not just hide.
+# Set at most ONE of these to 'true'. The result maps to
 # full.restrict_to in auth config (values: password, email_auth,
-# webauthn, sso). If more than one is set, all are ignored
-# (safe fallback to showing all enabled methods).
+# webauthn, sso). Setting more than one is a FATAL BOOT ERROR naming
+# every flag set — there is no "restricted to two methods".
+# The named method must actually be available (feature enabled and
+# credentials present); if it is not, that is also a FATAL BOOT ERROR.
+# Post-boot or per-host unavailability degrades to "sign-in
+# unavailable" — never back to showing all methods.
 # If none are set, all enabled methods are shown (default).
 #AUTH_PASSWORD_ONLY=false
 #AUTH_EMAIL_AUTH_ONLY=false
--- a/apps/api/domains/logic/signin_config/base.rb
+++ b/apps/api/domains/logic/signin_config/base.rb
@@ -94,7 +94,7 @@ module DomainsAPI
           # domain. The details.global_restrict_to field below keeps the
           # OPERATOR's own value — the UI labels the inherited operator
           # restriction separately from what resolves for this host.
-          inherited_restrict = Onetime::CustomDomain::SigninConfig.inherited_restrict_to(
+          inherited = Onetime::CustomDomain::SigninConfig.inherited_restrict_to(
             config,
             domain_id: domain_id,
             custom_host: true,
@@ -112,7 +112,7 @@ module DomainsAPI
             # implementation, shared with GET /api/invite/:token) — this app
             # no longer carries a private copy of the wire shape.
             effective_restrict_to: Onetime::CustomDomain::SigninConfig.resolve_restrict_to(
-              inherited_restrict,
+              inherited.value,
               config,
               # Post-boot availability of the global restriction
               # (ADR-034#degradation-is-fail-closed),
@@ -124,11 +124,15 @@ module DomainsAPI
               # capabilities exactly as it is at request time. Without it the
               # page showed `restricted/password` for a domain whose Rodauth
               # routes the gate had already taken to 404 (#4139).
+              #
+              # #4165: pass pin_established so the availability check trusts the
+              # SSO pin's own proof instead of re-consulting AuthConfig.
               available: Onetime::CustomDomain::SigninConfig.restriction_available_for_request?(
-                inherited_restrict,
+                inherited.value,
                 config,
                 domain_id: domain_id,
                 custom_host: true,
+                already_established: inherited.pin_established,
               ),
             ).to_wire,
             tenant_sso: tenant_sso_details(domain_id),
--- a/apps/api/domains/logic/signin_config/base.rb
+++ b/apps/api/domains/logic/signin_config/base.rb
@@ -59,8 +59,9 @@ module DomainsAPI
         # /signin page (same shared authority as
         # Core::Views::DomainSerializer#effective_signin_enabled?).
         #
-        # effective_restrict_to is the A2 resolver's output verbatim (ADR-024
-        # A4), so the settings UI stops re-deriving the effective restriction
+        # effective_restrict_to is the resolver's output verbatim
+        # (ADR-034#settings-api-serializes-effective-restrict-to), so the
+        # settings UI stops re-deriving the effective restriction
         # from global_restrict_to client-side. global_restrict_to stays: it
         # names the *inherited* restriction, which the UI still labels
         # separately from what resolves for this domain.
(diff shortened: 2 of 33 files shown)


delano:
Post-merge-ready review pass complete (independent reviewer, adversarial brief). No changes recommended.

The claim worth checking was "the gate is behaviourally unchanged". Verified by pulling both deleted copies and reducing them against the extracted method:

- old `ConfigSerializer#effective_global_restrict_to` — pins when `tenant_domain?(view_vars) && !signin_config&.enabled? && domain_id != DOMAIN_READ_FAILED`
- old `Auth::RestrictTo.global_restrict_to` — early-returns on `signin_config&.enabled?`, then pins when `domain_strategy == :custom && domain_id && sso_available_for_tenant_host?`

Both reduce exactly to `inherited_restrict_to`'s skip-when-enabled / pin-when-`custom_host && domain_id && sso_available_for_tenant_host?` / else-operator-value. Faithful extraction, not an approximation.

Also confirmed:

- `custom_host: true` in the settings API is a genuine invariant — `authorize_domain_signin_config!` → `authorize_domain_config!` → `load_custom_domain` always resolves a `CustomDomain` before `signin_override_details`, so this logic class never sees a canonical host.
- #4157 tri-state preserved: `restrict_to_resolution` still short-circuits to `:unavailable`/`:domain_read_failed` before the pin is computed, and `Auth::RestrictTo` still uses the non-swallowing `.from_display_domain`, so a read failure raises into `resolve_lookup_failure` rather than routing around the tri-state.
- Every fallback path bottoms out at `Onetime.auth_config.restrict_to`, which post-#4139 can never be a dead restriction (fatal boot error instead). No widening path found.
- The parity spec is a real regression pin — it fails if the settings-API call site is reverted to the verbatim pass-through, rather than asserting a tautology against mocks.

One pre-existing gap surfaced and filed as #4167 rather than folded in: an enabled `SigninConfig` with `restrict_to` unset and `signin_enabled=false` can resolve `:unrestricted` at the gate alongside the operator platform-SSO fallback. It lives in the skip-when-enabled branch both old copies already had, so it is identical pre- and post-commit. Fixing it here would have cost this commit its behaviour-preserving property, which is the basis for merging while the lane is red — and it needs a policy decision about whether "enabled but expressing no restriction" should count as speaking at all.

Lane flakiness filed separately as #4168.

Issue #4063, opened 2026-08-09
Add LINK_DOMAINS: operator-configurable link-domain pool, decoupled from the canonical host


delano:
## Problem

`buildAvailableDomains` unconditionally appends the canonical domain to the domain-context dropdown:

```ts
// src/shared/composables/useDomainContext.ts:94-105
const { canonicalDomain } = getConfig();
const domainNames = storeDomains.map((d) => d.display_domain);
if (canonicalDomain && !domainNames.includes(canonicalDomain)) {
  domainNames.push(canonicalDomain);
}
```

That is correct for the hosted platform. Anonymous users and free accounts with no verified custom domain need *some* domain to mint links with, and without the canonical entry the dropdown is empty and `setContext(canonical)` is a silent no-op (the split-deployment bug fixed in `47ad128790`, still unpushed on `fix/default-domain-canonical-drift`).

It is wrong for self-hosted and single-tenant installs. Our custom installs now get a `ge-abcd123.eu.otshosted.com` canonical host that is an internal platform address, not something a customer wants appearing in a link picker. More generally, an operator may want:

- only their own domain(s) offered, canonical hidden from the picker but still serving the app; or
- a shortlist of operator-blessed alternates (e.g. short-link domains with TLS already provisioned on the server) offered *alongside* or *instead of* canonical.

There is no way to express any of that today. Canonical is hardcoded into the pool.

## Proposal

Add `LINK_DOMAINS` (comma-separated) → `features.domains.link_domains`. Keep `DEFAULT_DOMAIN` exactly as it is.

Two derived sets, one setting:

| Set | Purpose | Contents |
|---|---|---|
| Canonical set | request classification / admission | `features.domains.default` + `site.host` + `link_domains` |
| Link pool | the dropdown | `link_domains` if set, else `[canonical_domain]` |

Hiding canonical from the picker therefore does not affect whether we serve on it. That separation is the whole point.

### Why `LINK_DOMAINS` must feed classification, not just the UI

`DomainStrategy::Chooserator.choose_strategy` (`lib/onetime/middleware/domain_strategy.rb:236-284`) resolves a host through: exact canonical match → registered `CustomDomain` → peer/parent sweep → subdomain sweep → `nil`. A host that is none of those returns `nil` and becomes `:invalid`.

[...]

PR #4195, opened 2026-08-17
Restrict version display to authenticated sessions

delano:
## Summary

Enhance security by hiding the deployment version string from unauthenticated users. Version information is now only displayed to authenticated sessions, reducing the attack surface by preventing fingerprinting of the installation and targeting of known CVEs.

## Changes

- Modified `useFooterConfig()` composable to add a new `showVersion` computed property that combines the UI configuration setting with authentication status
- Updated all footer components (WorkspaceFooter, ManagementFooter, TransactionalFooter, BrandedFooter) to use the new `showVersion` property instead of `showVersionConfig`
- Removed unused version-related imports and state from BrandedFooter
- Added test coverage for the new authentication-based version visibility logic

## Test Plan

Added unit test verifying that version link is hidden for unauthenticated sessions even when `show_version=true`. Existing tests updated to include `authenticated: true` in bootstrap state to maintain current behavior for authenticated scenarios.

https://claude.ai/code/session_01BaNzTTtHRYP6uYhocvFWj7


delano:
Review follow-ups resolved in 3a5119561, 0011721a4, 689860caa.

**1. Three specs asserted `version` on `GET /auth` (blocking, @claude)** — fixed in `3a5119561`. Rather than deleting the assertions, they're inverted: `availability_spec.rb`, `rodauth_spec.rb` and `advanced_auth_mode_spec.rb` now each pin the service message, assert the `version` key is absent, and assert the raw body carries no build version string, matching `auth_enabled_kill_switch_spec.rb`. The examples were renamed — "includes version information" was describing the opposite of what it now checks. Verified: `tests/lanes/run full-sqlite --only <the four /auth specs>` → 78 examples, 0 failures; same set on `full-pg` → 78/0. Swept `spec/ try/ tests/` for other stale `version` assertions against the `/auth` root and for every bare `/auth` call site; no others exist (`/auth/health` and `/api/v*/version` legitimately still carry it).

**2 + 3. `bundle exec rspec` in run instructions (@Copilot, @claude)** — fixed in `0011721a4`, and extended to `page_bootstrap_me_spec.rb:12`, which this PR touches and which was worse: `source .env.test && bundle exec rspec`, literally the ambient-env leak the lane runner exists to prevent. Both now document `tests/lanes/run unit --only <path>`; both invocations were run to confirm they work (8 and 51 examples, 0 failures).

**4. `attr_reader :version` in `apps/api/v3/logic/feedback.rb:31` (@claude)** — **not applying this one; it's a false positive.** The reader is load-bearing, and by this PR's own doing: `try/unit/v3_feedback_try.rb:87` reads `logic.version` to assert the value is stamped from `OT::VERSION.details`, and `:94` reads it again to assert a spoofed `99.99.99-evil` param is ignored. Those two tryouts are the coverage for the exact security property this PR adds. There's no inherited definition to fall back on (no `def version` or `attr_reader` for it in `apps/api/v3/logic/base.rb`, `apps/api/v2/logic/base.rb`, or `lib/onetime/logic/*.rb`), so dropping it raises `NoMethodError`. Every other `ReceiveFeedback` consumer was checked: the route table, `enqueue_email_locale_normalization_spec.rb` (uses `.allocate` + `send(:send_feedback)`), and the OpenAPI/request-schema generators — none read `.version`, and no `version` appears in the serialized response.

**Visual regression (27 shots)** — 21 regenerated in `689860caa`. These are real and expected: the anonymous footers lost the version line, so 15 shots lose a 14px band, home/receipt canonical mobile lose 16px of footer height, and the 3 secret canonical mobile shots lose 24px where `BrandedFooter`'s version line was removed outright. Regenerated per the throwaway-empty-Valkey recipe (a seeded store leaves a stale secret-verifier that fails the `secret--revealed` cells with a 500 that reads as a product bug): update run 52/52, verify run 52/52 against a freshly emptied store, all output `*-linux.png`.

The other 6 of the 27 were the inherited #4180 drift from #4199 — already fixed on `main` in `590fd3c7e`, and this branch has since been rebased onto it, so they're not in this changeset. 27 − 6 = 21 reconciles exactly.

One caveat worth a reviewer's eye: `home--default--canonical` desktop carries a **second** diff region unrelated to this PR — the domain badge drops its port (`canonical.example.org:7143` → `canonical.example.org`). That's #4063 drift: `normalizeDomainHost` has stripped the port since `a819493e2e` (2026-08-09), and this baseline hadn't been regenerated since 2026-07-15, while its mobile counterpart absorbed the change on 2026-08-10. Baking it in here is correct, but it isn't a change this PR made.

Issue #4223, opened 2026-08-20
Set request.host via PublicHostRewrite boundary middleware + forwarded-authority sanitization


delano:
## Problem: the two-host-truth seam

Behind the custom-domain proxy, the raw `Host` header is always the canonical origin; the public hostname travels in a forwarded header and is resolved by `Rack::DetectHost` + `DomainStrategy` into `env['onetime.display_domain']` / `env['onetime.domain_strategy']`. Every consumer that reads `request.host` / `HTTP_HOST` instead of the resolved keys picks the wrong truth, and the wrong truth is the ergonomic default — it's what every gem and every naive line of app code reaches for.

This seam has now produced three incidents:

- #4170 — `HttpOrigin` bypassed DetectHost, 403ing custom-domain `/auth/*` POSTs
- 2026-08-05 prod incident — trusted-proxy trust-signal coupling made all custom domains classify `:canonical` (fixed in PR #4021, depth mode in #4024)
- #4221 — Rodauth mints every transactional-email link (magic link, password reset, verify account, sso-link-confirm) on the canonical host; broken since the proxy architecture was introduced
- 2026-08-19 nz-prod incident — tenant SSO initiation on a custom domain 302s to `/signin?auth_error=sso_not_configured`: `omniauth_tenant.rb` keyed its `CustomDomain` lookup on `request.host`, which had been propped up by an `X-Forwarded-Host` the 08-14 edge hardening correctly deleted. Same request's `DomainStrategy` classified `custom`. Tactical fix (`public_host` helper reading `onetime.display_domain`) on `fix/tenant-sso`.

Per-consumer fixes (#4221's `base_url` override) stop individual bleeds but cannot protect the next mounted gem. App-side lint (RuboCop cop on `request.host`) can't see inside gems either — and gems have been the repeat offenders. The only fix that eliminates the class is making the correct host the default at the boundary.

## Aggravating finding: Rack 3.2.7 forwarded-authority is client-poisonable today

Rack 3.2.7 resolves `request.host` via `authority = forwarded_authority || host_authority || server_authority` (rack/request.rb:271-275), and `forwarded_authority` honors `Forwarded` and `X-Forwarded-Host` from **any** client, ungated by proxy trust (rack/request.rb:398-413; default `forwarded_priority = [:forwarded, :x_forwarded]`). DetectHost gates forwarded-host trust for `display_domain`, but nothing gates Rack's own authority chain for `request.host`. Consequences:

[...]

Issue #4221, opened 2026-08-20
Rodauth transactional emails mint canonical-host links for custom-domain users


delano:
## Symptom

A user on a custom domain (e.g. `nz.metalbaum.com`) requests a magic link. The email arrives, but the link points at the canonical host (`https://<canonical>/…email-login?key=…`). Clicking it lands on the canonical-host SPA, which POSTs the key to `/auth/email-login` there and receives a 404. Observed in the wild 2026-08-19.

The 404 comes from `Auth::SigninGate` rejecting `:email_auth` on the canonical host (ADR-034 byte-identical not-found) on installs where global signin is disabled and sign-in is offered only on tenant domains. But the 404 is a symptom, not the defect: even on installs where the canonical host accepts the login, the session cookie would be set for the canonical origin instead of the tenant domain the user started from.

## Root cause

Rodauth builds all emailed URLs from the raw `Host` header: `email_auth_email_link → token_link → route_url → base_url → domain → request.host` (rodauth 2.45 `base.rb:562-570`; nothing in our auth config overrides `base_url` or `domain`). Behind the custom-domain proxy, `Host` is always rewritten to the canonical origin — the public hostname travels only in a forwarded header, which `Rack::DetectHost` + `DomainStrategy` resolve into `env['onetime.display_domain']`. Rodauth never consults that key, so every link it mints is canonical-host.

This is the same two-host-resolver split as #4170, on the outbound side.

**Not a regression from the trusted-proxy work** (#4021/#4024/otto 2.9): nothing in any version of the stack rewrites `HTTP_HOST` — DetectHost only publishes `env['rack.detected_host']`. Even with forwarded-host trust fully working (and it was working: the original magic-link request on the tenant domain passed SigninGate's `domain_strategy` check, or no email would have been sent), Rodauth still reads `request.host`. This has been broken for custom-domain users since the proxy architecture was introduced.

## Blast radius

Every Rodauth transactional email mints canonical-host URLs for custom-domain users:

- `apps/web/auth/config/email/email_auth.rb` — magic link (`email_auth_email_link`); also passes `baseuri: request.base_url` and `display_domain: request.host` to the mail template, so the email *branding* names the wrong host too
- `apps/web/auth/config/email/reset_password.rb` — `reset_password_email_link`
- `apps/web/auth/config/email/verify_account.rb` — `verify_account_email_link`
[...]

Issue #4173, opened 2026-08-15
CSP form-action allowlist is boot-global: tenant SSO issuers require hand-maintained env


delano:
## Summary

`AuthConfig#sso_form_action_origins` (`lib/onetime/auth_config.rb:437`) builds the CSP `form-action` allowlist **once at boot, from process-global env**, and applies it to every request (`apps/web/core/application.rb:132`). Tenant-level SSO issuers live in per-domain records, not env, so they are never derived — the only way to admit them is to hand-maintain `SSO_FORM_ACTION_ORIGINS`.

The method's own docstring names the gap:

> the `SSO_FORM_ACTION_ORIGINS` override is merged in unconditionally — it covers sovereign clouds, an OIDC issuer that differs from its authorization endpoint, and **org-level SSO whose issuers are unknown at boot**.

They are not unknown at request time. By the time a signin page renders on a custom domain, the app has resolved the domain, the org, and its `CustomDomain::SsoConfig` — `issuer` for `oidc`, `tenant_id` for `entra_id` (`lib/onetime/models/custom_domain/sso_config.rb:108-109`).

## The two levels, and where each is served

```
                     WHERE THE IdP ORIGIN LIVES          DERIVED TODAY?

  install/platform   ENV: SSO_ENTRA_*, SSO_OIDC_ISSUER…     yes
       level         → active_provider_origins              (gated on sso_enabled?
                                                             + required_vars present)

  domain/tenant      CustomDomain::SsoConfig                 NO
       level         → .issuer / .tenant_id                  → operator must hand-copy
                        (per-domain, encrypted creds)           into SSO_FORM_ACTION_ORIGINS


   boot                                    request
    │                                         │
    ├─ read ENV ──► active_provider_origins   │
    ├─ read ENV ──► SSO_FORM_ACTION_ORIGINS   │
    │                     │                   │
    │                     ▼                   │
    │        merge_csp_directives(            │   display_domain → org → SsoConfig.issuer
    │          'form-action' => "'self' …")   │   known here, consulted nowhere
    │                     │                   │
    └─────────────────────┴───────────────────┴──►  one static header for every tenant
```

## Failure mode

A tenant configures OIDC SSO on their custom domain against `https://idp.customer.example`. Nothing in env mentions it, so:

1. `active_provider_origins` (`auth_config.rb:638`) returns nothing for it — it iterates `provider_definitions` and requires `required_vars` to be present in ENV. Per-domain credentials are not env vars.
[...]

Issue #4170, opened 2026-08-14
Auth app's HttpOrigin mount bypasses DetectHost, rejecting custom-domain /auth POSTs


delano:
## Summary

`Rack::Protection::HttpOrigin` resolves the request host through `Rack::Request`, bypassing `Rack::DetectHost` — the resolver the application mounts specifically because Rack's native host detection is wrong behind a proxy. On any deployment where the public host arrives in a header rather than in `Host`, the two disagree and every state-changing request from a custom domain is rejected with 403.

Gated behind `MIDDLEWARE_HTTP_ORIGIN`, which ships `false`, so only operators who opted in are affected.

## Impact

With `MIDDLEWARE_HTTP_ORIGIN=true` and a proxy tier that forwards the public host in `Apx-Incoming-Host` / `X-Forwarded-Host` / `X-Original-Host` while rewriting `Host` to the canonical origin:

- Every unsafe-method request from a custom domain returns 403 before reaching a route.
- `/auth/sso/*` is the most visible casualty: it bypasses token CSRF (`security.rb:163`, OAuth `state` covers it), so SSO sign-in from a custom domain fails outright.

Observed in production; log line is `attack prevented by Rack::Protection::HttpOrigin` followed by a 403 on `POST /auth/sso/<provider>`.

This is not specific to one proxy vendor. It is the normal shape for Approximated ingress, AWS ALB, and most nginx configurations that preserve the origin host in a header.

## Root cause

Two host resolvers, and the security decision runs on the wrong one.

`Rack::DetectHost` (`lib/middleware/detect_host.rb:108`) is authoritative. Its own class comment states why it exists:

> While Rack's `request.host` method provides basic host detection using the `Host` header, it does not, by default, consider proxy-related headers like `X-Forwarded-Host` unless explicitly configured.

It applies `HEADER_PRECEDENCE` (`X-Forwarded-Host` > `Apx-Incoming-Host` > `X-Original-Host` > `Forwarded` > `Host`), honours forwarded headers only behind trusted infrastructure, and publishes the result. `DomainStrategy` (`middleware_stack.rb:625`) classifies it and writes `env['onetime.display_domain']` + `env['onetime.domain_strategy']`.

`HttpOrigin` (`lib/onetime/middleware/security.rb:204`) is mounted below both, at `middleware_stack.rb:662`, and consults none of it:

```ruby
def base_url(env)
  request = Rack::Request.new(env)
  ...
[...]

PR #4226, opened 2026-08-20
fix(auth): key tenant SSO on the public host, not the rewritten Host header

### LINKED ISSUE #4224: Tenant SSO resolves the wrong host behind a Host-rewriting proxy
## Problem Statement

Tenant (per-custom-domain) SSO initiation fails behind a Host-rewriting proxy. `POST /auth/sso/entra` on a custom domain answers `302 → /signin?auth_error=sso_not_configured` even though the same request classifies as `:custom` with the correct display domain. Reported on v0.26.5 against an Approximated-fronted custom domain that worked previously. Hostnames below are redacted: `tenant.example.com` = the visitor's custom domain, `origin.example.net` = the origin target the proxy rewrites `Host:` to.

## Why it fails

Production log for a single request (same pid/thread), redacted:

| stage | host seen |
|---|---|
| `[DetectHost]` | `tenant.example.com` via `HTTP_APX_INCOMING_HOST` |
| `[DomainStrategy] determined` | `tenant.example.com` → `custom` |
| `[omniauth_tenant_resolution_start]` | **`origin.example.net`** |
| result | `omniauth_tenant_no_config` (warn) → 302 `sso_not_configured` |

`apps/web/auth/config/hooks/omniauth_tenant.rb:64` keys the tenant lookup on `request.host`. That is Rack's *authority*: `X-Forwarded-Host` / `Forwarded` when present, otherwise the raw `Host:` header. Approximated rewrites `Host:` to the origin target and carries the visitor's hostname in `Apx-Incoming-Host` — a header Rack knows nothing about. So:

1. `HELPERS.resolve_custom_domain('origin.example.net')` → `nil` (no CustomDomain with that display_domain)
2. `HELPERS.canonical_domain?('origin.example.net')` → `false` (not in the canonical set)
3. falls to `handle_missing_tenant_config` → platform fallback denied (secure default) → `sso_not_configured`

`Rack::DetectHost` already resolved the correct answer into `env['onetime.display_domain']` (honoring forwarded headers **only** from trusted infrastructure, then validated by `DomainStrategy`). Every other custom-domain surface reads that env value — `HttpOriginOptions` (#4170), `Auth::SigninGate`, `Auth::RestrictTo`, `TenantSsoResolution` (#4173). The omniauth tenant hook is the one consumer still reading the raw authority.

Secondary consequence: `request.host` in Rack 3.2 honors client-supplied `X-Forwarded-Host`/`Forwarded` **ungated by proxy trust** (see #4223), so keying tenant resolution on it is also the weaker input, not just the wrong one.

## Why v0.26.4 worked (inference, not proven from the logs)

The hook's code is byte-identical between `v0.26.4` and `v0.26.5` — `host = request.host` predates both, and nothing in either version rewrites `HTTP_HOST`. What changed is what `request.host` *resolves to* at the edge:

- The 2026-08-14 uk-prod Caddy fix deleted the forwarded-host header (`header_up -X-Forwarded-Host`) because pinning it to `{http.request.host}` was masking `Apx-Incoming-Host` in `DetectHost`'s precedence. Correct for `DetectHost` — but it also removed the input Rack's `request.host` was reading, so `request.host` fell through to the rewritten `Host:`.
- That same edge change is what surfaced #4170 (`HttpOrigin` 403ing custom-domain `/auth/*` POSTs). #4170's fix — `HttpOriginOptions` reading `display_domain` — landed in v0.26.5 (`lib/onetime/middleware/http_origin_options.rb` is new in `v0.26.4..v0.26.5`). So the POST that previously died at the middleware with a 403 now reaches the hook and dies one layer deeper with this 302.

Verification for an operator: check whether the edge still sends `X-Forwarded-Host`. If it does not, `request.host` is the origin target and this path cannot resolve a tenant.

## Affected Areas

- [ ] `apps/web/auth/config/hooks/omniauth_tenant.rb` — setup hook (`host = request.host`, L64) plus the callback-side uses: restrict_to log, `resolve_custom_domain(request.host)` tenant-mismatch check, mismatch log, validated log. Fix and its logging must move together or the callback will mismatch against a session value stored from a different host source.
- [ ] `apps/web/auth/spec/unit/omniauth_tenant_helpers_spec.rb` — unit coverage for the host-selection helper
- [ ] `apps/web/auth/spec/integration/full/` — no spec drives the Approximated topology (`Host:` origin target + `Apx-Incoming-Host:` tenant domain) through the full stack; that gap is why this shipped
- [ ] `lib/middleware/detect_host.rb`, `lib/onetime/middleware/domain_strategy.rb` — the authoritative resolution the hook should consume (no change expected)
- [ ] `etc/examples/Caddyfile-example` — operator guidance; the XFH deletion is correct for `DetectHost` but silently changes `request.host` for every mounted gem
- [ ] `apps/web/auth/config/hooks/omniauth.rb:438`, `apps/web/auth/config/email/{verify_account,reset_password,email_auth}.rb` — same seam, tracked separately as #4221 (canonical-host links in mail); out of scope here
- [ ] #4223 — the general fix (boundary rewrite of `HTTP_HOST` + forwarded-carrier sanitization). This issue is the scoped, shippable half.

delano:
Fixes #4224.

## The defect

`POST /auth/sso/<provider>` on a custom domain answered `302 → /signin?auth_error=sso_not_configured` even though the very same request classified as `:custom` with the correct display domain. Confirmed in production, not merely inferred — a single box shows `[DomainStrategy] determined host=<tenant domain> strategy=custom` and, 43ms later, `[omniauth_tenant_no_config] host=<origin target>`.

Behind Approximated-style ingress the browser's hostname arrives in `Apx-Incoming-Host` and `Host:` carries the origin target. `apps/web/auth/config/hooks/omniauth_tenant.rb` keyed its `CustomDomain` lookup on `request.host` — Rack's *authority*, which knows nothing about `Apx-Incoming-Host` — so the lookup missed, `canonical_domain?` was false too, and the request fell through to the denied platform fallback.

## The fix, part one: tenant resolution

`OmniAuthTenant.public_host(request)` reads `env['onetime.display_domain']` — what `Rack::DetectHost` + `DomainStrategy` already resolved — with a `request.host` fallback for bare-Rack specs. The setup hook, `restrict_to` enforcement, the callback's tenant-mismatch check and all four auth-event logs move to it together; splitting them would let the callback compare a session value stored from a different host source.

It is also the safer key. Rack 3.2's `request.host` prefers `X-Forwarded-Host`/`Forwarded` from **any** client, ungated by proxy trust (#4223); `DetectHost` accepts a forwarded host only from trusted infrastructure and `DomainStrategy` then validates it. This is the same source every other custom-domain surface already reads — `HttpOriginOptions` (#4170), `Auth::SigninGate`, `Auth::RestrictTo`, the tenant CSP derivation (#4173) — so the runtime SSO gate and the display gates that decide whether to render the button can no longer disagree about which tenant a request belongs to.

## Second half: the redirect_uri was on the wrong host too

Resolving the tenant's credentials is only half the flow. Every absolute URL the SSO request hands an IdP comes from `OmniAuth::Strategy#full_host`, which derives from `request.url` — Rack's authority — because `OmniAuth.config.full_host` was never set anywhere in the app. `callback_url` is `full_host + callback_path` (`omniauth-entra-id-3.1.1/lib/omniauth/strategies/entra_id.rb:138` defines it verbatim; the OAuth2 family inherits the shape), and the OIDC strategies read a `client_options.redirect_uri` the tenant hook composes from it.

On a rewriting edge that authority is the origin target, so the tenant's IdP was handed `redirect_uri=https://<origin target>/auth/sso/<provider>/callback` — unregistered, so rejected; or honored, landing the visitor on a host they never authenticated from. **Without this, the first commit only turns `sso_not_configured` into a failure one hop later.**

`Auth::Config::Features::OmniAuth.install_public_host_full_host!` sets a resolver at boot. OmniAuth calls the Proc form per request with the Rack env, so one process-global assignment stays request-scoped and covers every strategy at once instead of patching the OIDC branch alone. It swaps only the hostname — scheme and port still come from the request and honor the proxy's `X-Forwarded-*` exactly as before.

**Scoping.** The resolver applies when the display domain is outside the canonical set. `DomainStrategy` pins `display_domain` to the canonical host on two paths that say nothing about where the browser is — the domains feature being off, and a detected host failing validation — so honoring those would bounce a local `localhost:3000` flow to the canonical domain mid-authentication. A host outside the canonical set can only be `DetectHost`'s validated answer for this request.

It is deliberately **not** gated on `domain_strategy == :custom`. `Chooserator` wraps its chain in a rescue, so a datastore blip — or an unparseable canonical host — classifies a real customer domain `:invalid` while `display_domain` stays correct. That is not hypothetical: the first draft gated on `:custom`, and the new integration example caught it immediately, with the tenant resolving fine and `redirect_uri` still on the origin target.

## Correction to the issue: there is no first bad version

The issue's "Why v0.26.4 worked" section is inference, and the sweep in this PR disproves it. `host = request.host` has keyed that lookup since the hook was introduced (858d6a7fa, #2730), byte-identical in every tag.

A probe sweep across published images **v0.25.11 and every v0.26.x**, fixture seeded and `DOMAINS_ENABLED=true`, reproduces the split identically in all of them: T3 (Apx-only, the production shape), T6 (`X-Original-Host`) and T8 (`Host:` onto canonical) report `SEAM_SPLIT` in every release; T9–T11 report `SPOOF_ACCEPTED` in every release. The rack 3.2.6→3.2.7 / otto 2.6→2.9 bump changed nothing here. **Do not bisect for a regression — there isn't one.** Where tenant SSO worked, the edge was sending `X-Forwarded-Host`, which Rack folds into `request.host`; the 2026-08-14 uk-prod Caddy change that deleted it (correctly, for `DetectHost`) removed that accident.

The one genuine app-side transition the sweep found is unrelated to this bug: T7 (RFC 7239 `Forwarded`) goes drift→ok at v0.26.5-rc1, where otto 2.9's `DetectHost` gained `Forwarded` support. T7 on v0.26.0 was flaky across two runs — don't lean on it.

## Edge hardening: client `X-Forwarded-Host`

`etc/examples/Caddyfile-example` forwarded the client's `X-Forwarded-Host` verbatim. `DetectHost` ranks it above `Apx-Incoming-Host` and takes the first *syntactically* valid hostname without checking whether the domain is known, so a visitor can inject a header that outranks the one the ingress set. `DomainStrategy` then rejects the unknown domain and falls back to canonical — denial, not impersonation, but it takes that tenant's custom-domain surfaces down for the request. The example now strips it; nothing is lost, since the same block pins `X-Original-Host` to Caddy's own view and preserves `Host`. This matches `scripts/host-seam/caddy-approximated-lane.caddy`, which already stripped it.

The comments also record the two things the config cannot show: that a Host-rewriting layer in front makes `request.host` the origin target for the whole app **including mounted gems** (Rodauth's `base_url`, tracked as #4221), and that pinning `X-Forwarded-Host` to `{http.request.host}` is not the fix — it masks `Apx-Incoming-Host` and produces the same failure by another route.

## Why this shipped: the missing lane

Local development is structurally blind to this class. The ops Caddyfile custom-domain block does a bare `reverse_proxy 127.0.0.1:7143`, and Caddy v2 preserves `Host`, so the public host and `request.host` are always equal locally. No spec drove the rewriting topology either.

`scripts/host-seam/` closes both:

| file | purpose |
|---|---|
| `topology-probe.sh` | 11 proxy topologies against a running instance; oracle is the `O-Domain-Strategy`/`O-Display-Domain` headers (stable since v0.25.0) vs. the SSO POST's `Location`. Verdict `SEAM_SPLIT` = strategy says `custom` while SSO says `sso_not_configured` on one request |
| `caddy-approximated-lane.caddy` | turns `local-secrets4/5` into a Host-rewriting lane so the topology is reproducible locally |
| `seed-tenant.rb` | fixture domain + `SsoConfig` |
| `release-sweep.sh` | runs the matrix across ghcr image tags |
| `README.md` | verdict taxonomy, setup, traps |

Two traps the harness now handles, both of which bit during the sweep: containers **must** get `DOMAINS_ENABLED=true` (the feature defaults off, `DomainStrategy` pins canonical, and the sweep reports "no transitions" vacuously — there is now a direct-request control reporting `UNTESTABLE(strategy_control)`), and an unseeded fixture domain used to masquerade as `SEAM_SPLIT` (now `FIXTURE_MISSING`). The probe needs `TRUSTED_PROXY_ENABLED=true` or `DetectHost` discards forwarded headers and every topology collapses to canonical for unrelated reasons.

#4223 names this matrix as the gate for its flag flip.

## Testing

- `apps/web/auth/spec/unit/omniauth_tenant_helpers_spec.rb` — `.public_host` prefers the display domain, falls back on absent and on blank.
- `apps/web/auth/spec/unit/omniauth_full_host_spec.rb` — the resolver across both proxy topologies, the canonical-set carve-out (including a split deployment's second canonical host and the local-`localhost:3000` case), the `:invalid` degradation, port/scheme composition, and that a Proc is installed rather than a String. **16 examples.**
- Whole auth unit suite: **479 examples, 0 failures.**
- `apps/web/auth/spec/integration/full/tenant_sso_proxy_host_spec.rb` — full stack, three examples: the rewriting topology resolves the tenant, the `redirect_uri` in the authorize URL names the tenant domain and not the origin target, and the preserved-`Host` topology still resolves. Needs `ORGS_SSO_ENABLED=true` and `AUTH_DATABASE_URL` (PostgreSQL).
- SSO-adjacent integration specs (tenant resolution, missing-config fallback, callback validation, CSRF, route registration, restrict_to): **85 examples, 0 failures**, 4 pre-existing lane pendings.
- `rubocop` clean; `caddy validate -c etc/examples/Caddyfile-example` → Valid configuration; `caddy fmt` reports no changes.

## Out of scope

- #4221 — canonical-host links in Rodauth mail. Same seam, different consumer.
- #4223 — the general fix (boundary rewrite of `HTTP_HOST` + forwarded-carrier sanitization). This PR is the scoped, shippable half.


delano:
Ready after checks

PR #4232, opened 2026-08-20
*fix(ci): actually deliver frontend sourcemaps to Sentry

delano:
## What

Frontend stack traces in Sentry have never symbolicated — on any release built here. Two independent causes, both of which reported green on every build.

Split out of the telemetry-boundary work so a release-engineering review and a privacy review don't gate each other. This PR touches no application code.

## Cause 1 — the upload has always shipped zero files

The workflow ran `sentry-cli sourcemaps upload ./public/web/dist` on the runner. But the frontend is compiled **inside the image** (the Dockerfile `build` stage runs `pnpm run build`), and `public/web/dist` is gitignored — so that directory has never existed at that point in the job. `sentry-cli` treats an empty directory as a successful upload of zero artifacts.

Verified against the instance: the `frontend` project holds **zero artifact bundles for every release in its history**, including `b7aaea0`. Same for `frontend-staging`, `frontend-tundra` and `backend`.

Assets are now extracted from the image that was just pushed, rather than rebuilt on the runner. Vite emits content-hashed filenames, so a second build would produce artifacts named for files no browser ever requests — green, and just as useless.

## Cause 2 — the dist join key never matched

Sentry resolves an artifact bundle only when **release AND dist both match**. This upload has always passed `--dist=frontend`, while every sampled event carried `dist: null`. Even a working upload could not have symbolicated a frame.

**This PR fixes only CI's half.** The frontend half — `dist: 'frontend'` in `src/plugins/core/enableDiagnostics.ts` — lands with the companion telemetry-boundary PR. Until then the preflight reports the mismatch as an explicit dist-tag warning on every run instead of staying silent.

Neither merge order breaks anything that currently works (symbolication is already broken), but **both are required before a frame symbolicates**. The preflight reads the TypeScript literal and compares it to `SENTRY_DIST` in both directions, so future drift fails loudly rather than silently.

## Making delivery observable

`continue-on-error` is correct on these steps — telemetry plumbing must never fail a production image build. But it makes each step's exit code meaningless as a signal, which is why this went unnoticed for so long: a green check said nothing about whether anything arrived.

Every Sentry step now reports through `scripts/ci/sentry-status.sh`, which emits a GitHub annotation and a row in a job-summary table that outlives the log. States are deliberately split:

| State | Meaning | Signal |
|---|---|---|
| `SKIPPED` | no `SENTRY_AUTH_TOKEN` — normal on forks and unconfigured clones | log line only, never cries wolf |
| `BLOCKED` | credentials **are** configured but a precondition failed | `::warning::` |
| `FAILED` | credentials **are** configured and the command errored | `::error::` |

New scripts:

- **`sentry-sourcemap-preflight.sh`** — pre-upload assertions: dist tree present, `.map` files present, admin bundle present, release matches `__SENTRY_RELEASE__`, project/dist agree with what a frontend event reports.
- **`sentry-verify-artifacts.sh`** — asks the *server* after the upload. Input checks and exit codes were both satisfiable while the project held zero artifacts. Only a post-upload query distinguishes a real delivery from a successful-looking one. **This is the only check that would have caught the original defect.**
- **`extract-frontend-dist.sh`** — pulls the built assets out of the pushed image.
- **`sentry-status.sh`** — the annotation/summary reporter.

## Also removed

`SENTRY_DIST` from `.env.reference`. Nothing in `lib/`, `apps/` or `src/` ever read it and the preflight never checked it — it advertised a runtime knob for a build-time literal, while offering no way to actually set it.

## New optional secret

`SENTRY_FRONTEND_PROJECT` (default `frontend`) names the project the sourcemap upload targets. It must be one of `SENTRY_PROJECTS`, or the release and its artifact bundles land in different projects and symbolication silently fails again.

## Verification

- `actionlint`: **zero new findings** vs `main` (10 before, 10 after — the three moved `SC2086` hits are the same intentional `$PROJECT_FLAGS` word-splitting, relocated into the new loop).
- `shellcheck -S warning` on all four scripts: **clean**.
- Workflow YAML parses.
- `bash -n` clean on all four scripts.

The end-to-end path can only be confirmed by a real push build, since it depends on the pushed image and live Sentry credentials. The verification step is what will report it — and, unlike before, it will say so loudly if it fails.


.github/workflows/build-and-publish-oci-images.yml
```diff
@@ -288,28 +384,137 @@ jobs:
...
+            rc=$?
+            printf '%s\n' "$out"
+            if [ "$rc" -ne 0 ]; then
+              bash scripts/ci/sentry-status.sh record release failed \
+                "releases ${phase} exited ${rc} for release ${SENTRY_RELEASE} (projects: ${SENTRY_PROJECTS}): $(printf '%s' "$out" | tail -n 2)"
+              exit 0
+            fi
+          done
+          bash scripts/ci/sentry-status.sh record release ok \
+            "release ${SENTRY_RELEASE} created, commits associated and finalized for: ${SENTRY_PROJECTS}"
```
delano: Secret values are interpolated into publicly-visible output. `SENTRY_PROJECTS` is a repository *secret*, and this detail (plus the `blocked`/`failed` details above) goes into a `::warning::`/`::notice::` annotation **and** into the job-summary table on a public repo.

Both outcomes are bad:
- exact match → the runner masks it, so the row renders `finalized for: ***` — unreadable, defeating the summary this PR exists to produce;
- sentry-cli stderr tails (`$(printf %s "$out" | tail -n 2)`) name the org and each project slug *individually*; a substring of a space-separated secret is **not** masked, so the private Sentry org/project identifiers publish on the public run page.

Prefer a count or a fixed label (`"release created for all configured projects"`), and drop raw CLI tails from details that reach the summary.

scripts/ci/sentry-sourcemap-preflight.sh
```diff
@@ -0,0 +1,232 @@
...
+fi
+
+if [ -n "$UPLOAD_DIST" ] && [ -z "$frontend_dist" ]; then
+  warn dist-tag warn \
+    "upload tags bundles with --dist='${UPLOAD_DIST}' but the frontend sets no dist (neither src/plugins/core/enableDiagnostics.ts nor the backend diagnostics config emits one), so events carry dist=undefined and will never resolve these bundles. Drop --dist from the upload, or set a matching dist in Sentry.init."
+elif [ -z "$UPLOAD_DIST" ] && [ -n "$frontend_dist" ]; then
+  warn dist-tag warn \
+    "frontend events carry dist='${frontend_dist}' but the upload passes no --dist — bundles are filed dist-less and will not resolve. Pass --dist='${frontend_dist}'."
+elif [ -n "$UPLOAD_DIST" ] && [ "$frontend_dist" != "$UPLOAD_DIST" ] && \
+     [ "$frontend_dist" != "<non-literal>" ] && [ "$frontend_dist" != "<from backend config>" ]; then
```
delano: The sentinel exemptions turn an unknown dist into an affirmative false OK, not just a silenced warning.

When `frontend_dist` is `<non-literal>` (someone writes `dist: someVar`) or `<from backend config>` (any future Ruby line in `config_serializer.rb` matching the loose `['"]?dist['"]?[[:space:]]*[:=]` grep), this branch is skipped and execution falls through to the final `elif [ -n "$UPLOAD_DIST" ]` → `report dist-tag ok "upload --dist='frontend' matches the frontend dist"`. The preflight then asserts the join key matches when it has no idea whether it does.

Altitude: the release half is checked against *ground truth* (grep the built chunks for the inlined `__SENTRY_RELEASE__`). The dist half text-scrapes source files. Define `__SENTRY_DIST__` in `vite.config.ts` next to `__SENTRY_RELEASE__` and grep the built assets for it — same mechanism, no scrape to rot.

.github/workflows/build-and-publish-oci-images.yml
```diff
@@ -288,28 +384,137 @@ jobs:
...
+            bash scripts/ci/sentry-status.sh record sourcemaps failed \
+              "sentry-cli exited ${rc} uploading ${SENTRY_DIST_DIR} for release ${SENTRY_RELEASE}: $(printf '%s' "$out" | tail -n 2)"
+            exit 0
+          fi
+
+          # A zero-exit upload that moved no files is the failure this whole
+          # step exists to surface. The preflight already proved .map files are
+          # on disk, so sentry-cli reporting nothing here means it rejected or
+          # ignored them.
+          if printf '%s' "$out" | grep -qiE 'nothing to upload|no files|0 files'; then
```
delano: The `0 files` alternative is unanchored, so it substring-matches a healthy upload.

`sentry-cli sourcemaps upload` prints count lines like `Found 17 files` / `Bundled 17 files for upload`. Any run whose count ends in zero — `Bundled 10 files for upload`, `Found 20 files` — contains `0 files` and gets recorded as `sourcemaps warn: "reported no uploaded files"`. That fires on a fully successful delivery and trains reviewers to ignore exactly the warnings this branch adds.

Anchor it, e.g. `(^|[^0-9])0 files`, or drop the output scrape entirely and lean on `sentry-verify-artifacts.sh`, which asks the server two steps later.

scripts/ci/sentry-status.sh
```diff
@@ -0,0 +1,171 @@
...
+set -uo pipefail
+
+STATUS_FILE="${SENTRY_STATUS_FILE:-${RUNNER_TEMP:-/tmp}/sentry-delivery-status.tsv}"
+
+# Annotations are parsed out of the log stream, so any text interpolated into
+# one must not be able to forge a workflow command or break out onto a new
+# line. Detail strings here can carry sentry-cli output, which is external
+# input. Same reasoning as validate-register.yml:135-143.
+sanitize() {
+  printf '%s' "$*" | tr '\n\r\t' '   ' | sed 's/::/;;/g' | cut -c1-700
```
delano: The 700-char truncation can defeat GitHub secret masking. Masking is exact-value: if an interpolated detail embeds a secret (a `SENTRY_URL` inside a sentry-cli request-failed message, a project slug) that *starts* before char 700 and ends after it, `cut` emits a prefix that no longer matches the registered secret, so the runner cannot redact it and the partial secret publishes in the annotation and in the job summary of this public repo.

Truncate to a boundary that cannot split a value mid-token, or (better) keep externally-sourced CLI/server output out of `detail` entirely and log it separately to the step log.

scripts/ci/sentry-verify-artifacts.sh
```diff
@@ -0,0 +1,140 @@
...
+set -uo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+STATUS="${REPO_ROOT}/scripts/ci/sentry-status.sh"
+
+ORG="${SENTRY_ORG:-}"
+PROJECT="${SENTRY_PROJECT:-frontend}"
+RELEASE="${SENTRY_RELEASE:-}"
+BASE_URL="${SENTRY_URL:-https://sentry.io}"
+BASE_URL="${BASE_URL%/}"
```
delano, line 90: Stripping the trailing slash can defeat masking of `SENTRY_URL`. GitHub masks the registered secret value (plus known encodings), not near-matches. If the operator stored `SENTRY_URL` as `https://sentry.internal.example/`, `${BASE_URL%/}` is a *proper prefix* and is printed **unmasked** into a public `::warning::`/`::notice::` and a summary row at lines 128 and 139 — disclosing the self-hosted Sentry hostname.

Either way one failure fires: stored without the slash, the same rows render `***` and say nothing. Do not interpolate `$BASE_URL` into reported details at all — the org/project/release identify the target well enough.

.github/workflows/build-and-publish-oci-images.yml
```diff
@@ -288,28 +384,137 @@ jobs:
...
+      # Pre-upload assertions. `sentry-cli sourcemaps upload` happily "succeeds"
+      # against an empty directory, so without this the only way to learn that
+      # nothing shipped is to go looking in Sentry. The script classifies and
+      # reports; it never fails, it just tells the upload step whether to run.
+      # Gated on the extraction: when that step reports anything other than
+      # `extracted` it has already recorded the loud, specific reason, and
+      # re-reporting "dist tree missing" here would only bury it.
+      - name: Sentry sourcemap preflight
+        id: sentry_preflight
+        if: github.event_name != 'pull_request' && steps.extract_assets.outputs.status == 'extracted'
```
delano, line 480: Gating the whole chain on a step output makes a crashed extract step *silent* — the exact state this PR exists to eliminate.

If `extract-frontend-dist.sh` dies before `emit_status` runs (runner OOM mid `docker pull`, cancellation, step timeout, a future `set -u` error), `continue-on-error` swallows the exit, `steps.extract_assets.outputs.status` is empty, and preflight (424), upload (439) and verify (509) are all skipped. No row is ever recorded for sourcemaps, so `render` prints the release rows and then `> All Sentry delivery checks passed or were cleanly skipped.` The only trace is a tolerated-failure step icon — which this PR argues nobody reads.

Add a fallback: have the summary step assert a `frontend-assets` **and** a `sourcemaps` row exist for non-PR events, and record BLOCKED when either is missing.

scripts/ci/extract-frontend-dist.sh
```diff
@@ -0,0 +1,183 @@
...
+if [ -z "$IMAGE_REF" ]; then
+  report frontend-assets blocked \
+    "could not resolve a tag for bake target '${BAKE_TARGET}' in group '${BAKE_GROUP}' from ${BAKE_FILE} --print — the tag set is empty, which means IMAGE_TAG and EXTRA_TAGS were both empty for this event (see the tags() function in ${BAKE_FILE} and the 'Compute version and tags' step). No image to extract from, so no sourcemaps will ship."
+  emit_status blocked
+  exit 0
+fi
+
+# Prefer the digest the push actually produced. Tags are mutable; a digest names
+# the exact image whose bundles this release's events will come from.
+if [ -n "${BAKE_METADATA:-}" ]; then
```
delano, line 115: The digest fallback is silent. If `BAKE_METADATA` carries no `containerimage.digest` (metadata-shape change, provenance disabled), `IMAGE_REF` quietly stays the **mutable tag** and nothing records that the pin was lost — the only trace is the `Extracting … from …` log line.

Concretely: two pushes to the same branch both tag `edge`; run 1 pulls `edge` after run 2 overwrote it and uploads run 2's bundles under run 1's release SHA. The preflight release-parity check would notice (run 1's SHA is not in run 2's chunks) but only as a non-blocking `warn`, and this step still reports `ok`.

Record a `warn` when the digest is absent, so an unpinned pull is visible rather than assumed benign.

scripts/ci/sentry-status.sh
```diff
@@ -0,0 +1,171 @@
...
+      case "$state" in
+        ok) label="OK" ;;
+        skipped) label="SKIPPED" ;;
+        warn) label="WARNING" ;;
+        blocked) label="BLOCKED" ;;
+        failed) label="FAILED" ;;
+        *) label="UNKNOWN" ;;
+      esac
+      printf '| %s | **%s** | %s |\n' \
+        "${component//|/\\|}" "$label" "${detail//|/\\|}"
```
delano: Only `|` is escaped, but job summaries render GFM — links, images (Camo-proxied) and an HTML subset. `detail` strings carry registry, docker and sentry-cli/server output, i.e. text controlled by a remote the workflow merely trusts. A `docker pull` failure body containing a markdown link or `` injects a phishing link or a viewer-tracking beacon into this public repo's run page.

Wrap details in backticks with backtick-escaping, or strip `<>[]()` in `sanitize()` alongside the existing `::` rewrite.

.github/workflows/build-and-publish-oci-images.yml
```diff
@@ -288,28 +384,137 @@ jobs:
...
+              "sentry-cli exited 0 but reported no uploaded files for release ${SENTRY_RELEASE} — bundles present on disk were not accepted: $(printf '%s' "$out" | tail -n 2)"
             exit 0
           fi
-          npx @sentry/cli sourcemaps upload ./public/web/dist \
-            --release="$SENTRY_RELEASE" \
-            --project=frontend \
-            --dist=frontend \
-            --url-prefix='~/dist'
+
+          MAP_COUNT=$(find "$SENTRY_DIST_DIR" -type f -name '*.map' | wc -l | tr -d ' ')
```
delano, line 557: The success message counts `.map` files **on disk**, not what sentry-cli reported shipping. If the CLI skipped or rejected a subset while still exiting 0 (dedup against an existing bundle, per-file errors), the `::notice::` claims `uploaded N sourcemap(s)` for the disk count — a small re-run of the "looks delivered but was not" gap this PR closes. Parse the count out of `$out`, or word it as `N sourcemap(s) on disk submitted`.

scripts/ci/sentry-verify-artifacts.sh
```diff
@@ -0,0 +1,140 @@
...
+  exit 0
+fi
+
+# Returns the element count of a JSON array endpoint, or -1 when the endpoint
+# could not be read (HTTP error, or a body that is not a JSON array — e.g. an
+# error object, which self-hosted returns for an unknown release).
+count_endpoint() {
+  local url="$1" body http_code
+  body="$(curl -sS --max-time 30 -w $'\n%{http_code}' \
+    -H "Authorization: Bearer ${SENTRY_AUTH_TOKEN}" \
```
delano: The auth token is passed as a command-line argument, so it sits in `/proc/<pid>/cmdline` for the duration of each of up to nine `curl` invocations, readable by any same-user process on the runner. Marginal on an ephemeral GitHub-hosted runner where the token is already in the step environment, but free to fix: `curl --config -` on stdin, or `-H @headerfile`, keeps it off argv.

delano:
Good to go after checks

PR #4233, opened 2026-08-20
Use public hosts for auth links and secure proxy diagnostics

### LINKED ISSUE #4221: Rodauth transactional emails mint canonical-host links for custom-domain users

delano:
## Summary
- Resolve transactional email links, WebAuthn origins, and OmniAuth redirect URLs from trusted public-host metadata instead of proxy-rewritten request authority, preventing wrong-domain links for custom-domain deployments.
- Add a Colonel proxy-header diagnostic that returns a fixed Caddy/Rack field set and is protected by `network=admin`; the route returns 404 unless explicit host and CIDR admin gates are configured and admit the request.
- Memoize the mounted Rack stack and correct merged integration-hook ordering so the custom-domain test axis remains stable while these behaviors are exercised.

## Review guide
- Start with `apps/web/auth/lib/public_host.rb` and the Rodauth/OmniAuth consumers; verify trusted-host resolution, canonical-host filtering, and fallback behavior.
- Review `lib/onetime/application/network_requirements.rb`, `lib/onetime/middleware/admin_network_isolation.rb`, and the Colonel route/auth strategy for fail-closed enforcement and the restricted diagnostic payload.
- Check the added auth, network-requirement, proxy-diagnostic, and public-host coverage plus the operator documentation for deployment requirements.

## Validation
- Pre-commit, pre-push, and CI checks will validate this change.

Closes #4221


delano:
Ready after checks

PR #4240, opened 2026-08-21
Serialize Colonel subscription_period_end as a string, and make the failure that hid it diagnosable

delano:
Missed the v0.26.6 cutoff — no PR was ever opened, so none of this landed. Targeting v0.26.7.

Verified against `main`: `git cherry -v main HEAD` marks all 22 commits `+`, and of the 72 files touched, zero are identical in main. Test-merge is clean (no conflicts); the stale pre-v0.26.6 changelog fragments still in the branch tree drop out correctly, leaving exactly one new fragment.

## What started this

The Colonel organization-detail page broke on a schema-validation failure, and the Sentry event said only that validation failed. Finding it took a manual bisect of the payload. Two things came out of that:

**The fix.** Five serializers emitted `subscription_period_end` raw, so a numeric datastore value hit a `z.string().nullable()` and blew up the parse. All now serialize with `&.to_s`, and the schema accepts a numeric legacy value while a mixed-version deployment or cached response is still possible. The fifth — `Org::Reconcile`'s summary hash — was missed on the first pass and is fixed here too.

A sixth site turned up on a different surface: `build_subscription_data`, behind the billing overview endpoint. Because the field is untyped on the model and its writers are inconsistent, that endpoint emitted a JSON *number* from some paths and a string from others — one wire key, two types. Fixed the same way, with a regression spec that assigns an Integer and asserts the string comes back (it returns `1772940425` without the fix).

The frontend's `BillingOverviewResponse` declared `period_end: number` and was equally wrong; it now declares `string | null`. Both consumers parse with `parseDateValue` rather than relying on `'1772940425' * 1000` coercing, which yields an Invalid Date for anything non-numeric. A zero epoch now reads as an unset field instead of rendering 1 Jan 1970 — the truthiness check it replaces excluded 0 by accident, and a plain parse would not have.

**The reason it was hard to find.** A `ZodError` reaching Sentry was rendered in a way that discarded the field path. The same event now reports `record.subscription_period_end`, `invalid_type`, expected `string`, received `number` — path, issue code, and the types on both sides, with no value from the payload.

## The boundary

Making that diagnostic safe to send is the bulk of the diff. Five parts, each enforced structurally rather than by convention:

- **Value-free schema diagnostics** — one projection for all `ZodError` rendering. Zod's issue text, `issue.input`, `unrecognized_keys` key names and `issue.params` are refused; an audit found a `params` carrying a live Stripe secret key. Rows are flat by contract, because Sentry's `normalizeDepth` silently collapses a nested row to `[Object]`.
- **Pseudonymous actor references** — `Onetime::Utils::DiagnosticsRef` derives an opaque, keyed, one-way `user.id`. Deliberately not the federation email hash, which is both a queryable datastore index and a field in Stripe customer metadata.
- **Parameterized route tags** — `/api/colonel/organizations/:org_id`; the resolved form is never retained, even in memory.
- **Pseudonymous resource references** — because the route is parameterized, "one organization is broken" reads identically to "every organization is broken". `organization_ref` (opaque, keyed, 16-hex, same derivation) rides the Colonel organization-detail record and is attached as a tag on a failing parse, answering that by cardinality alone. Read off the raw payload, since a failed parse leaves no parsed record; emitted only for schemas in an exact-match `(schema, path)` allowlist — today just this one response.
- **Metadata-only breadcrumbs** — per-category key allowlists with declared primitive types, so a key that does not exist yet is refused by default. Bodies, headers and console argument lists are dropped.

Free text reaching the surface is scrubbed by shape (emails, verifiable identifiers, Onetime extids, prefixed credentials, UUIDs, IPs), each net emitting its own sentinel so a reader knows what was removed. The nets stay narrow where it matters for diagnosis — trace ids, commit hashes, version strings, Ruby constant paths and schema field names survive intact.

Diagnostics here means making a defect reproducible by the people who have to fix it. Nothing in this subsystem measures usage or profiles behaviour.

## Residency

References are scoped to one data-residency jurisdiction, not one federation. Regional instances share `FEDERATION_SECRET` by design and report into one Sentry instance, so a region-independent reference would emit the identical `user.id` from EU and US — a join key proving one data subject is present in both. The scope is mixed into the derivation unconditionally, no opt-out. An undeclared residency *withdraws* the shared key rather than widening the reference: `FEDERATION_SECRET` is refused and the derivation falls back to per-deployment `ACCOUNT_ID_SECRET`, with the emitted `actor_scope` label narrowing to match.

## Sourcemaps

The fragment notes frontend events carried `dist: null` while CI uploaded with `--dist=frontend`, so no frame could ever have symbolicated. This branch is the frontend half; CI's half landed on main in #4232. Both are required, in either merge order — with this merged, both halves are present.

## Operator impact

**None required.** With neither `FEDERATION_SECRET` nor `ACCOUNT_ID_SECRET` configured, no reference is emitted (`organization_ref` is `null` — the default in dev and test) and nothing else changes. Diagnostics stay off entirely unless `DIAGNOSTICS_ENABLED` is set.

New and optional: `DIAGNOSTICS_REF_REGION`, for the case where several instances share `FEDERATION_SECRET` and serve different jurisdictions without using the regions feature. Otherwise `JURISDICTION` already supplies it. Changing it re-keys every reference on that install.

## Naming

The later commits rename `telemetry` → `diagnostics` throughout (`TelemetryRef` → `DiagnosticsRef`, `TELEMETRY_REF_REGION` → `DIAGNOSTICS_REF_REGION`) and consolidate `src/utils/telemetry` plus the plugin scrubbers under `src/utils/diagnostics`. "Telemetry" reads as analytics/metrics, which this is not. The unrelated `access_telemetry.rb` is untouched.

## AI assistance

Claude built the boundary and its acceptance suite. Six adversarial passes were run, each executing attacks against the tree rather than reading it. The recurring defect they caught was comments asserting privacy guarantees the code did not deliver — including a correlation-radius claim in the TypeScript contract that the Ruby derivation contradicted, and a residency guard that covered the raising fault but not the falsy one.


delano:
Superseded. Split into three PRs so each reviews on its own subject: https://github.com/onetimesecret/onetimesecret/pull/4242 (subscription_period_end + diagnostics boundary — same 24 commits rebased onto current main), https://github.com/onetimesecret/onetimesecret/pull/4243 (Colonel proxy-header diagnostic), https://github.com/onetimesecret/onetimesecret/pull/4244 (Caddyfile static-asset example).

PR #4256, opened 2026-08-23
Standardize text capitalization and update security policy

delano:
> Updating this description: the original called the locale changes "purely cosmetic". Most are, but four of them correct entitlement statements that were factually wrong, and one removes a tier name that does not exist. Those deserve a reviewer's attention rather than a skim.

## Summary

Two things: English UI copy is brought in line with the project's own voice guide, and `SECURITY.md` is restructured. Along the way, several in-product plan-entitlement strings turned out to be inaccurate and are corrected here.

### Entitlement copy — please review these closely

These were wrong in the shipped product, checked against `pricing/compare-plans.md` in the docs repo:

| Before | After |
|---|---|
| "Team management requires **Identity Plus** or higher" | "…requires **Team Plus** or higher" |
| "**Custom domains** require Identity Plus or higher" | "**Additional** custom domains require Identity Plus or higher" — Free includes one |
| "**API access** requires Identity Plus or higher" | "**Full** API access requires a signed-in account" — Free has API access |
| "Audit logs require **Multi-Team plan** or higher" | "…require **Team Plus** or higher" — "Multi-Team plan" is not a tier that exists |

### Capitalization

`translations/universal/voice-and-tone.md` prescribes sentence case for buttons ("Save changes", "Generate password"), but the product shipped roughly half title case — including the guide's own counter-example. Converted button, action, form-label, and CTA strings, plus the nav labels that render beside them, so a single screen is not internally inconsistent.

Deliberately **not** converted: `web.TITLES.*`, email subject lines, and proper-noun feature names. Where a button and a page title previously shared one string, they now diverge — that is intended.

### `SECURITY.md`

Supported Versions was prose with no version numbers, so a reader could not tell whether their release was covered; it is now a table. Normalized mixed time units (5 business days / 14 business days / every 7 days). Dropped the note asking reporters to confirm their question was not already answered — friction belongs least on that channel. Every existing commitment is preserved: acknowledgment and assessment windows, the update cadence, confidentiality until a fix ships, encrypted email, `security@onetimesecret.com`, the subject-line convention, and the no-paid-bounty stance.

### Dead copy and hygiene

- `error-pages.json` carried four overlapping 404 strings, two byte-identical. Only the one rendered by `ErrorNotFound.vue` is referenced anywhere; removed the other three, mirrored across every locale so key sets stay aligned.
- Stripped a leading space that rendered in the password-generator description. Also stripped the trailing space on `web.COMMON.sent_to` — safe, because `ShowReceipt.vue:151` renders it inside its own `span` element carrying an `mr-2` margin class, so the gap comes from CSS rather than from the string.
- Normalized ellipsis punctuation; rewrote the homepage CTA subtitle, which echoed "share with confidence" from the title directly above it.
- Hardcoded English literals in `WorkspaceLayout.vue` / `ManagementLayout.vue` re-cased. Worth flagging separately: these sit outside the locale system entirely, so they render in English in all 29 locales. Pre-existing gap, not introduced here.

## Test Plan

`T2 · TypeScript Unit Tests` passes, covering the updated assertions in `MastHead.spec.ts`, `SplitButton.spec.ts`, `SplitButton.a11y.spec.ts`, and `DomainSignup.spec.ts`. `T1 · i18n Validation` passes, confirming `content_hash` freshness after `locales:hashes:apply` (the tool's `--max-rewrites` guard trips at 99 rewrites — verified 1:1 against 100 changed text values, with all 3,976 en hashes independently recomputed, so that is intended source churn).

**Known red check:** `bin/visual` — 12 baselines changed, all of them the intentional copy edits above, on exactly the surfaces whose strings changed. Regenerating them needs `bin/visual --update` in the pinned podman container, which I could not run; rendering outside it produces phantom diffs from font-stack drift. Needs a maintainer. Details in [this comment](https://github.com/onetimesecret/onetimesecret/pull/4256#issuecomment-5383699362).


delano:
### Visual regression: 12 diffs are expected, but I can't regenerate the baselines here

All 12 changed shots are caused by the intentional sentence-case copy changes in this PR, not by a layout or styling regression. The failing surfaces line up exactly with the strings that changed:

| Surface | Strings on it that changed |
|---|---|
| `home--default--*` | `Generate Password` → `Generate password`, `Create Account` → `Create account`, `Sign In` → `Sign in`, `Get Started` → `Get started` |
| `incoming--form/success` | `Create Incoming` → `Create incoming`, `Secret Information` → `Secret information` |
| `receipt--fresh/viewed/burned` | `Sent to: ` → `Sent to:`, `Show More/Less` → `Show more/less` |

**On the one change that looked risky:** `web.COMMON.sent_to` previously carried a trailing space. I checked the caller before stripping it — `src/apps/secret/reveal/ShowReceipt.vue:151` renders it in its own `<span class="mr-2">`, so the gap comes from the CSS margin, not the string. No `Sent to:alice@example.com` regression.

**Blocker:** the baselines need `bin/visual --update`, which requires podman and the pinned Playwright container. This environment has docker but not podman, and `bin/visual` is explicit that rendering outside the pinned container produces phantom diffs from font-stack drift — so generating them here risks committing subtly wrong `*-linux.png` baselines, which is worse than leaving the check red. `.github/workflows/visual.yml` only compares; it has no baseline-refresh input I could dispatch.

So this needs a maintainer to run `bin/visual --update` locally and commit the 12 refreshed shots. Happy to take another pass at anything else on the PR in the meantime.

---

PR #4259, opened 2026-08-23
Clarify and lock in LOGO_SHOW_NAME precedence on custom domains

### LINKED ISSUE #4241: LOGO_SHOW_NAME override is silently defeated by domain_strategy guard
## Problem Statement

`LOGO_SHOW_NAME=true` does not reliably force the product-name wordmark to appear next to the masthead logo. On some local/dev instances the wordmark never renders — authenticated or not — regardless of the env var setting.

## Early Diagnosis

`MastHead.vue`'s `getShowSiteName()` (src/shared/components/layout/MastHead.vue:126-134) checks `identityStore.showPlatformIdentity` *before* consulting `LOGO_SHOW_NAME` (via `headerConfig.logo.show_name`). If `showPlatformIdentity` is `false`, the function returns `false` immediately and `LOGO_SHOW_NAME` is never read.

`showPlatformIdentity` is computed as `!isCustom && !logoUri` (identityStore.ts:238), and `isCustom` is `domainStrategy === 'custom'` (identityStore.ts:197), sourced from the bootstrap payload's `domain_strategy`.

`domain_strategy` is resolved server-side in `Onetime::Middleware::DomainStrategy#call` (lib/onetime/middleware/domain_strategy.rb:214-242). When `DOMAINS_ENABLED` is true, it runs `Chooserator.choose_strategy` against `CANONICAL_DOMAIN` — a request host that doesn't match resolves to `:custom` (or `:invalid`), which unconditionally suppresses the wordmark.

This is likely correct behavior for genuine third-party custom domains (the A3 brand-leak guard), but it means:
1. `LOGO_SHOW_NAME` cannot override the suppression even when an operator explicitly wants the wordmark shown — the env var comment in `.env.reference:1562-1570` doesn't mention this precedence, which is confusing/undocumented at minimum.
2. On dev/local setups where `DOMAINS_ENABLED=true` but the request host doesn't match `CANONICAL_DOMAIN`, the wordmark silently vanishes with no indication why, which reads as a bug from the operator's perspective.

Needs a decision: is `LOGO_SHOW_NAME` intended to be able to override the platform-identity guard, or is the current precedence intentional and only the docs need clarifying? Either way the current behavior and the documented behavior are out of sync.

## Affected Areas

- [ ] `src/shared/components/layout/MastHead.vue` — `getShowSiteName()` priority ladder (lines 114-134)
- [ ] `src/shared/stores/identityStore.ts` — `showPlatformIdentity` computed (line 238), `isCustom` (line 197)
- [ ] `lib/onetime/middleware/domain_strategy.rb` — `domain_strategy` classification (lines 214-242)
- [ ] `.env.reference` — `LOGO_SHOW_NAME` documentation (lines 1562-1570) doesn't mention the domain_strategy precedence

delano:
## Summary

[#4241](https://github.com/onetimesecret/onetimesecret/issues/4241)

The issue asked for a decision: should `LOGO_SHOW_NAME` be able to override the `showPlatformIdentity` guard, or is the current precedence intentional and only the docs need clarifying?

Tracing `getShowSiteName()` (`MastHead.vue`) → `showPlatformIdentity` (`identityStore.ts`) → `isCustom`, the guard already tests strictly `domainStrategy === 'custom'`. A request host that fails to resolve to any known canonical or tenant domain classifies `'invalid'` (e.g. a dev `CANONICAL_DOMAIN` mismatch), not `'custom'` — so it never trips the guard, and `LOGO_SHOW_NAME` already reaches rung 3 and applies normally there. No functional bug exists in that path today.

For a **genuine** custom (tenant) domain, however, the suppression is absolute and deliberate — an extensive, pre-existing "A3 leak guard" test suite (`MastHead.customDomain.spec.ts`) locks in that the operator's own product name must never render on a tenant's white-labeled domain, logo or not. `LOGO_SHOW_NAME` is an install-wide operator setting; letting it cross that guard would leak the operator's identity onto every customer's custom domain. That precedence should not change.

So the actual bug was that this precedence — and its precise scoping to `'custom'` rather than any non-canonical classification — was **undocumented and untested**, which is exactly how it read as a silently "defeated" override. This PR:

- Documents the precedence in `.env.reference` (the operator-facing `LOGO_SHOW_NAME` reference), `docs/architecture/branding.md`, and the `getShowSiteName()` / `showPlatformIdentity` code comments.
- Adds regression coverage locking in both halves that were previously unverified:
  - `LOGO_SHOW_NAME=true` stays defeated on a genuine custom domain with **no** uploaded tenant logo (the domain-with-logo case was already covered; this combination wasn't).
  - `LOGO_SHOW_NAME=true` still works on an unresolved/misconfigured request host (`domain_strategy: 'invalid'`), at both the store level (`identityStore.showPlatformIdentity`) and the component level (`MastHead`'s rendered wordmark).

No runtime logic changed — `lib/onetime/middleware/domain_strategy.rb` classification is untouched, since it was already correct and is covered by its own pinned contract spec.

## Related Issues

Closes #4241

## Test Plan

- [x] `pnpm test:base run src/tests/stores/identityStore.spec.ts src/tests/shared/components/layout/MastHead.spec.ts src/tests/shared/components/layout/MastHead.customDomain.spec.ts` — 132 passed (129 pre-existing + 3 new)
- [x] `eslint` on all touched `.ts`/`.vue` files — 0 errors (pre-existing style warnings only, unrelated to this change)

---
_Generated by [Claude Code](https://claude.ai/code/session_015TKwDqYCqV2QXLBmumiMKk)_

### DIFF
--- a/.env.reference
+++ b/.env.reference
@@ -1580,6 +1580,16 @@ HELP_ENABLED=true
 # auto: shown, unless a custom brand logo is configured (custom logos usually
 # embed their own wordmark). Set 'true'/'false' to force either way; the
 # product name stays active in page titles and MFA labels regardless.
+#
+# Does NOT apply on a tenant's custom domain (#4241): LOGO_SHOW_NAME=true
+# cannot override the suppression there, with or without an uploaded tenant
+# logo. This is intentional — showing the operator's own product name on a
+# tenant's white-labeled domain would leak the operator's identity onto it.
+# If the wordmark unexpectedly stays hidden on what you expect to be YOUR OWN
+# canonical/subdomain host, check DOMAINS_ENABLED and CANONICAL_DOMAIN: a
+# request host that doesn't match your configured canonical domain(s) can
+# resolve to a custom-domain classification even for your own install.
+#
 # Migration note (#3612): unset previously meant always-on. If you relied on
 # the site name showing next to a custom LOGO_URL/BRAND_LOGO_URL, set
 # LOGO_SHOW_NAME=true to keep the wordmark.
--- a/.env.reference
+++ b/.env.reference
@@ -1581,14 +1581,9 @@ HELP_ENABLED=true
 # embed their own wordmark). Set 'true'/'false' to force either way; the
 # product name stays active in page titles and MFA labels regardless.
 #
-# Does NOT apply on a tenant's custom domain (#4241): LOGO_SHOW_NAME=true
-# cannot override the suppression there, with or without an uploaded tenant
-# logo. This is intentional — showing the operator's own product name on a
-# tenant's white-labeled domain would leak the operator's identity onto it.
-# If the wordmark unexpectedly stays hidden on what you expect to be YOUR OWN
-# canonical/subdomain host, check DOMAINS_ENABLED and CANONICAL_DOMAIN: a
-# request host that doesn't match your configured canonical domain(s) can
-# resolve to a custom-domain classification even for your own install.
+# On requests classified as custom, this setting cannot show the install
+# product name. This prevents the install identity from appearing on a
+# tenant's white-labeled domain (#4241).
 #
 # Migration note (#3612): unset previously meant always-on. If you relied on
 # the site name showing next to a custom LOGO_URL/BRAND_LOGO_URL, set
--- a/docs/architecture/branding.md
+++ b/docs/architecture/branding.md
@@ -141,14 +141,10 @@ logos usually embed their own wordmark). Previously, unset meant always-on —
 an install that set `LOGO_URL` and relied on the implicit wordmark should set
 `LOGO_SHOW_NAME=true` to keep it.
 
-`show_name=true` cannot re-expose the wordmark on a genuine custom (tenant)
-domain: `identityStore.showPlatformIdentity` (the A3 leak guard) suppresses it
-unconditionally there, logo or not, and `MastHead.getShowSiteName()` checks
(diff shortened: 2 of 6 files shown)


delano:
Thanks for the review. Addressed one, deferring the other:

**Suggestion 2 (test comment ambiguity)** — fair point, fixed in 29f2fe7e. The "logo or not" phrasing in the no-logo test's comment could read as claiming that one test covers both rendering branches; reworded to say explicitly that it covers the `DefaultLogo`-sentinel branch while the pre-existing `domain_logo`-set test covers the ``+wordmark-span branch.

**Suggestion 1 (split `showPlatformIdentity`/`isCustomDomain`)** — not doing this here. It's a real naming-clarity observation, but `showPlatformIdentity` is a pre-existing computed on a shared store consumed by `MastHead.vue` and asserted on directly by name in three spec files; splitting it is a public-API change to the store, not a local fix, and it's orthogonal to what #4241 asked for (documenting/testing the `LOGO_SHOW_NAME` precedence, not restructuring the guard). For what it's worth, the dual semantics you flagged are already both documented in the updated JSDoc ("False on a custom domain ... **and** whenever a per-tenant logo is present") and independently tested (`identityStore.spec.ts` — "is false whenever a per-tenant logo is present, independent of strategy"), so I don't think it's silently wrong today — just a naming choice worth revisiting on its own, if the maintainer agrees it's worth the churn across call sites.

---

PR #4279, opened 2026-08-25
Restore custom-domain guest TTL policy

delano:
## Summary
- Restore the policy distinction between canonical guests and guests creating secrets on branded custom domains: canonical hosts retain the configured 7-day default, while custom hosts use the owning organization lifetime policy of 14 or 30 days.
- Resolve and publish the effective guest ceiling from one shared server-side policy so V1/V2 enforcement and the duration picker cannot drift; tenant-resolution failures fall back to the narrower canonical policy.
- Preserve V1 legacy constructor compatibility, add regression coverage, and document the request-boundary rule in API, TTL-policy, security-review, deployment, and changelog references.

## Review guide
- Start with lib/onetime/secret_lifetime_policy.rb: it owns the canonical-versus-custom-host decision, plan and entitlement handling, outer bounds, and failure fallback.
- Review the V1/V2 delegates and ConfigSerializer build_secret_options together. The bootstrap ttl_max_anonymous name is retained for compatibility but now carries the effective ceiling for the current request host.
- This intentionally restores longer anonymous lifetimes on tenant-owned custom domains. It does not implement saved per-domain privacy-control defaults; that remains follow-up work.

## Validation
- tests/lanes/run unit
- pnpm type-check
- pnpm vitest run src/tests/composables/usePrivacyOptions.spec.ts
- bundle exec rubocop lib/onetime/secret_lifetime_policy.rb spec/unit/onetime/secret_lifetime_policy_spec.rb


delano:
Review follow-up:

1. **Greptile — unresolved plan state widens retention:** Confirmed. `Organization#limit_for` materializes an unknown limit as `0`; the current policy treated that as unlimited.

2. **Copilot — missing/unknown plan limit:** Confirmed; this is the same failure mode.

The local patch reserves `config_max` for explicit `Float::INFINITY` only. `nil`, `0`, `NaN`, and negative infinity now use the canonical fail-narrow ceiling. Added regression coverage for each case.

Validated with `tests/lanes/run unit --only spec/unit/onetime/secret_lifetime_policy_spec.rb` (12 examples) and `bundle exec rubocop lib/onetime/secret_lifetime_policy.rb spec/unit/onetime/secret_lifetime_policy_spec.rb`.

The patch is not committed or pushed yet.

PR #4280, opened 2026-08-25
Put the check-email copy back on the page

delano:
The post-signup `/check-email` page had been trimmed until its explanation sat behind a question-mark help tooltip, leaving the address alone mid-page. That made the user hover to find the one thing the page exists to say, on a dead-end screen they land on once and must act on elsewhere. The copy is back on the page — "We sent a verification link to" above the address, "Click the link in the email to activate your account." below — and the AuthView brand icon is restored so the page doesn't open on a bare heading.

`web.auth.check_email.sent_to` / `.instructions` return with their original text and content hashes (`e795f9b7` / `389221ed`); the tooltip-only `.help` key is dropped. Those two keys never existed in a non-en locale, so all 29 are fresh translations, seeded from each locale's deleted `help` string (which carried both halves of this copy, already reviewed) and cross-checked against its `sent_to_generic` sibling.

Reviewer note on the locale diff: `sent_to` is an English sentence fragment that continues into the address below it, and that shape doesn't survive every language. 11 locales keep it; 18 restructure to a label ending in a colon because a bare preposition reads as truncation — verb-final German (`Verifizierungslink gesendet an:`), particle-final Japanese nominalized to 送信先, Turkish needing `şu adrese` to carry the dative so the address stays unsuffixed, Czech/Polish/Russian/Slovene needing an explicit "address" noun for the case to attach to. The colon in those locales is intended, not drift from en.

Tests: 11/11 in `CheckEmail.spec.ts` — the help-affordance case is replaced by one asserting the copy above and below the address, plus a new case covering the instruction line surviving the no-address fallback (a shared link or new tab, where the generic line replaces the address). Note `sent_to` is a string-prefix of `sent_to_generic`, so that case asserts the address element's absence structurally rather than by text match. `prettier --write` also fixed two pre-existing formatting hunks in the touched files that already failed `--check` at HEAD.

### DIFF
--- a/locales/content/ar/session-auth.json
+++ b/locales/content/ar/session-auth.json
@@ -436,13 +436,17 @@
     "text": "تحقّق من بريدك الإلكتروني",
     "source_hash": "77322879"
   },
+  "web.auth.check_email.sent_to": {
+    "text": "أرسلنا رابط تحقق إلى:",
+    "source_hash": "e795f9b7"
+  },
   "web.auth.check_email.sent_to_generic": {
     "text": "أرسلنا رابط تحقق إلى عنوان بريدك الإلكتروني.",
     "source_hash": "4e5510af"
   },
-  "web.auth.check_email.help": {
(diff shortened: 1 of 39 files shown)


delano:
Ready after checks

PR #4297, opened 2026-08-26
[#4288] Cut the secret link page's render path

### LINKED ISSUE #4288: Large render-blocking asset on the secret link page
Sentry: [FRONTEND-17K](https://catch.onetimesecret.com/organizations/73ffddd61e4fbb7e/issues/FRONTEND-17K) (36 events), [FRONTEND-16S](https://catch.onetimesecret.com/organizations/73ffddd61e4fbb7e/issues/FRONTEND-16S) (20)

**56 events.**

## Symptom

Sentry performance issue `Large Render Blocking Asset` on the secret link view.

## Why it matters

This is the highest-intent page in the product and the one most often opened on a phone,
on a mobile network, inside someone else's webview. It is also the page where a slow
first paint reads to the recipient as "the link is broken".

## Fix direction

- Identify the blocking asset from the Sentry event detail (likely the main CSS bundle or a webfont).
- Inline critical CSS for the secret view; defer the rest.
- `font-display: swap`, and self-host rather than blocking on a font CDN.
- Confirm the secret view is not pulling in the full dashboard bundle.

## Acceptance

Render-blocking asset issue clears; LCP on the secret page measurably improves.

delano:
## Summary

[#4288](https://github.com/onetimesecret/onetimesecret/issues/4288)

The heavy asset on the secret link page is the JS bundle, and **89% of it is translations nobody on that page reads.**

`src/i18n.ts` eagerly imported all 30 pre-merged locale files, putting ~8.8 MB of JSON inside the single customer chunk. Every recipient downloaded all 30 languages to read one, before the page could render a character.

| | before | after |
|---|---|---|
| `main.js` raw | 9.93 MB | 2.53 MB |
| `main.js` gzipped | **2.66 MB** | **695 KB** (−74%) |
| font preloads | 8 files, ~675 KB, highest priority | none |
| `style.css` gzipped | 36 KB (blocking) | unchanged |

### What changed

**Locales out of the bundle.** English stays in — it is the source language and the tail of every fallback chain, which makes it the safety net when a locale fetch fails. The other locales are emitted as standalone hashed JSON assets (`?url` keeps the content out of the chunk) and fetched on demand: one request, for the one language the visitor reads. `main.ts` mounts after that fetch settles, so a non-English recipient doesn't watch the page repaint from English into their own language.

A failed locale fetch is not fatal: `loadLocaleMessages` resolves to `null`, the app mounts, and the page renders in English rather than raw message keys.

**The shell preloads the active locale's asset** (`as="fetch"`, `crossorigin`). Without the hint that fetch cannot start until the chunk has downloaded *and parsed*; with it, the two overlap. The backend picks the file out of the Vite manifest and there is no bundled-locale constant duplicated on the Ruby side — the build emits no asset for the locale it bundles, so `en` simply gets no preload.

**No more font preloads.** The shell preloaded every font in the manifest — eight files, ~675 KB including four legacy `.woff` fallbacks no modern browser uses — at the browser's highest fetch priority, ahead of the bytes the page cannot paint without. They are now discovered when the stylesheet parses and load at normal priority, with `font-display: swap` (was `fallback`) so text paints immediately in the serif fallback and restyles when Zilla Slab arrives.

### Deliberately not changed

The stylesheet stays render-blocking. It is the only asset the browser actually reports as `renderBlockingStatus: "blocking"`, but at 36 KB gzipped it is far under the detector's size floor, and deferring it in an SPA trades a slightly earlier first paint for FOUC risk on the highest-intent page in the product. Cutting ~2.6 MB of competing traffic ahead of it is the better lever.

Also unchanged: the single-chunk / no-code-splitting build. That is load-bearing for the CSP nonce model, and this change gets the win without touching it.

## Related Issues

Fixes #4288

## Test Plan

**Measured, not assumed.** Probed the built bundle in Chromium (`PerformanceResourceTiming`) to identify what the browser actually flags: `style.css` → `blocking` (36 KB), `main.js` and every font preload → `non-blocking`. That ruled the CSS and fonts out on size and pointed at the bundle.

Booted the real built app in Chromium against a local static server:

- **German** — app mounts, renders fully in German, **exactly one** locale request, no preload-mismatch warning.
- **English** — app mounts, **zero** locale requests (bundled).
- **Locale asset 404s** — app still mounts and renders in English, not raw keys, not blank.

The preload/fetch pairing was verified empirically across four combinations; three of them cause a **double download**. Only `crossorigin` + default `fetch(url)` credentials matches (1 request, no warning) — that pairing is now commented on both sides.

Dev server checked too: the `?url` glob resolves to a working `/dist/@fs/...` URL serving `application/json`.

- `pnpm vitest run` — 8913 passed, 394 files, 0 failures
- `pnpm type-check` — clean
- `eslint` + `prettier --check` on changed files — clean
- New `apps/web/core/spec/views/helpers/vite_manifest_spec.rb` covers: active-locale preload emitted and `crossorigin`; `de_AT` not served `de`'s asset; nothing for a bundled/unknown locale; non-locale-code input rejected; unhashed (`build:local`) asset names matched; no font preloads emitted.

**Not run here:** the Ruby lane (`tests/lanes/run unit`) — this environment has no Docker daemon for the backing services, and no `.test-mode` sentinel. The new spec is syntax-checked and its subject method was exercised directly against a real build manifest; CI runs the lane.

https://claude.ai/code/session_01KVuYZ7U8XAhzHANyo8domV

---
_Generated by [Claude Code](https://claude.ai/code/session_01KVuYZ7U8XAhzHANyo8domV)_

### DIFF
--- a/apps/web/core/spec/views/helpers/vite_manifest_spec.rb
+++ b/apps/web/core/spec/views/helpers/vite_manifest_spec.rb
@@ -1,0 +1,186 @@
+# apps/web/core/spec/views/helpers/vite_manifest_spec.rb
+#
+# frozen_string_literal: true
+
+require 'fileutils'
+require 'json'
+require 'tmpdir'
+
+require_relative '../../../../../../spec/spec_helper'
+
+require_relative '../../../views/helpers/vite_manifest'
+
+# Tests for Core::Views::ViteManifest, the helper that turns a Vite manifest
+# into the <script>/<link> tags the HTML shell emits.
+#
+# What these lock down is the render path of the secret link page (#4288):
+# what the shell tells the browser to fetch at high priority before it can
+# paint anything. Two invariants:
+#
+#   1. The shell preloads the ACTIVE locale's messages asset, and only that
+#      one. Whether the fetch in src/i18n.ts overlaps the JS bundle or queues
+#      behind it hangs on this hint being present and correct.
+#   2. The shell preloads no fonts. It used to preload every font in the
+#      manifest -- eight files, ~675 KB, ahead of the bytes the page actually
+#      needs to render.
+RSpec.describe Core::Views::ViteManifest do
+  # Bare includer: the helper needs nothing from BaseView, and going through
+  # the view stack would make these assertions about serializers instead.
+  let(:includer_class) do
+    Class.new do
+      include Core::Views::ViteManifest
+    end
+  end
+
+  subject(:helper) { includer_class.new }
+
+  let(:public_dir) { Dir.mktmpdir('vite-manifest-spec') }
+  let(:nonce) { 'test-nonce' }
+
+  # Mirrors a real build: one chunk, one stylesheet, per-locale JSON assets
+  # for every locale EXCEPT the one bundled into the chunk (en), plus the
+  # font files the stylesheet references.
+  let(:manifest) do
+    {
+      'main.ts' => {
+        'file' => 'assets/main.abc123.js',
+        'name' => 'main',
+        'src' => 'main.ts',
+        'isEntry' => true,
+        'assets' => [
+          'assets/de.d3adb33f.json',
+          'assets/de_AT.c0ffee12.json',
+          'assets/pt_BR.f00dcafe.json',
+          'assets/ZillaSlab-Regular.aaaa1111.woff2',
+          'assets/ZillaSlab-Regular.bbbb2222.woff',
+          'assets/ZillaSlab-Bold.cccc3333.woff2',
+        ],
+      },
+      'style.css' => { 'file' => 'assets/style.def456.css' },
+    }
+  end
+
+  def write_manifest(data = manifest, filename: 'manifest.json')
+    dir = File.join(public_dir, 'dist', '.vite')
+    FileUtils.mkdir_p(dir)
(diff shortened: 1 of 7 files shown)


delano:
**The PR makes locale flashing worse.** 

After review:
- The PR makes locale flashing worse for `?locale=de_DE`. The server does not recognize or preload that form, so the page first shows English and then repaints in German after the client fetches translations.
- Returning users can see the same issue. Their saved locale is only available in `sessionStorage` or their account, so the server loads the wrong language first; the client later corrects it through a visible network-driven repaint.
- The admin console is a direct regression: it has no locale loading gate, so non-English locales never load and the UI remains English.
- Other confirmed risks include a blank page when the mount-time locale fetch stalls, concurrent `setLocale` calls applying results out of order, unsupported brand locale formats rendering custom domains in English, and an English flash in the branding preview.
- The underlying problem is that the server, client initialization, and language store each choose a locale using different precedence rules. Fixing one path will not eliminate the flash.
- A complete fix requires the server to access the client’s chosen locale, such as through a locale cookie, or to preload all relevant normalized and fallback locale options.

**Status:** All 14 findings have been verified. Twelve remain: 10 confirmed and 2 plausible. Two were disproven.

Issue #4305, opened 2026-08-27
Preserve `redirect` through the complete signup and verification flow


delano:
### Problem

The `redirect` query parameter is intended to return a user to the internal page that initiated authentication.

It works for direct sign-in, for example:

```text
/signin?redirect=/secret/abc123
```

However, it does not survive a normal new-user journey:

```text
/signup?redirect=/some/internal/page
→ /check-email?redirect=/some/internal/page
→ verification email link
→ /verify-account?key=…
→ /signin
→ /
```

[`useAuth.verifyAccount()`](https://github.com/onetimesecret/onetimesecret/blob/main/src/shared/composables/useAuth.ts) always navigates to plain `/signin` after successful verification. It drops the validated return destination before the user can sign in.

This affects arbitrary internal return paths, including protected secret/reveal and account pages, and makes the authentication flow lose the user’s original context.

### Expected behavior

For a valid internal redirect:

```text
/signup?redirect=/some/internal/page
```

a successful signup, verification, and subsequent sign-in should end at:

```text
/some/internal/page
```

The same behavior must remain correct for direct sign-in.

### Scope

- Preserve a validated internal `redirect` across:
  - signup;
  - the check-email screen;
  - the verification link, including when opened in a new browser session;
  - post-verification sign-in;
  - MFA, if required.
- Preserve the full internal path, including query string and hash fragment.
- Keep `redirect` separate from billing plan-selection state.
- Define precedence when an auth flow also supplies a billing intent. A valid paid-plan intent should take precedence; `redirect` is the fallback navigation target.
- Continue rejecting invalid, external, or protocol-relative redirect targets at every trust boundary.

### Acceptance criteria

- [ ] Direct sign-in at `/signin?redirect=/account/settings/security` lands at `/account/settings/security` after successful authentication.
- [ ] Signup at `/signup?redirect=/account/settings/security` lands at that path after email verification and sign-in.
- [ ] The flow works when the verification link is opened in a fresh browser session, not only the session that submitted signup.
- [ ] Query strings and fragments survive, e.g. `/secret/abc?view=raw#content`.
- [ ] `/signup?redirect=https://attacker.example` and equivalent protocol-relative inputs cannot redirect externally.
- [ ] MFA preserves the return destination until authentication is fully complete.
- [ ] A browser-level E2E test covers the complete signup → email verification → sign-in → return path, rather than only helper or route-query behavior.

### Relevant code

- `src/apps/secret/support/Pricing.vue`
[...]

PR #4308, opened 2026-08-27
[#4305] Follow-ups: shared redirect-parity fixture + full-auth E2E CI lane

delano:
Stacked on #4307. Two follow-ups that turn the weakest points of that PR into enforced guarantees.

## 1. Shared parity fixture for the twin redirect validators

`Onetime::Utils::RedirectPaths.safe_internal_path?` (Ruby) and `isValidInternalPath` (TS) are one security ruleset in two languages, previously kept in sync only by "change both or neither" comments.

- **`tests/fixtures/redirect_path_cases.json`** — 87 accept/reject cases (24 accept / 63 reject) in 14 groups, including both ~2KB length-boundary strings. Same pattern as `tests/fixtures/email_redaction_corpus.json`.
- Both `spec/unit/onetime/utils/redirect_paths_spec.rb` and `src/tests/utils/redirect.spec.ts` generate one example per case, nested under the group taxonomy. A rule relaxed on one side turns exactly one suite red.
- Anti-erosion guards: non-empty case list, unique ids, 8 pinned #4305 acceptance-criteria ids, and length boundaries pinned against each implementation's real constant (`MAX_PATH_LENGTH` / newly exported `MAX_REDIRECT_LENGTH`). Weakening the ruleset now requires a visible three-file change.
- Language-local cases the JSON can't express stay hand-written (Ruby: nil/non-String/invalid encoding; TS: undefined/null/number/object/array).
- Validator sources: comment-only changes; non-comment diff is empty.

## 2. `e2e-full-auth.yml` — a CI lane for the journey that caught the bug

`e2e/auth/signup-redirect-preservation.spec.ts` needs full-mode Rodauth with `verify_account` enabled, a real SMTP path, and a mail worker — none of which the container E2E lane provides (it runs simple mode with `AUTH_AUTOVERIFY=true`, and `etc/defaults/auth.defaults.yaml` force-disables verify_account under `RACK_ENV=test`). Until now the strongest guard for #4305 was local-only.

The new lane, shaped like `visual.yml` (runner-hosted app, not the OCI image):

- `compose.test.yml` gains a digest-pinned **Mailpit** service (SMTP `127.0.0.1:2125`, HTTP `127.0.0.1:12125`), following the 21xx port scheme.
- Boots with `RACK_ENV=production`, `AUTHENTICATION_MODE=full`, `AUTH_AUTOVERIFY=false`, `AUTH_VERIFY_ACCOUNT_ENABLED=true`, sqlite boot-migrated authdb, `EMAILER_MODE=smtp` into Mailpit, `JOBS_ENABLED=true` with RabbitMQ + a real `bin/ots worker` (mail is enqueued, never inline).
- `bin/ots worker --check` declares the queue topology before the server boots, so no publish can race the queue into the void.
- Runs `signup-redirect-preservation` + `sso-missing-email` (`--project=chromium`); the other three `e2e/auth/` specs are deliberately excluded (dormant behind env gates / need billing) with the rationale in-file.
- Flaky gate (retry-only passes fail the job) + failure forensics: server/worker logs, Mailpit mailbox dump, RabbitMQ queue depths, Playwright traces.

## Review

An adversarial review pass ran over the full stacked diff before pushing: all 87 fixture verdicts hand-traced through both implementations (no wrong verdicts, no cross-language divergence), workflow env/step-ordering/port matrix verified against the code. Its findings are fixed in the last two commits — notably the `pull_request` `branches:` filter is dropped so **this stacked PR itself runs the new lane** (the paths filter keeps it off unrelated PRs), and `timeout-minutes` was raised to 45 so a job timeout can't kill the artifact-upload steps on the exact run whose traces matter.

⚠️ First real execution of the lane is this PR's checks — it was previously actionlint-verified only. If it goes red, the likely suspects (in order): `bin/setup` on a cold ubuntu-24.04 runner, worker log-line grep, Playwright runtime budget.

## Test evidence (local)

- `spec/unit/onetime/utils/redirect_paths_spec.rb`: 100 examples, 0 failures
- `src/tests/utils/redirect.spec.ts`: 112 passed (full `src/tests/utils/`: 655 passed)
- `actionlint` clean on the workflow


delano:
First-run report for the new lane, as promised in the PR body.

**Run 1 failed** — but not on any of the predicted suspects. The entire stack booted clean in under 4 minutes (bin/setup, build, queue topology, server, worker); the failure was `POST /auth/create-account` → 403 on all attempts, and the lane's own forensics steps produced the diagnosis directly from the server log:

- `session.secure` is absent from rendered config unless `SSL=true`, and `boot.rb#ssl_enabled?` defaults it to **true** under `RACK_ENV=production` — even with `SSL=false`
- `Onetime::Session` then refuses to write a Secure cookie over plain HTTP (`cookie NOT written` warn)
- no session persists → no CSRF token in session → `[CsrfResponseHeader] CSRF 403 session-continuity break` on every auth POST

This never showed locally (the local full-mode recipe sits behind Caddy with real TLS) and never bit `bin/visual` (same env, but it only screenshots — its `SSL=false pins http:// links/cookies` comment was true for links, silently false for cookies).

**Fix** (`e89d31255f`): the session block in `etc/defaults/config.defaults.yaml` emits `secure: false` only for the exact string `SSL=false`. Unset stays absent, so the production fallback to Secure is untouched; `SSL=true` unchanged. `ASSUME_HTTPS=true` was considered and rejected — it upgrades the scheme for redirects/HSTS on a server with no TLS listener.

**Run 2 passed**: `signup → verify email → signin (real SMTP)` green in 3m32s — the full journey through RabbitMQ, the EmailWorker, and Mailpit, with the redirect preserved end to end.

PR #4314, opened 2026-08-30
Fix TypeScript type-check warnings in test code

delano:
## Summary

`pnpm run type-check:tests` (the `typescript-unit` CI job) was failing with
638 TypeScript errors across 87 test files — see the
[failing run on #4311](https://github.com/onetimesecret/onetimesecret/actions/runs/33294368641/job/99211502453?pr=4311).
That check isn't required for merge, so it had gone unaddressed for a while.

This PR fixes the underlying type mismatches — mostly test fixtures/mocks
that had drifted from the real v3 schema shapes and component/composable
APIs that production code now uses — rather than suppressing errors with
`any` or `@ts-expect-error`. No production runtime behavior is changed.

**Result: `pnpm run type-check:tests` now passes clean (0 errors, was 638).**
Full suite re-run after all fixes: **403 test files passed / 1 pre-existing
skip, 9227 tests passed / 86 skipped / 9 todo, 0 failures.**

Along the way, one genuine (pre-existing, unrelated to types) runtime test
bug was found and fixed in `authStore.spec.ts`: a timer spy was targeting
`vi.setSystemTime` instead of `setTimeout`, which also masked a
`vi.spyOn`/`mockRestore` interaction that left `globalThis.setTimeout`
broken for later tests once corrected naively — fixed by manually
saving/restoring the timer reference instead.

### Two production-code observations (not changed here, flagged for a maintainer)

Both were found while fixing test types; neither is a behavior bug today,
and fixing either is a production change outside this PR's scope:

1. `src/schemas/api/incoming/requests/incoming-secret.ts`: `memo` is
   `z.string().optional().default('')`, so zod's output type makes it
   *required* — contradicting the field's own docstring, which says it's
   optional. Harmless today (the one real caller always populates it), but
   a latent trap for the next caller.
2. `ObjId`/`ExtId` are declared via two independent `unique symbol` brands —
   one in `src/schemas/utils/identifiers.ts` (backs `lenientObjIdSchema`/
   `lenientExtIdSchema`) and one in `src/types/identifiers.ts` (backs
   `toObjId`/`toExtId`). They're nominally incompatible despite identical
   shape/name (deliberate, per that file's own comment, to avoid a circular
   import), so a value built with `toObjId()`/`toExtId()` won't type-check
   against a field typed through the `lenient*` schemas. Worth consolidating
   or documenting more prominently.

## Related Issues

None filed — this addresses the CI failure linked above directly.

## Test Plan

- `pnpm run type-check:tests` — 0 errors (was 638).
- `pnpm exec vitest run src/tests/` — 403 passed / 1 skipped test files,
  9227 passed / 86 skipped / 9 todo tests, 0 failures.
- Each changed file/group was also re-run individually against its
  "must stay green" consumers during development to catch regressions
  early (see commit messages for per-group detail).

### DIFF
--- a/src/tests/apps/admin/AdminCustomerSessionsSection.spec.ts
+++ b/src/tests/apps/admin/AdminCustomerSessionsSection.spec.ts
@@ -29,7 +29,10 @@ vi.mock('@/shared/components/icons/OIcon.vue', () => ({
 }));
 
 import AdminCustomerSessionsSection from '@/apps/admin/components/AdminCustomerSessionsSection.vue';
-import { colonelCustomerSessionsResponseSchema } from '@/schemas/api/internal/responses/colonel-customer-sessions';
+import {
+  colonelCustomerSessionsResponseSchema,
+  type AdminCustomerSession,
+} from '@/schemas/api/internal/responses/colonel-customer-sessions';
 import { createTestI18n } from '@tests/setup';
 
 const i18n = createTestI18n();
@@ -40,7 +43,7 @@ const USER_ID = 'ur_abc123';
 const COUNTRY_HEADER = 'web.admin.customers.detail.sessions.columns.country';
 const UNKNOWN = 'web.admin.customers.detail.sessions.unknown';
 
-function sessionRow(overrides: Record<string, unknown> = {}) {
+function sessionRow(overrides: Partial<AdminCustomerSession> = {}): AdminCustomerSession {
   return {
     session_id: 'sid_1',
     user_id: USER_ID,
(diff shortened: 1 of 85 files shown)


delano:
Re: the automated review above (posted 14s after PR creation, so it ran against the initial Wave-1-only commit, before `receipt.compat.spec.ts`/`receipt.roundtrip.spec.ts` were touched) — the 3 P1s are already fixed on the current head. Re-verified just now against `496d5480`:

```
pnpm exec vitest run src/tests/schemas/shapes/receipt.compat.spec.ts src/tests/schemas/shapes/receipt.roundtrip.spec.ts
 Test Files  2 passed (2)
      Tests  137 passed (137)
```

All three traced back to the same root cause the review identified (the `is_viewed`/`is_received`/`received`/`viewed` → `is_previewed`/`is_revealed`/`revealed`/`previewed` rename in `serializers.ts`/`receipt.fixtures.ts`): the two spec files were updated in a later commit to use the current field names, including the exact assertions called out (`v3Wire.is_previewed`, the `is_previewed` override key, and `canonical.revealed`/`canonical.previewed` as the comparison source in the "Deprecated Field Aliasing" block).

The 2 non-blocking follow-ups are also already addressed, both with inline comments explaining the reasoning:
- `support/axe.ts`: the deep `vitest-axe/dist/matchers` import is deliberate — the public entry's declarations are type-only under `isolatedModules` (TS1362) even though the runtime export is real; `dist/matchers` has a correct, value-preserving declaration for the same underlying module. Confirmed at runtime too: all 6 `shared/a11y/*.a11y.spec.ts` files pass in the full suite run.
- `bootstrap-test-schema.ts`: the `as z.output<...>` casts are intentional, not a truncation bug — zod v4's `.default()` substitutes the literal verbatim (no cascade), and `bootstrap-schema-contract.spec.ts` pins that exact partial shape. Cascading via `.parse()` instead would change the runtime value and break that assertion.

Full suite on the current head: 403 test files passed / 1 pre-existing skip, 9227 tests passed / 86 skipped / 9 todo, 0 failures.

---

Issue #4323, opened 2026-09-01
Epic: Colonel Hardening


delano:
Tracking issue for the highest-urgency cluster from the colonel operational gap review (2026-08-31).

**Theme:** every safety interlock on destructive colonel actions currently lives in the browser. Typed-confirmation dialogs, danger flags, and warnings are all Vue-side; the API executes purge/delete/revoke/role-change on a bare authenticated request. There is no step-up re-authentication, no rate limiting on `/api/colonel/*`, and no guard against a colonel demoting the last colonel. A single leaked colonel session cookie is unbounded, un-throttled destructive capability.

**Definition of done:** a colonel session alone is no longer sufficient for destruction — confirmation is checked server-side, destructive verbs require a fresh step-up window, role changes have interlocks, and the colonel API surface is rate-limited.

Sub-issues attached below. Related: the audit-side counterpart lives in the Audit Integrity epic; full-mode (Rodauth/SQL) auth administration is deliberately out of scope here — it is being chartered as a separate codebase.

PR #4363, opened 2026-09-01
Colonel Hardening: server-side confirmation, step-up sudo, interlocks, rate limits, opaque session IDs, admin lifetime & network gating (#4323)

### LINKED ISSUE #4326: Enforce destructive-action confirmation server-side
**Severity: Critical** · gap review A-1 (new finding)

Typed-confirmation tokens exist only in `AdminConfirmDialog.vue`. No colonel logic class accepts or checks a confirmation parameter, so purge-user, delete-secret, revoke-all-sessions and DLQ-purge execute on a bare POST/DELETE with a colonel session. All confirmation is theater once you're past the browser.

**Fix shape:** thread a `confirm` param (e.g. the resource identifier or a typed token) through the destructive logic classes and reject requests without it, mirroring how `dry_run` is already threaded. The dry-run-default pattern on `delete_organization.rb` is the template: default to preview, require explicit confirmation to execute.

**Evidence:** `src/apps/admin/components/AdminConfirmDialog.vue` (client-only); grep of `apps/api/colonel/logic/` shows no confirmation param in any destructive class.

### LINKED ISSUE #4327: Add step-up re-authentication (sudo window) for destructive colonel actions
**Severity: Critical** · gap review A-2 (known — colonel-ui gap analysis §3.3, rated P1 there)

No sudo mode, no password re-entry, no MFA prompt, no session-age check exists anywhere in the colonel ops — and `account_management.rb:126` sets `require_password_confirmation? false`. MFA is not required to hold the colonel role.

**Fix shape:** a ~10-minute elevation window — password or MFA re-entry — recorded on the session and checked **server-side** by the destructive logic classes (not the router), so it composes with the server-side confirmation work. The in-repo session gap analysis notes this is also the prerequisite blocking elevation-state display and any future impersonation feature.

**Evidence:** `apps/web/auth/config/features/account_management.rb:126`; grep across `apps/api/colonel/logic/` for any re-auth/session-age check (none).

### LINKED ISSUE #4328: Add interlocks to SetUserRole and other self-target actions
**Severity: Critical** · gap review A-3 + A-8 (new findings)

`SetUserRole` has no self-target guard (unlike `PurgeUser`), no peer-colonel guard, no last-colonel count, and no dry-run. In the UI, `setRole` is not in `DANGER_ACTIONS`, so promoting an account to colonel is a plain one-click confirm. A colonel can demote every colonel including themselves; recovery is CLI-only.

**Fix shape:**
- Block (or require force + step-up for) self-demotion and demotion of the last remaining colonel. The org layer already has the sole-owner pattern to copy: `apps/api/colonel/logic/colonel/memberships/remove.rb:28`.
- Add `setRole` (promotion to colonel especially) to `DANGER_ACTIONS` in the admin UI.
- Same self-target check for `DeleteSession` (a colonel can currently delete their own active session with no warning — minor, but it's the same guard).

**Evidence:** `apps/api/colonel/logic/colonel/set_user_role.rb`, `memberships/remove.rb:28`, `delete_session.rb`, admin UI `DANGER_ACTIONS` constant.

### LINKED ISSUE #4329: Rate-limit the colonel API surface
**Severity: Critical** · gap review A-4 (new finding)

The rate limiter registry is perimeter-only (login, feedback, etc.); no middleware or `limit_action` covers `/api/colonel/*`. A scripted compromise of a colonel session can purge at wire speed — and, because the audit trail trims at a 10k count cap, evict the evidence of its own actions while doing so (see the Audit Integrity epic).

**Fix shape:** a modest per-session limiter on colonel mutations (destructive verbs tighter than reads). Even a generous ceiling converts "unbounded" into "bounded and alarmed."

**Evidence:** `lib/onetime/ratelimit/registry.rb:41–93` (registry contents; nothing colonel-scoped).

### LINKED ISSUE #4330: Stop rendering raw session IDs in the sessions console
**Severity: Critical** · gap review A-5 (known — colonel-ui gap analysis §1.6, rated P0 there; re-verified current 2026-08-31)

`Sessions::Store.summarize` returns the raw `session_id` (live bearer credential); `AdminSessions.vue` displays it and uses it in the revoke URL. Any colonel screen-share, log capture, or browser extension sees credentials that can hijack any listed session.

**Fix shape:** replace with an opaque handle or digest (e.g. HMAC of the sid) resolved server-side for revoke operations. The gap analysis flags this as a prerequisite for any impersonation feature.

**Evidence:** `lib/onetime/operations/sessions/store.rb:190`, `src/apps/admin/views/AdminSessions.vue:66,219`.

### LINKED ISSUE #4331: Separate admin session lifetime from the tenant app
**Severity: High** · gap review A-6 (known — colonel-ui gap analysis §3.1–3.2, §3.13–3.14)

One `onetime.session` cookie serves colonel and customer alike, with the same 24h-rolling TTL: no shorter admin lifetime, no absolute cap enforced on the request path, no idle timeout. `AdminNetworkIsolation` compensates at the network layer but does not substitute for session hygiene.

**Fix shape:** enforce a shorter idle + absolute lifetime for sessions holding the colonel role (checkable in the existing session-auth path without splitting the cookie as a first step). Splitting the admin surface onto its own cookie/scope is the fuller fix and pairs with the step-up work.

**Evidence:** session middleware + `docs/specs/colonel-ui/spec-session-gap-analysis.md` §3.

### LINKED ISSUE #4332: Apply network=admin gating beyond the proxy-headers debug route
**Severity: High** · gap review A-7 (new finding)

The strongest network control in the stack — the `network=admin` annotation (404-on-miss, host-provenance-aware CIDR check) — is applied to exactly one route: `GET /system/proxy-headers`, a read. All 48 mutating colonel routes ride the general posture, where the CIDR allowlist is empty by default on self-hosted installs.

**Fix shape:** annotate the destructive route groups (purge, delete, revoke-all, DLQ purge, role changes) with `network=admin`, or make the annotation the default for mutating colonel routes with an explicit opt-out. Document the empty-allowlist default so self-hosted operators know the gate is open until configured.

**Evidence:** `apps/api/colonel/routes.txt:40`, `lib/onetime/application/network_requirements.rb`.

delano:
## Summary

[#4323](https://github.com/onetimesecret/onetimesecret/issues/4323) — Epic: Colonel Hardening

Every safety interlock on destructive colonel actions previously lived in the browser: typed-confirmation dialogs, danger flags, and warnings were all Vue-side, while the API executed purge/delete/revoke/role-change on a bare authenticated request. There was no step-up re-authentication, no rate limiting on `/api/colonel/*`, and no guard against a colonel demoting the last colonel. A single leaked colonel session cookie was unbounded, un-throttled destructive capability.

This stack moves enforcement server-side into the **logic classes** (so it composes and covers the CLI, not just the router) and adds the missing interlocks. A colonel session alone is no longer sufficient for destruction.

The work is a stack of one commit per sub-issue, ordered so shared primitives land first:

| # | Commit | Sub-issue |
|---|--------|-----------|
| 1 | `fix(colonel): replace raw session ids with opaque handles in the console` | #4330 |
| 2 | `feat(colonel): enforce destructive-action confirmation server-side` | #4326 |
| 3 | `feat(colonel): add a step-up (sudo) window for destructive colonel actions` | #4327 |
| 4 | `feat(colonel): interlock self-demotion, last-colonel lockout and self-revoke` | #4328 |
| 5 | `perf(colonel): rate-limit the colonel API surface` | #4329 |
| 6 | `feat(colonel): bound admin-surface session lifetime` | #4331 |
| 7 | `feat(colonel): gate destructive colonel routes on network=admin_if_configured` | #4332 |

Plus three follow-up `fix(colonel): … (#4323)` commits from whole-stack validation (order-independent test stub, a layout realignment, and a missing test prop).

### What changed

**Shared primitives (land in #4330/#4326/#4327):**
- `ColonelAPI::Logic::DestructiveAction` — a concern included into the colonel logic base carrying a hard five-step **guard order**: role check → target resolution → `require_elevation!` (tier 1) then `require_confirmation!` → per-verb interlocks → `charge_destructive_budget!` (tier 1). Elevation precedes confirmation so there is no token oracle; interlocks follow both so their 422s can't be used to probe "who is the last colonel"; the tight destructive budget is charged last so only executable attempts consume it.
- `ColonelAPI::DestructiveActions` — a machine-readable tier registry (TIER1 = 15 classes, TIER2 = 10, plus an explicit `TIER3_REVIEWED` map) that a spec cross-checks against `routes.txt` so no mutating route can ship un-triaged.
- New error classes (`ConfirmationRequired`, `ElevationRequired`, `ElevationFailed` < `Forbidden`; `GuardMisconfigured` < `Problem`) with exact-class Otto handlers, excluded from failure auditing so a cookie holder cannot flush the count-capped audit trail.

**Server-side confirmation (#4326):** destructive logic classes require a confirmation token, carried in the `X-OTS-Confirm` request header (percent-encoded, POST and DELETE alike) rather than a query parameter — 18 of 25 confirmation tokens are PII (target email, org name, domain) that a URL would write into every access log and browser history. `dry_run` classes gate only the apply path.

**Step-up / sudo window (#4327):** a 10-minute elevation window stored as an identity-bound sidecar (`{extid, exp}`, never a bare epoch, so login/rotation cannot let identity B inherit A's elevation). Two factors — password re-entry (branching on auth mode) and `recent_auth` for accounts that cannot use a password (SSO-only). New `GET/POST/DELETE /api/colonel/elevation`. Both login paths drop the field.

**Interlocks (#4328):** `Customers::RoleSupport` re-validates `active_colonels` against the authoritative role field (`exists? && role == 'colonel' && verified?`) rather than the upward-drifting role index, which would fail open. Refusals for self-demotion, last-colonel demotion, and unverifying the last colonel land in the **Operations** layer (so `bin/ots` is covered too) via a single `build()` exit so the `:failure` audit can't be skipped. Fixes `SetUserRole`'s `sanitize_identifier` → `sanitize_account_identifier`, and routes self-targeted revoke-all to "revoke all except current" (a containment step for a leaked cookie).

**Rate limits (#4329):** `Onetime::Security::ColonelRateLimiter` with four buckets, all keyed on `cust.extid` (never the bearer session id): a broad mutation bucket charged from the logic base constructor (covers all 46 mutating verbs), a tight destructive bucket charged last with a lockout window, a handle-resolve bucket, and the elevation bucket. Registry rows kept byte-identical to the key builders so `bin/ots ratelimit` and reset can see them.

**Admin session lifetime (#4331):** a 1h idle / 12h absolute bound enforced in the session-auth strategy, scoped to `/api/colonel` so it never expires the shared cookie for tenant-app use. `0` disables a bound; a missing session sidecar skips the idle bound rather than logging anyone out.

**Network gating (#4332):** a new `network=admin_if_configured` route strategy applied to the 15 tier-1 routes — enforces isolation where fully configured, falls through when unconfigured (so default self-hosted installs are not bricked), but fails closed if the isolation middleware is ever removed. `/system/proxy-headers` keeps the strict `network=admin` token. Operator docs updated with the five-posture table and the empty-allowlist residual-risk note.

**Frontend:** the admin console is kept working end-to-end — opaque handles throughout `AdminSessions.vue`, `confirmHeaders` threading, `setRole`/`unverify` added to `DANGER_ACTIONS`, a no-polling elevation composable + prompt/banner, and `[ADMIN_SESSION_EXPIRED]` handling. Zod/TS schemas and `.env.reference` mirror the new server contracts.

## Related Issues

Closes #4326
Closes #4327
Closes #4328
Closes #4329
Closes #4330
Closes #4331
Closes #4332

Part of the #4323 epic.

## Test Plan

Whole-stack validation on Ruby 3.4.10 (all green):

- `bundle exec rake spec:fast` — 5,438 + 5,587 + 56 examples, 0 failures
- `bundle exec rspec apps/api/colonel/spec` — 533 examples, 0 failures (up from 139 pre-epic)
- `bundle exec rspec spec/integration/all` (simple mode) — 952 examples, 0 failures
- `bundle exec rake spec:apps:web_auth` — 1,417 examples, 0 failures (the `set_role` / `set_verification` / `change_email` op interlocks)
- `bundle exec rake try:unit` — 8,244 tryout cases, 0 failed
- `bundle exec try try/integration/api/colonel` (simple mode) — 372 cases, 0 failed (real audit-trail writes)
- `pnpm exec vitest run src/tests/apps/admin` — 888 passed / 1 skipped
- `pnpm run type-check` + `type-check:tests` + `eslint` on the 53 changed frontend files — clean
- `bundle exec rubocop` on the 66 changed Ruby files — clean (4 pre-existing convention offenses on untouched lines remain)
- Boot smoke — `bin/ots --version` and a full Rack::URLMap load confirm tier registry (TIER1=15, TIER2=10, TIER3=22) and guard composition wire up with no require-order breakage

New coverage includes the guard-order contract, the tier-registry-vs-routes cross-check, dry-run side-effect-freedom, the identity-bound elevation regression, the role-index-drift fail-open regression, the last-colonel/unverify lockout paths, the rate-limiter bucket ordering, opaque-handle round-trips, and the network-posture matrix.

New server-side gates ship enabled with documented config knobs (`site.admin.*`) and fail safe for default self-hosted installs; `spec/config.test.yaml` disables the rate limits so unrelated tests are unaffected.

---
_Generated by [Claude Code](https://claude.ai/code/session_01M8yncPbzHMNh74gsMLnUXf)_

### DIFF
--- a/.env.reference
+++ b/.env.reference
@@ -1393,6 +1393,11 @@ ASSUME_HTTPS=false
 # and both surfaces 404. Filter mode with no explicit CIDRs trusts every
 # private-network peer as a proxy, which restores exactly the forwarded-host
 # spoofing the provenance rule exists to block.
+# THIS AND ADMIN_ALLOWED_CIDRS ARE ALSO WHAT PROMOTE the 15 destructive colonel
+# routes (purge, role change, revoke, delete, DLQ purge) from an ADVISORY
+# network requirement to an ENFORCED one (#4332). Set BOTH or neither: setting
+# only one gives you neither promotion and logs a boot WARN saying so. A
+# wildcard here does not count as set for this purpose.
 # Example: admin.example.com
 #ADMIN_ALLOWED_HOSTS=
 
@@ -1415,6 +1420,9 @@ ASSUME_HTTPS=false
 # matcher the privacy layer installs, so /32 and /128 entries work — see
 # docs/operations/admin-network-isolation.md, "CIDR precision and privacy
 # masking".
+# Paired with ADMIN_ALLOWED_HOSTS above, this is what promotes the 15
+# destructive colonel routes from an advisory network requirement to an
+# enforced one (#4332). Set BOTH or neither.
 # Example: 100.64.0.0/10,10.0.0.0/8
 #ADMIN_ALLOWED_CIDRS=
 
--- a/.env.reference
+++ b/.env.reference
@@ -1454,6 +1454,36 @@ ASSUME_HTTPS=false
 #COLONEL_ELEVATION_MAX_ATTEMPTS=5
 #COLONEL_ELEVATION_RATE_WINDOW=900
 #COLONEL_ELEVATION_LOCKOUT=900
+# Every MUTATING colonel verb (issue #4329), charged once per request from the
+# colonel logic base constructor. 120 per 5 minutes is far above any human
+# operator and above a bulk console session, so it only trips on scripted abuse.
+# A lockout here also blocks POST /ratelimit/reset (itself a mutation), so clear
+# this one with the commands `bin/ots ratelimit keys colonel_mutation <extid>`
+# prints.
+#COLONEL_MUTATION_RATE_LIMIT_ENABLED=true
+#COLONEL_MUTATION_MAX_ATTEMPTS=120
+#COLONEL_MUTATION_RATE_WINDOW=300
+#COLONEL_MUTATION_LOCKOUT=300
+# TIER 1 (destructive) verbs only: purge account, change role, revoke sessions,
+# delete org/domain/secret, DLQ purge. Charged LAST — after step-up,
+# confirmation and the per-verb interlocks all pass — so a rejected attempt
+# costs nothing and an attacker holding the cookie cannot burn the real
+# operator's budget with cheap 403s. 10 real actions per 5 minutes with a
+# 15-minute lockout: enough for an incident-response burst, and low enough that
+# a scripted purge stalls long before it can flush the count-capped audit trail.
+# Bulk work belongs on the CLI. Clear a lockout with POST /ratelimit/reset
+# (kind=colonel_destructive), which needs no elevation.
+#COLONEL_DESTRUCTIVE_RATE_LIMIT_ENABLED=true
+#COLONEL_DESTRUCTIVE_MAX_ATTEMPTS=10
+#COLONEL_DESTRUCTIVE_RATE_WINDOW=300
+#COLONEL_DESTRUCTIVE_LOCKOUT=900
+# The two session reads that resolve an opaque session handle (#4330) and may
+# fall back to a bounded 10 000-key scan plus as many HMACs. Every OTHER colonel
+# read is deliberately unlimited — the console fetches them on every screen.
+#COLONEL_HANDLE_RESOLVE_RATE_LIMIT_ENABLED=true
+#COLONEL_HANDLE_RESOLVE_MAX_ATTEMPTS=60
+#COLONEL_HANDLE_RESOLVE_RATE_WINDOW=300
+#COLONEL_HANDLE_RESOLVE_LOCKOUT=300
 
 
 # ═══════════════════════════════════════════════════════════
--- a/.env.reference
+++ b/.env.reference
@@ -1484,6 +1484,24 @@ ASSUME_HTTPS=false
 #COLONEL_HANDLE_RESOLVE_MAX_ATTEMPTS=60
 #COLONEL_HANDLE_RESOLVE_RATE_WINDOW=300
 #COLONEL_HANDLE_RESOLVE_LOCKOUT=300
+# Session bounds for the ADMIN API SURFACE (/api/colonel) only (issue #4331).
+# These do NOT shorten the onetime.session cookie and do NOT gate the /colonel
+# SPA shell: one session serves the admin console, the tenant app and the auth
+# app, so expiring the object would log a colonel out of the customer app too —
+# and on a self-hosted install the colonel is often the only customer. A stale
+# admin tab loads the shell, its first API call 401s, and the console renders an
+# expired banner with a sign-in link.
+# ADMIN_SESSION_IDLE_TIMEOUT reads the best-effort SessionMetadata sidecar and
+# is SKIPPED (never failed) when no record exists. It only bites because the
+# admin console makes no periodic requests, and the sidecar is a site-wide
+# activity clock, so tenant traffic keeps the admin window open too.
+# ADMIN_SESSION_ABSOLUTE_TIMEOUT reads session['authenticated_at'] — nothing a
+# stolen cookie can do advances it, so it is the bound that always binds.
+# Set either to 0 to disable that bound; ADMIN_SESSION_LIFETIME_ENABLED=false
+# disables both and restores the pre-#4331 24h rolling posture.
+#ADMIN_SESSION_LIFETIME_ENABLED=true
+#ADMIN_SESSION_IDLE_TIMEOUT=3600
+#ADMIN_SESSION_ABSOLUTE_TIMEOUT=43200
 
 
 # ═══════════════════════════════════════════════════════════
--- a/.env.reference
+++ b/.env.reference
@@ -1418,6 +1418,43 @@ ASSUME_HTTPS=false
 # Example: 100.64.0.0/10,10.0.0.0/8
 #ADMIN_ALLOWED_CIDRS=
 
+# Step-up (sudo) re-authentication for destructive colonel actions
+# (config site.admin.elevation, issue #4327). ON BY DEFAULT.
+# A colonel session alone is no longer sufficient for a tier-1 verb (purge
+# account, change role, revoke sessions, delete org/domain/secret, DLQ purge):
+# the operator must have re-proven a credential through
+# POST /api/colonel/elevation within the window. Confirmation (#4326) still
+# applies on top. Set to false to restore pre-#4327 behaviour.
+#COLONEL_ELEVATION_ENABLED=true
+# Window length in seconds (default 600 = 10 minutes).
+#COLONEL_ELEVATION_WINDOW=600
+# Seconds after sign-in during which a PASSWORD-LESS (SSO-only) colonel may
+# elevate with factor=recent_auth and no credential. DEFAULT 0 = OFF.
+# Available ONLY to accounts that cannot satisfy the password factor: giving it
+# to password holders would make step-up a no-op for the first N seconds after
+# every sign-in. In FULL auth mode every account counts as password-holding
+# (the Rodauth hash table is not reachable from the logic layer), so recent_auth
+# is unavailable there and an SSO-only full-mode fleet sets
+# COLONEL_ELEVATION_ENABLED=false instead. MFA as a step-up factor is not
+# implemented. AN SSO-ONLY FLEET MUST SET ONE OF THESE TWO OR ITS COLONELS
+# CANNOT PERFORM TIER-1 ACTIONS AT ALL.
+#COLONEL_ELEVATION_REAUTH_GRACE=0
+
+# Rate limits for the colonel API surface (config site.admin.rate_limit).
+# Keyed on the acting colonel's public external id (extid) — the same value the
+# audit trail records as `actor` — never on a session id. The parent flag
+# short-circuits every bucket.
+#COLONEL_RATE_LIMIT_ENABLED=true
+# Step-up attempts. The Rodauth password check behind POST /api/colonel/elevation
+# is an internal request and does NOT increment Rodauth's own lockout counter,
+# so this is the only backstop against guessing there. A locked-out operator
+# clears it with POST /api/colonel/ratelimit/reset (kind=colonel_elevation) or
+# the commands `bin/ots ratelimit keys colonel_elevation <extid>` prints.
+#COLONEL_ELEVATION_RATE_LIMIT_ENABLED=true
+#COLONEL_ELEVATION_MAX_ATTEMPTS=5
+#COLONEL_ELEVATION_RATE_WINDOW=900
+#COLONEL_ELEVATION_LOCKOUT=900
+
 
 # ═══════════════════════════════════════════════════════════
 #  MIDDLEWARE
--- a/.env.reference
+++ b/.env.reference
@@ -1444,8 +1444,10 @@ ASSUME_HTTPS=false
 # (the Rodauth hash table is not reachable from the logic layer), so recent_auth
 # is unavailable there and an SSO-only full-mode fleet sets
 # COLONEL_ELEVATION_ENABLED=false instead. MFA as a step-up factor is not
-# implemented. AN SSO-ONLY FLEET MUST SET ONE OF THESE TWO OR ITS COLONELS
-# CANNOT PERFORM TIER-1 ACTIONS AT ALL.
+# implemented. AN SSO-ONLY FLEET MUST CONFIGURE THIS OR COLONEL_ELEVATION_ENABLED
+# OR ITS COLONELS CANNOT PERFORM TIER-1 ACTIONS AT ALL — and which one depends on
+# the auth mode: in SIMPLE mode set this grace > 0 (or ENABLED=false); in FULL
+# mode recent_auth is never offered, so only ENABLED=false works.
 #COLONEL_ELEVATION_REAUTH_GRACE=0
 
 # Rate limits for the colonel API surface (config site.admin.rate_limit).
--- a/apps/api/colonel/auth_strategies.rb
+++ b/apps/api/colonel/auth_strategies.rb
@@ -3,7 +3,6 @@
 # frozen_string_literal: true
 
 require 'otto/utils'
-require 'uri'
 
 module ColonelAPI
   module AuthStrategies
@@ -52,8 +51,8 @@ module ColonelAPI
 
       # DoS ceiling on the ENCODED header, not a token-length limit: the real cap
       # is MAX_CONFIRM_LENGTH (characters, post-decode) in the logic layer. It is
-      # measured on the encoded bytes, which encodeURIComponent expands up to 12
-      # per 4-byte codepoint, so a max-length (255-char) token can approach 3 KB
+      # measured on the encoded bytes, which percent-encoding expands up to 12 per
+      # 4-byte codepoint, so a max-length (255-char) token can approach 3 KB
       # encoded. An over-long header is REJECTED, never truncated: slicing the
       # encoded bytes would sever a %XX escape or a multibyte character and turn a
       # valid token — e.g. a 100-char CJK org display_name, ~900 bytes encoded —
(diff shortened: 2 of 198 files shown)


delano:
@greptile-apps please go ahead despite the many files

delano:
Thanks for the review — all four defects are addressed and pushed. Each fix was reproduced first, then verified on Ruby 3.4.10 / Node (the CI toolchain).

| # | Finding | Resolution |
|---|---------|------------|
| 1 | Admin-console delete sent the **secret** shortid, not the receipt shortid → 403 for every delete | `39a20ede` — `deleteToken` now reads `details.metadata.shortid` (the receipt shortid the read-out already shows); stale comment corrected. |
| 2 | Expired-admin-session recovery link is a no-op in simple mode | `e0211875` — recovery now does a best-effort `GET /auth/logout` to clear the session, **then** navigates to sign-in, so re-auth re-stamps `authenticated_at`. Banner link → button; covered by `adminSessionExpiry`/`ElevationBanner` specs. |
| 3 | `admin_interface_spec.rb` not updated for the new gates → 4 red `ruby-integration-full` lanes | `82e179e8` (pushed before this review posted) — the 4 delete cases now send `X-OTS-Confirm` (the receipt shortid). Reproduced the 403s on `full-sqlite` with the header removed, 0 failures with it. |
| 4 | Unverify rollback re-verified with `verified_by: nil`, wiping provenance | `b4c92859` — capture the target's original `verified_by` **before** the write and restore it on rollback; guarded to the unverify-of-a-colonel path. Spec updated to assert the original tag, not `nil`. |

On your two non-blocking follow-ups — both confirmed, deferring as separate items rather than widening this PR:

- **`RoleSupport#sole_active_colonel?` returns `true` on an empty roster** (`[].none?`). Agreed it errs safe (refuses the demotion); only the refusal wording is misleading in the zero-effective-colonels state. Worth a message tweak, not a behavior change.
- **`supplied_confirmation` caps at 255 while `expected` is uncapped** — a real sharp edge only for an org `display_name` > 255 chars. Capping `expected` the same way slightly weakens uniqueness, and adding a name-length cap is a separate policy call, so I'd rather not decide it inline here.

Validation: `set_verification_spec` 13/0; admin vitest 17/0; `vue-tsc` (app + tests) and eslint clean; rubocop clean on `set_verification.rb`; `last_colonel_interlock` tryout 18/0 in simple mode (the lane CI runs it in). Happy to fold in the two follow-ups here if you'd prefer them in-PR.

---

delano:
Re-review findings fixed and pushed (`226b0f62`). You were right to elevate both from the earlier non-blocking list — both are genuinely reachable.

**Finding 1 — empty-roster refuses demoting an unverified colonel** (`b4c9285`… → `226b0f62`)
Fixed in **both** places, not just the pre-check: adding `verified?` only to `last_colonel?` would have moved the failure to the post-write `demotion_left_no_colonel?` (set_role.rb:154), which gates on the FROM-role but not `verified?` — so the write would happen and then roll back to `:last_colonel`. Both now require `customer.verified?`: an unverified colonel-role account isn't an active colonel, so demoting it can't empty the roster. A genuinely sole verified colonel stays protected. New spec: demoting an unverified colonel with an empty active roster → `:success`.

**Finding 2 — confirm-header cap on the encoded value** (`226b0f62`)
The cap was measured on the percent-encoded bytes and applied by truncation, severing a multibyte org name (≤100 chars ≈ 900 encoded bytes) into a permanent 403. Now: reject an over-long header **whole** rather than slicing it, raise `MAX_CONFIRM_BYTES` to 4096 (the real limit stays `MAX_CONFIRM_LENGTH`, post-decode), and decode with `URI.decode_uri_component` to match the console's `encodeURIComponent` — which also fixes the `+`→space mangling you flagged (plus-addressed emails sent raw by a non-browser client). Specs cover the 100-char CJK token, `+` preservation, and reject-not-truncate.

**Non-blocking follow-ups**
1. `set_verification.rb` rollback wiping `verified_by` — already fixed in `b4c92859` (this review ran on `82e179e8`, before that push): the rollback now captures the target's original `verified_by` before the write and restores it.
2. `resolve_handle_via_owner` resolving a stale index entry → `DeleteSession` returning 403 instead of 404 for an already-gone session — confirmed, leaving as a follow-up: it fails closed (refuses rather than mis-targets), and tightening it means confirming blob existence in the resolver. Happy to fold it in if you'd like it here.

Validation (CI toolchain): `set_role_spec` + `auth_strategies_spec` + `set_verification_spec` 41/0; rubocop clean on the three changed sources; `role_support` / `last_colonel_interlock` / `destructive_confirmation` tryouts 42/18/12 in simple mode (the lane CI runs them in).

---

delano:
Both findings fixed and pushed, plus the non-blocking follow-up. Each verified on Ruby 3.4.10.

**Finding 1 — `/ratelimit/reset` clears its own step-up lockout** (`b6b67231`)
`ResetRateLimit` now refuses resetting any `colonel_*` limiter for the **caller's own extid** over HTTP. Those buckets exist to bound the acting colonel — `colonel_elevation` is the sole step-up brute-force backstop (the internal-request password check never trips Rodauth's lockout), `colonel_destructive` holds the audit-cap line — so self-clearing them from a cookie made them no-ops. Self-recovery is now CLI-only; **peer-colonel resets and all non-colonel limiters are unaffected** (the operator-recovery path). Registry comment updated. Tests: self-reset of each `colonel_*` refused, peer + non-colonel allowed.

**Finding 2 — single-colonel self-email lockout** (`1b534360`)
`ChangeUserEmail`'s apply path now refuses when the change would strip the **last active colonel's** verification (which `has_system_role?` turns into a full `/colonel` lockout, since `ChangeEmail` clears verification with `enforce_interlocks:false`). `keep_verified=true` is the documented escape hatch — it preserves verification on the new address (the lesser evil you noted) for an operator who can't promote a second colonel first. Tests: refused for the last colonel, allowed with `keep_verified` and when a second verified colonel exists.

**Non-blocking follow-up — token computed before elevation** (`58b35b16`)
`guard_destructive_action!` now accepts a callable for `confirm_with` and invokes it **after** `require_elevation!`; `DeleteSecret` passes `-> { confirmation_token }`. A receiptless secret's fail-closed `GuardMisconfigured` (500) no longer pre-empts elevation, so an unelevated caller gets `403 ElevationRequired` first (no 500 oracle), and the fail-closed still applies once elevated. Tests cover both orderings.

Also, the earlier T3-Simple red was a test-encoding mismatch I fixed in `682354ac` (the org tryouts encoded the confirm header with `Rack::Utils.escape` — space→`+` — against the new `URI.decode_uri_component` decoder; now `URI.encode_uri_component`, matching the browser). Colonel suite: 552 unit + 125 tier-registry examples 0 failures; 28/28 colonel tryout files pass; rubocop clean.

---

delano:
Addressed both review rounds; pushed `9a8baaf2`. Verified on Ruby 3.4.10 + Node.

**Confirm-token decoding broke form-encoded (`+`) tokens** (`5acd7bbf`)
My earlier switch to `URI.decode_uri_component` (to match `encodeURIComponent`) kept `+` literal, so a form-encoded space (`Rack::Utils.escape`, `curl`, any non-browser client) stopped matching org `display_name` tokens — that's what turned the four `ruby-integration-*` lanes red. Reverted to `Rack::Utils.unescape` (form decoding), which accepts `%20`, a raw space, **and** `+`; kept the byte-cap fix (reject-oversized, the real multibyte-truncation bug). The one cost is a raw *unencoded* literal `+` decoding to a space — clients that mean `+` send `%2B` (encodeURIComponent and every form encoder do). The two org tryouts go back to `Rack::Utils.escape`.

**Full-mode expiry recovery was a no-op** (`90065981`)
`recoverAdminSession` used `GET /auth/logout`, but in full mode Rodauth owns `/auth` and only logs out on POST, so the session stayed authenticated and the router bounced `/signin` back to `/colonel` in a loop. Now uses `GET /logout` — the core app's logout (session.clear + sid renew), served in **both** modes and not shadowed by the `/auth` mount.

**`admin_timeout_setting` `to_i` footgun** (`31d9b20e`)
`"off"→0` silently disabled the bound; `"12h"→12`s. Now accepts only a clean non-negative integer (Integer or all-digits String); anything else (non-numeric string, YAML `off`→`false`) falls back to the default. Explicit `0` still disables. Mirrors `positive_colonel_setting`.

**`PurgeUser` last-colonel interlock** (`9a8baaf2`) — the follow-up you flagged. Refuses purging the last active colonel (pre-check only; purge is irreversible, so no post-write rollback — the residual concurrent double-purge is documented). Completes the #4328 coverage alongside SetRole/SetVerification/ChangeUserEmail.

On the other non-blocking notes: the `ResetRateLimit` self-reset of `colonel_destructive` you mentioned is already closed by `b6b67231` (self-reset of any `colonel_*` bucket is now refused over HTTP), so the config's audit-cap sizing claim holds again. The `delete_session` owner-token shape mismatch and the `resolve_handle` negative-path scan cost I'm leaving as genuine follow-ups (both fail closed / are bounded+authenticated) unless you'd like them in this PR.

Validation: colonel suite 704 examples + admin_session_lifetime + destructive_actions 0 failures · 28/28 colonel tryout files · admin vitest 17 · rubocop clean.

---

delano:
Pushed `4e94f95c` addressing the review and the red integration lane.

**Red CI — `colonel_rate_limit_spec.rb:257` (every T3 lane).** This test still asserted a colonel could clear their **own** `colonel_destructive` bucket over HTTP (200). The self-reset interlock now refuses that (422), so the test encoded the pre-interlock world. Replaced it with two that match the shipped contract: self-reset is refused (422) and the lockout holds, and a **peer** colonel's reset clears the bucket end to end — the operator-recovery path the refusal deliberately leaves open. Green locally in simple mode (10 examples, 0 failures).

**Finding 2 — recovery docs contradict the code.** Fixed. The module doc (`colonel_rate_limiter.rb`), `colonel-admin-guide.md` (both the elevation-throttle line and the "Clearing a lockout" section) and `pentest-scope.md` all told operators to self-reset via `POST /ratelimit/reset`. Aligned all three with `reset_rate_limit.rb`: your own lockout is CLI-only; a peer colonel (or any non-colonel limiter) is the HTTP path. `registry.rb`'s `colonel_destructive` row already stated this correctly.

**Finding 1 — non-numeric `ADMIN_SESSION_*_TIMEOUT`.** Already fixed on the reviewed head — `admin_timeout_setting` uses a numeric-string guard (`case raw when Integer … when String then raw.match?(/\A\d+\z/) …`), matching `positive_colonel_setting`, with spec coverage for `off`/`12h`/`false`/`"0"`. No further change.

**Non-blocking #1 — `ChangeUserEmail` confirm token.** Adopted the suggestion: `confirm_with: user.email` → `account_confirm_token(user)` (the helper `PurgeUser` already uses), so an email-less account falls back to its extid and gets a clean confirmation refusal instead of a `GuardMisconfigured` 500. Added a spec for that arm.

**Non-blocking #2 — raw vs normalized path in the isolation prefilter.** Leaving as-is: it fails **closed** (a percent-encoded `/api/%63olonel/…` 404s a route the router would serve), which is the safe direction. Happy to tighten the prefilter to `normalize_path` if you'd prefer parity — say the word and I'll do it in this PR.

**T2/T3 failures on `9a8baaf2`** were Docker-registry infra flakes during image pull (`502 Bad Gateway` on `valkey`/`rabbitmq`; `unauthorized: authentication required` on `postgres`) — both died before any test body ran. This push triggers a fresh run.

---

PR #4390, opened 2026-09-08
Wire the three Rodauth Admin integration seams into the colonel console

delano:
The standalone Rodauth Admin (onetimesecret/rodauth-admin, CHARTER §4) allows exactly three touchpoints with this repo, and this PR builds all of them. A new optional, credential-free `RODAUTH_ADMIN_URL` (`site.admin.rodauth_admin_url`) drives outbound links only: the customer detail page, the customers list drawer, and every sessions-console row and drawer deep-link to the matching Rodauth account by `external_id == extid`, targeting the admin's `/account?q=<extid>` lookup route. In `full` auth mode the sessions console and per-customer sessions panel now say they are not the session authority and link out, instead of presenting the Redis store as complete. The dev-only `/auth/admin/stats` stub is deleted and `docs/specs/rodauth-admin/` becomes a pointer to the standalone codebase.

The server decides whether a link exists (full mode and URL set), so the admin hostname never reaches a public bootstrap payload; unset renders plain text. No shared session, no cross-service call. Locale keys are added without `content_hash` on purpose; the rehash script owns those.


delano:
Addressed both review findings (pushed as `ee50e15` + `3f2c23c`):

**Session-authority notice contradicting Revoke-all (Item A)** — `SessionAuthorityNotice` is now surface-aware via a `context` prop. The per-customer panel passes `context="customer"` and renders a new `web.admin.sessions.authority.descriptionRevokeAll` string that distinguishes single-revoke (leaves the Rodauth row to self-expire) from **Revoke all** (which purges the account's `account_active_session_keys`). The global sessions console keeps the original copy, which is accurate there (single-revoke only).

**Unvalidated `RODAUTH_ADMIN_URL` (Item B, matches Greptile)** — `RodauthAdmin.base_url` now validates the value is an absolute http(s) URL (`URI::HTTP` with a host). A schemeless/foreign-scheme/unparseable value is treated as unset (callers render plain text instead of a dead link) and logged once. The warn memo is keyed on the raw value, so a runtime config reload to a *different* invalid value re-warns rather than staying silent behind the first.

Tests: Ruby unit spec 17/17 (accept https, reject schemeless/non-http/unparseable, warn-once, re-warn-on-change); admin vitest covers both console and customer notice surfaces; type-check clean.

Issue #4427, opened 2026-09-15
Enable tenant Connect: flip release gate and swap UI suppression to identity evidence


delano:
## Problem Statement

Tenant Connect is fully implemented server-side but remains fail-closed behind an interim release gate. Once #3849 acceptance coverage is complete and #4408 DoD is satisfied, three coupled changes must land together to make the feature live. This ticket exists so the flip is not lost between #3849's acceptance checklist and #4408's DoD, and so the coupled UI change is not forgotten.

## Early Diagnosis

The release gate is a single boolean and the UI suppression is a single call site. The risk is not complexity but coupling: enabling the callback pipeline without swapping the UI suppression rule would hide valid tenant connections whose provider route name matches an existing platform identity but whose `(provider, issuer, uid)` tuple is distinct. Both must ship in the same PR as the final acceptance suite.

## Affected Areas

- [ ] `apps/web/auth/config/hooks/omniauth_connect.rb:15` — flip `tenant_connect_enabled?` from `false` to configured/true; remove the release-gate comment
- [ ] `src/shared/utils/sso-link-evidence.ts:46-52` — replace route-name suppression in `isKnownLinked` with server-provided identity evidence (compare full `(provider, issuer, uid)` tuple, honor `_surface`)
- [ ] `apps/web/auth/spec/integration/full/omniauth_connect_link_spec.rb:948` — remove the `tenant_connect_enabled?` stub-open now that the gate is live; verify all tenant scenarios pass without the stub
- [ ] `docs/authentication/per-domain-sso.md` — remove interim-refusal language; document the enablement conditions actually met
- [ ] `docs/authentication/per-install-sso.md` — mirror any surface-facing wording changes

## Preconditions

Do not open a PR for this until:

- #3849 acceptance checklist is fully checked, including the full success / refusal / MFA / session-surface / membership-scope / identity-conflict / replay case matrix
- #4408 Definition of Done is satisfied
- The regression suite passes on tenant-surface scenarios with the stub removed

## Related

- Epic: #4408
- Pipeline: #3849
- Interim UI suppression: #4412 (closed; this ticket swaps it out)

Issue #4412, opened 2026-09-13
UI: keep tenant Connect available until callback identity is known


delano:
Parent epic: #4408

## Problem

`ConnectedIdentities.vue` suppresses Connect based on provider route name. For tenant providers, route-name or issuer equality does not prove that the callback identity's full `(provider, issuer, uid)` tuple is already linked. This can hide a valid Connect action before the application has enough information to compare identities.

## Scope

- Stop treating provider route-name equality as proof of an existing tenant link.
- Stop treating issuer equality alone as proof of an existing tenant link.
- Keep Connect available when identity equivalence is unknown before the callback.
- Defer the definitive comparison to the callback, where `(provider, issuer, uid)` is complete.
- Preserve safe suppression when the UI has authoritative evidence of the exact linked identity.

## Acceptance criteria

- [ ] Two identities sharing a route name do not cause Connect to be hidden.
- [ ] Two identities sharing an issuer but differing in `uid` do not cause Connect to be hidden.
- [ ] The callback remains authoritative for exact identity equality and conflict handling.
- [ ] UI tests cover unknown equivalence, same route/different identity, same issuer/different UID, and an exact known link.
- [ ] No frontend comparison can bypass the callback controls in #3849.

## Dependency

Coordinates with #3849; it does not weaken the existing fail-closed tenant callback refusal.

Issue #3849, opened 2026-07-22
SSO: authenticated tenant-surface identity linking


delano:
## Status

Authenticated identity linking is available on the platform surface. Tenant/custom-domain Connect remains deliberately refused until this issue is implemented completely.

The remaining scope is to support authenticated tenant-surface linking without allowing a tenant-controlled IdP to attach a login identity outside the tenant authorization boundary. Existing surface-isolation behavior and cross-tenant takeover regression coverage are already in place.

## Source of truth

The implementation specification is maintained in the authentication documentation:

- [Per-domain SSO: Requirements for authenticated tenant linking (#3849)](https://github.com/onetimesecret/onetimesecret/blob/main/docs/authentication/per-domain-sso.md#requirements-for-authenticated-tenant-linking-3849)
- [Per-install SSO: Connected Identities — Platform-only](https://github.com/onetimesecret/onetimesecret/blob/main/docs/authentication/per-install-sso.md#platform-only)

Those sections define the threat model, required controls, callback ordering, refusal behavior, identity-equivalence rules, membership requirements, re-authentication strength, MFA handling, and WebAuthn surface constraints. They are the source of truth if this tracker and the documentation diverge. Keep detailed design changes in the documentation rather than copying them into this issue.

## Remaining work

- Implement a tenant-surface-scoped authenticated session and enforce it against the resolved custom domain.
- Require recent full local re-authentication, including every MFA factor required by the account policy, before creating a tenant Connect intent.
- Require an active organization membership authorized for the exact custom domain before binding.
- Extend the existing fail-closed Connect callback path to tenant surfaces as specified in the per-domain documentation.
- Resolve identity ownership from the callback's full `(provider, issuer, uid)` tuple. Do not infer equivalence from provider or issuer alone.
- Update Connected Identities so Connect remains available when equivalence cannot be established before the callback.
- Apply the documented recent-re-authentication requirement to the existing platform Connect path.
- Add coverage for all specified success, refusal, MFA, session-surface, membership-scope, identity-conflict, and replay cases.

## Acceptance

- [ ] Tenant Connect conforms to the linked per-domain specification.
- [ ] A tenant identity is bound only to the authenticated session account, after exact-domain membership authorization and recent full local re-authentication on that tenant surface.
- [ ] Email claims, email authentication, password interstitials, mailbox-proof linking, and automatic membership creation cannot satisfy or bypass the tenant Connect gates.
[...]

PR #4428, opened 2026-09-15
[#3849] Implement tenant Connect callback pipeline behind release gate

delano:
Implements the shared Connect callback pipeline (`hooks/omniauth_connect.rb`) that consumes intent before every gate, then runs session-account, Customer status, surface binding, exact-domain membership, and full-tuple ownership checks. `tenant_connect_enabled?` stays hard-coded `false` so tenant Connect remains refused in production until the acceptance matrix is complete; the pipeline is exercised via the new `omniauth_connect_link_spec.rb` full-lane spec.

Frontend: Connected Identities panel keeps the interim tenant route-name suppression (#4412) while the release gate is closed, since a Connect action the callback refuses unconditionally would only burn the user's re-auth proof. The swap to full-tuple identity evidence (`isKnownLinked` suppressing only on the platform surface) ships with the gate flip (#4427).

Docs: `per-domain-sso.md` records the pipeline order (consume intent before tenant validation), the current implementation status, and the acceptance gate — the concrete regression matrix and focused spec files required before flipping the release gate.

Related to #3849


delano:
Tracked the unmerged review follow-ups separately:

- #4433 — platform-fallback Connect is advertised on custom hosts but refused at callback
- #4431 — broad Connect rescue can misclassify an ordinary sign-in as an identity conflict
- #4432 — hook ownership guard does not cover prepended callback wrappers

PR #4449, opened 2026-09-17
Document bootstrap snapshot ordering contract

delano:
## Summary
- Add proposed ADR-046 defining server-authoritative ordering for HTML hydration and `GET /bootstrap/me` snapshots.
- Scope ordering to a SID-derived opaque epoch with an atomic SessionSidecar `INCR`, avoiding exposure of installation-wide snapshot volume.
- Keep `snapshot_generated_at` diagnostic-only and specify atomic client acceptance, request-generation invalidation, and authenticated epoch transitions.

## Review guide
- Start with the server-allocation section and verify the HMAC epoch, sidecar lifetime, allocation-failure posture, and SID-renewal boundary.
- Review the client-acceptance rules for atomic rejection, downgrade handling, and authorization of epoch transitions.
- The implementation notes record rejected datetime, Redis statistics, global counter, and obfuscation approaches.

## Validation
- Commit hooks passed, including ADR reference lint.
- Pre-push checks passed; runtime tests were not run for this documentation-only change.


delano:
Proposed ADR edit, for consistency with the workplan (PR #4471, step 9) and #4460. Not applied here because the ADR worktree has uncommitted edits in progress.

**Where:** "Server allocation", allocation-failure item, `GET /bootstrap/me` bullet. It currently ends: "the failure counts toward the existing consecutive-failure limit." The existing limit (`AUTH_CHECK_CONFIG.MAX_FAILURES` in `authStore.checkWindowStatus`) calls `authStore.logout()`, so as written the ADR inherits logout-on-outage.

**Suggested replacement for that sentence:**

> The client treats it as a failed refresh: it retains its last accepted state, and the coordinator retries with backoff and jitter, honoring `Retry-After`. A failed refresh is not an authentication verdict. Reaching the consecutive-failure limit moves the client to the verification-unavailable state, which blocks protected content and keeps retrying; it does not log the user out and does not touch the server session. That state is specified in #4460, not here.

**Suggested entry under "alternatives considered and rejected":**

> - **Logging out at the consecutive-failure limit:** Rejected. A transport failure or 5xx carries no information about the credential (RFC 9110 §15.5.2 and §15.6.4; RFC 6749 separates `invalid_grant` from `temporarily_unavailable`; MSAL.js requires re-authentication only for `InteractionRequiredAuthError`). `authStore.logout()` clears client state only, so the page would show a signed-out user while the cookie still authenticates, against the OWASP Session Management Cheat Sheet's requirement that logout invalidate the session server-side. Fail-safe defaults call for withholding access when authority is unknown, which the verification-unavailable state does without asserting an identity change. Mass logout during a transient outage also sends every open tab to the login endpoint while the backend is degraded.

Two smaller consistency points:
- The rejected alternative "Rejecting an unexpected epoch on an ordinary refresh" says "the consecutive-failure limit would never fire". Still true, but after this change the limit no longer means logout, so the sentence reads better as "the failure handling would never engage".
- Sub-issues #4457, #4464, #4465 now defer to this ADR instead of restating it, so revisions here do not need matching issue edits.

Issue #4482, opened 2026-09-18
Design SAML Single Logout (SLO) against active-session semantics


delano:
Follow-up to #4450.

## Problem

SAML Single Logout (SLO) coordinates logout with the identity provider (IdP) and other applications in the same IdP session.

It has two directions:

1. **SP-initiated SLO:** the user logs out of Onetime Secret; we notify the IdP, which can end the user’s sessions in other applications.
2. **IdP-initiated SLO:** the IdP notifies us that the user logged out elsewhere; we end the corresponding Onetime Secret session.

`omniauth-saml` exposes these endpoints as `/slo` and `/spslo` below the strategy path, and enables them by default.

## Why it is intentionally disabled

The gem’s default handler for an IdP-initiated logout calls `session.clear` on the Rack session. That removes the cookie blob but does not touch the application’s active-session rows.

Our session vocabulary and source-of-truth model are deliberate:

- **revoke** = active-session row;
- **destroy** = cookie/session blob;
- **refuse** = authentication gate.

Enabling the gem defaults would let an IdP observe an apparently successful logout while the corresponding active-session row remains alive. This violates the application’s session lifecycle semantics.

Therefore, `slo_enabled: false` is required today. Both SLO endpoints return `501` until the protocol is designed against this vocabulary. Ordinary local logout remains supported; it simply does not propagate to the IdP.

## Scope for a future design

Design SLO as an application-level session-lifecycle feature, not a default OmniAuth handler. The implementation must specify both SP- and IdP-initiated behavior.

The design should cover:

- mapping a SAML logout request or response to the exact active-session row(s) to revoke;
- correct ordering and failure behavior for row revocation and cookie/blob destruction;
- signed request/response validation, issuer/audience checks, replay protection, and correlation of SLO request IDs;
- IdP-initiated logout behavior when no local session cookie is present, or when multiple local sessions could match;
- SP-initiated propagation, callback completion, partial IdP failure, timeouts, and user-visible local-logout semantics;
- tenant/platform configuration, endpoint routing, IdP metadata, and cross-site POST/origin handling;
- audit logging that excludes assertions and other sensitive SAML payloads;
- integration tests proving that every accepted IdP-initiated logout revokes the relevant active-session row(s), not merely the Rack cookie.

## Acceptance criteria

- SLO endpoint handling uses application session revocation semantics rather than `session.clear` alone.
- Both directions of SLO have explicit correlation, validation, replay, and failure policies.
[...]

Issue #4481, opened 2026-09-18
Design support for IdP-initiated SAML SSO


delano:
Follow-up to #4450.

## Problem

We currently support only **SP-initiated SAML SSO**: the user starts on our sign-in page, selects “Sign in with SAML”, and we redirect to the identity provider (IdP) with an AuthnRequest. The IdP response includes `InResponseTo`, which must match the AuthnRequest ID saved in the browser session.

**IdP-initiated SSO** starts elsewhere. A user opens their company app portal (such as Okta or Microsoft Entra), selects the Onetime Secret tile, and the IdP POSTs a signed SAML response directly to our callback URL. We never sent an AuthnRequest.

The portal-tile experience is intentionally unsupported today: users must begin from our sign-in page.

## Why it is intentionally disabled

An IdP-initiated response has no `InResponseTo`, so it cannot be bound to a pending request in the user’s browser session. Accepting it would recreate the login-CSRF condition described in precondition 5 of #4450: a cross-site POST could log a browser into whichever account the signed assertion names.

The SAML strategy subclass therefore refuses any callback without a pending request ID, returning `:saml_no_pending_request`. By construction, that rejects IdP-initiated flows.

## Scope for a future design

Before enabling IdP-initiated SSO, define and implement a separate secure browser/session-binding mechanism. A RelayState-based design is one possible direction, but it must be specified and reviewed rather than treated as an option toggle.

The design should cover:

- the required browser binding and how it resists login CSRF;
- validation and integrity requirements for RelayState or an alternative binding;
- tenant and platform routing/issuer resolution for unsolicited callbacks;
- callback error behavior and operator-facing configuration;
- assertion replay handling alongside the one-shot assertion-ID cache;
- dedicated tests for cross-site unsolicited POSTs and account-binding failures.

## Acceptance criteria

- IdP-initiated callbacks cannot log a browser into an assertion-selected account without a deliberate, validated browser binding.
- The design preserves the request/response and replay protections introduced for #4450.
- Portal-initiated SAML works only after the security model, configuration contract, and tests are implemented.
- Until then, `:saml_no_pending_request` remains the expected refusal for callbacks lacking a pending AuthnRequest.

## Related

- #4450 — Add SAML SSO strategy: ADR-044 proposal and preconditions

PR #4484, opened 2026-09-18
Add SAML SSO strategy: request-bound omniauth-saml subclass, platform and tenant surfaces

### LINKED ISSUE #4450: Add SAML SSO strategy: ADR-044 proposal and preconditions
## Problem Statement

We have no SAML sign-in path. Enterprise buyers ask for it by name, and some regulated or older organisations run IdPs with no OIDC option, so generic OIDC cannot reach them. `omniauth-saml` is current and widely used, but as shipped it resolves to the issuerless `''` sentinel (refused on the tenant surface), accepts unsolicited responses, and pulls in `ruby-saml`, which has a history of signature-verification CVEs. This issue is the ADR-044 admission proposal and records the preconditions that must be met before the gem is added.

## Early Diagnosis

Checked against omniauth-saml 2.2.5 and ruby-saml 1.18.1 sources and `main` at 23df2ffbfa. `file:line` references to `settings.rb`, `response.rb` and `saml.rb` are in those gems.

### Admission

ADR-044 criterion 1: an IdP that speaks only SAML has no OIDC login flow, so generic OIDC is not an alternative. No market exception claimed.

**Validated issuer:** not as shipped. The strategy resolves to the `''` sentinel today and would be refused on the tenant surface. It becomes issuer-capable only after preconditions 1 and 2. This proposal is for an issuer-capable, tenant-eligible strategy; an issuerless platform-only SAML is not proposed.

### Dependency

- omniauth-saml 2.2.5 (omniauth ~> 2.1, ruby-saml ~> 1.18)
- ruby-saml 1.18.1. Its last two feature releases were critical-CVE fixes (1.17.0 signature-verification bypass, 1.18.0 parser differentials).
- Pin ruby-saml exactly in the Gemfile and add it to advisory tracking. Re-verify the preconditions below on every bump, same as the rodauth-omniauth ordering invariant in `features/omniauth.rb`.

#### 1. `issuer:` must never reach the SAML strategy options

ruby-saml's `issuer` setting is a deprecated alias for the SP EntityID (`settings.rb:121-122`, `@sp_entity_id || @issuer`). `resolve_issuer` reads `strategy_options[:issuer]` at precedence 1 (`features/omniauth.rb:220`). A definition that declares `issuer:` the way `sso_provider/apple.rb:113` does would key every tenant's identities on our own EntityID, one constant, collapsing the `(provider, issuer, uid)` partition.

- Use `sp_entity_id:` only. Registry spec asserts the SAML definition's strategy options carry no `:issuer` key.
- New resolve branch: read the IdP EntityID from `extra.response_object.issuers` (`response.rb:305`; Response + signed-Assertion Issuer, uniq'd). Require exactly one value, else sentinel (tenant refusal follows). `extra.raw_info` is the attribute set and has no `iss`, so the existing token-issuer branch cannot serve.

#### 2. `idp_entity_id` required

`validate_issuer` returns true when `settings.idp_entity_id` is nil (`response.rb:746`). Without it the value read in #1 is unvalidated and attacker-chosen. `SsoConfig#validation_errors` must require it for `saml`, and the platform env config must fail boot without it. The resolve branch should also compare the response issuer to the configured one and fail closed on mismatch rather than trusting the gem alone.

#### 3. `SsoConfig` shape

The model is OAuth-shaped: universal `client_id` requirement (`sso_config.rb:339`), `issuer`/`tenant_id` discriminators, `client_id`/`client_secret` the only `encrypted_field`s. SAML needs `idp_sso_service_url`, `idp_entity_id`, `idp_cert` and has no client credential.

- `client_id` requirement becomes per-provider.
- `idp_cert` is not secret but is the trust anchor. `encrypted_field ... aad_fields: [:domain_id]` is the only domain binding in the model; store the cert (and `idp_entity_id`, `idp_sso_service_url`) that way so a swapped value fails authentication. Do not accept `idp_cert_fingerprint` (ruby-saml defaults the fingerprint algorithm to SHA1, `settings.rb:280`).
- `PROVIDER_TYPES`, `PROVIDER_METADATA`, `PROVIDER_ROUTE_MAP`, schema `models/domain-sso-config`, `BackfillTenantIssuer#resolve_issuer` (must byte-match).

#### 4. Stable uid

Default uid is the NameID (`saml.rb:111-121`). Transient NameIDs rotate per session: one identity row per login. Request `name_identifier_format` persistent, and refuse the callback when `response_object.name_id_format` is transient unless `uid_attribute` is configured. Missing `uid_attribute` already raises in the gem.

#### 5. Unsolicited responses and replay

omniauth-saml never records the AuthnRequest ID and never passes `matches_request_id`, so `validate_in_response_to` is a no-op (`response.rb:626-628`). SAML has no `state` parameter. Result: IdP-initiated responses are accepted by default, which is login CSRF on a cross-site POST, and nothing stops an assertion being replayed inside its validity window.

- Store the request `uuid` in the session at request phase, pass it as `matches_request_id` at callback. Needs a strategy subclass (request_phase keeps the Authrequest local).
- One-shot assertion ID cache in Valkey, TTL = NotOnOrAfter.
- Tenant Connect intent (#3849) is consumed before gates; an unsolicited response must not find one.

#### 6. Turn off what we do not use

- `slo_enabled` defaults true and exposes `/slo`, `/spslo` under the request path; the IdP-logout default is `session.clear` on the Rack session, which bypasses active-session rows. Set `slo_enabled: false` until SLO is designed against revoke/destroy vocabulary.
- ruby-saml security defaults: `want_assertions_signed: false`, SHA1 digest/signature, `check_idp_cert_expiration: false`, `strict_audience_validation: false`. Set all four explicitly.
- `extra.response_object` is the full Response including the raw assertion. Confirm it is never logged, persisted to the session, or sent to diagnostics.

### What already fits

- Cross-site POST callback: `HttpOriginOptions` exists for Apple's form_post. Tenant SAML needs `tenant_idp_origin` to derive from the tenant's `idp_sso_service_url`: add `saml` to `ISSUER_DERIVED_PROVIDER_TYPES` dispatch in `tenant_origin_source` (`auth_config.rb:29`), reading the SSO URL rather than `issuer`. Session cookie SameSite behavior on the POST callback is whatever Apple already settled.
- `email_verification_hold`: absence of `email_verified` is not a hold, so SAML JIT behaves like Entra. No change.

### Size

Comparable to the tenant OIDC work, not an Apple-sized drop-in. Suggested order: strategy subclass + #1/#2/#4/#5/#6 on the platform surface first (env-configured, single IdP), then #3 and tenant eligibility.

## Affected Areas

- [ ] `Gemfile` — omniauth-saml, exact ruby-saml pin, advisory tracking
- [ ] `lib/onetime/sso_provider/registry.rb` — SAML definition; no `:issuer` strategy option
- [ ] `apps/web/auth/config/features/omniauth.rb` — `resolve_issuer` branch for the IdP EntityID; strategy registration
- [ ] `apps/web/auth/config/hooks/omniauth.rb` — transient-NameID refusal; keep `response_object` out of logs
- [ ] `apps/web/auth/config/hooks/omniauth_tenant.rb` — tenant credential injection for `saml`
- [ ] `lib/onetime/models/custom_domain/sso_config.rb` — provider type, per-provider required fields, AAD-bound IdP fields
- [ ] `apps/web/auth/operations/backfill_tenant_issuer.rb` — issuer resolution must byte-match
- [ ] `lib/onetime/auth_config.rb` — `tenant_idp_origin` / `tenant_origin_source` from the tenant SSO URL
- [ ] `lib/onetime/middleware/http_origin_options.rb` — cross-site POST callback allowance
- [ ] `docs/authentication/adding-sso-providers.md` — checklist entry and operator docs
- [ ] `docs/adr/adr-044-omniauth-adapter-admission-criteria.md` — reference case for criterion 1

delano:
Closes #4450.

## What

Adds SAML 2.0 as an SSO provider on both surfaces: the platform (env-configured, single IdP) and custom-domain tenants (`SsoConfig` provider type `saml`). The gem is admitted under ADR-044 criterion 1 (an IdP that speaks only SAML has no OIDC flow), and the ADR records it as the reference case.

`omniauth-saml ~> 2.2` with `ruby-saml` pinned exactly at 1.18.1; the Gemfile comment carries the CVE history and the list of gem internals to re-verify on every bump. A `gem-audit` job (bundler-audit) now runs on every PR. sqlite3 is bumped 2.9.5 → 2.9.6 in the lockfile so that job is green (GHSA-mwm8-39rw-8826, unrelated to SAML).

## How the preconditions from #4450 landed

All SAML-specific gates live in one strategy subclass, `OmniAuth::Strategies::RequestBoundSAML` (`lib/onetime/sso_provider/request_bound_saml.rb`), so nothing depends on hook ordering and the rest of the pipeline never sees the raw response.

1. **No `issuer:` option.** The definition uses `sp_entity_id` only (registry spec asserts no `:issuer` key). Identities are keyed on the IdP EntityID read from the validated response, which must equal the configured `idp_entity_id` byte-for-byte (stricter than ruby-saml's `uri_match?`). Blank or mismatched → refused on **both** surfaces, never the `''` sentinel.
2. **`idp_entity_id` required.** Platform: `SAML_IDP_ENTITY_ID`, `SAML_IDP_SSO_SERVICE_URL`, `SAML_IDP_CERT` are `required_vars` with a `vars_valid` check (cert parses, URL https). Tenant: required in the model, PUT/PATCH and the form.
3. **`SsoConfig` shape.** `client_id`/`client_secret` are now per-provider. The SAML trio is stored as `encrypted_field ... aad_fields: [:domain_id]`, so a swapped trust anchor fails to decrypt; the API reports such fields in `unreadable_fields` and the form shows an error state instead of a blank. No fingerprint parameters are accepted.
4. **Stable uid.** Persistent `name_identifier_format` requested; a transient NameID is refused unless `uid_attribute` is configured.
5. **Unsolicited responses and replay.** The AuthnRequest id is stored one-shot in the session and passed as `matches_request_id`; a callback with no pending request is refused. Assertion ids are claimed atomically in Valkey (`Onetime::Security::SamlAssertionReplayGuard`, TTL = NotOnOrAfter + clock drift); replay or datastore error fails closed.
6. **Off / hardened.** `slo_enabled: false`; the full ruby-saml `security` hash is supplied from one shared builder (a partial hash *replaces* the gem defaults, it does not merge); `want_assertions_signed`, cert-expiry check, strict audience, SHA-256 fingerprint algorithm. `extra` is rewritten to scalars — `response_object` never reaches the auth hash, logs, session or diagnostics. ruby-saml's logger is redirected so AuthnRequest XML never hits STDOUT.

## Decisions that differ from the issue text

- **Missing platform config skips the provider; it does not fail boot.** `configure_provider` is built never to kill boot (ADR-033 compromise). A missing or invalid `SAML_*` var logs loudly and leaves the sign-in button unadvertised. Documented in `adding-sso-providers.md`.
- **Tenant identities are scoped to the domain.** Review found that a tenant admin could configure another IdP's EntityID with their *own* cert and mint assertions for any NameID at that IdP, matching the victim's identity row (reproduced end to end before the fix). Tenant SAML identities are now keyed on `"<domain_id>|<EntityID>"`; the platform keeps the bare EntityID. `BackfillTenantIssuer` writes the same form and refuses an unscoped `--issuer` override for `saml`.
- **SHA-1 signatures are refused.** ruby-saml's `digest_method`/`signature_method` only govern what *we* sign and the verifier accepts whatever the response declares. The subclass reads every `ds:Signature` from the document the gem validated and refuses anything outside an RSA/ECDSA SHA-256/384/512 allowlist.
- **No SSRF host check on the tenant SSO URL.** The server never fetches it (browser navigation target only), so the OIDC-issuer check was refusing legitimate private-network IdPs while protecting nothing. It must still be https, without userinfo, with a plain hostname that yields a CSP/HttpOrigin-safe origin (no trailing dot). The OIDC issuer keeps the SSRF check because that one is fetched.
- **Cross-site POST callback.** `HttpOriginOptions` only knew platform IdP origins; tenant SAML callbacks would have been refused with 403. It now admits the resolved tenant IdP origin (callback path, POST only, fail closed), and `tenant_origin_source` became a per-type field map so CSP form-action and HttpOrigin agree.
- **OmniAuth failure logging is bounded** for every provider (500 chars, one line) — a pre-Success-status response can carry an unsigned, attacker-chosen `StatusMessage` into the gem's exception text.

## Deliberately not in this PR

- **IdP-initiated SSO is unsupported** by design (the request-bound callback refuses responses without a pending AuthnRequest). Follow-up: #4481 — Design support for IdP-initiated SAML SSO.
- **Single Logout is off.** omniauth-saml's IdP-logout default is `session.clear`, which bypasses active-session rows. Follow-up: #4482 — Design SAML Single Logout against active-session semantics.
- Non-`en` locale strings for the 20 new form keys (normal translation workflow).

## Open design question (not blocking)

`SubjectConfirmationData/@InResponseTo` is checked only when present (ruby-saml behaviour). Requiring it would bind the AuthnRequest id to the *signed* assertion rather than the unsigned Response attribute plus our one-shot session key, at the cost of refusing IdPs that omit it. Not implemented; wants a product decision.

## Testing

Real signed responses from a runtime-minted test IdP (`spec/support/saml/test_idp.rb`, no key material committed). New lane `full-saml-platform` (wired into CI) drives platform sign-in end to end through real Rodauth; `tenant_saml_sso_spec.rb` does the same for tenants, including the cross-tenant attack, replay, unreadable-record and SHA-1 refusals. Tryouts hit the real datastore for the replay guard and the model.

All lanes green after the last commit: unit (8371 tryouts; 6269/6096/56 specs), full-pg 82, full-sqlite 2060, full-pg-agnostic 2025, full-mfa 10, full-saml-platform 13, simple 1403 (+1292 tryouts), api 268, disabled 1007, migrations-sqlite 18; bundler-audit clean; rubocop clean; vitest 4825; type-check clean; locale hashes 0 drift. `origin/main` merged in (no conflicts) after that run; the merge brought only docs and package bumps.

Three independent reviews (security, correctness, coverage) produced 12 findings, all confirmed and fixed with tests; a delta review of the fixes produced 3 low findings, also fixed. Visual-baseline CI is red on every PR and can be ignored.


delano:
We made a lot of progress on the implementation, including addressing a lot of the feedback. Closing this to open a fresh PR. We'll carryover any remaining feedback items.

